PR #6, opened 2017-10-17
add EntryPoint Attribute

lightszero:


Neo.SmartContract.Framework/EntryPointAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Neo.SmartContract.Framework.Services.Neo;
+using System;
+
+namespace Neo.SmartContract.Framework
+{
+    [AttributeUsage(AttributeTargets.Method )]
+    public class EntryPointAttribute : Attribute
+    {
+        public TriggerType type { get; }
```
erikzhang: The first letter of public fields should be capitalized.

PR #9, opened 2017-12-02
[NEP4] Adding support for Dynamic Contract Invokes

localhuman:
In regards to this NEP https://github.com/neo-project/proposals/pull/19

Adding method `DynamicAppCall` to Neo.SmartContract.Framework.SmartContract for dynamic app call invokes.

### DIFF
--- a/Neo.SmartContract.Framework/DynamicCallAttribute.cs
+++ b/Neo.SmartContract.Framework/DynamicCallAttribute.cs
@@ -2,12 +2,9 @@
 
 namespace Neo.SmartContract.Framework
 {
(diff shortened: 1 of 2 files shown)


localhuman:
Recent change set uses `System.Func` instead.  A proposed usage of this approach is here:

https://github.com/localhuman/examples-csharp/blob/feature-dynamic-invoke/RuntimeInvoke/RuntimeInvoke.cs

erikzhang:
We can make it easier:
```c#
public static void Main(Func<int, int, int> add)
{
    add(1, 2);
}
```

erikzhang:
@lightszero Can we change the compiler to support this?

lightszero:
yes,working on it.

PR #21, opened 2018-03-26
Update Map.cs

erikzhang:


saniales:
Am i able to serialize it into the Storage? I used map.Serialize() but I cannot verify if it's working...

PR #29, opened 2018-10-14
Explicit byte conversion

igormcoelho:
This helps partially resolving issue https://github.com/neo-project/neo-compiler/issues/134

### DIFF
--- a/Neo.SmartContract.Framework/Helper.cs
+++ b/Neo.SmartContract.Framework/Helper.cs
@@ -6,51 +6,92 @@ namespace Neo.SmartContract.Framework
 {
     public static class Helper
     {
+        /// <summary>
+        /// Converts byte[] to BigInteger. Examples: [0x0a] -> 10; [0x80] -> -128; [] -> 0; [0xff00] -> 255
+        /// </summary>
         [Nonemit]
         public extern static BigInteger AsBigInteger(this byte[] source);
 
+        /// <summary>
+        /// Converts BigInteger to byte[]. Examples: 10 -> [0x0a]; -128 -> [0x80]; 0 -> []; 255 -> [0xff00]
+        /// </summary>
         [Nonemit]
         public extern static byte[] AsByteArray(this BigInteger source);
 
+        /// <summary>
+        /// Converts string to byte[]. Examples: "hello" -> [0x68656c6c6f]; "" -> []; "Neo" -> [0x4e656f]
+        /// </summary>
         [Nonemit]
         public extern static byte[] AsByteArray(this string source);
         
-        // a <= x && x < b
+        /// <summary>
+        /// Converts byte[] to string. Examples: [0x68656c6c6f] -> "hello"; [] -> ""; [0x4e656f] -> "Neo"
+        /// </summary>
+        [Nonemit]
+        public extern static string AsString(this byte[] source);
+        
+        /// <summary>
+        /// Returns true iff a <= x && x < b. Examples: x=5 a=5 b=15 is true; x=15 a=5 b=15 is false
+        /// </summary>
         [OpCode(OpCode.WITHIN)]
         public extern static bool Within(this BigInteger x, int a, int b);
         
-        // a <= x && x < b
+        /// <summary>
+        /// Returns true iff a <= x && x < b. Examples: x=5 a=5 b=15 is true; x=15 a=5 b=15 is false
+        /// </summary>
         [OpCode(OpCode.WITHIN)]
         public extern static bool Within(this int x, int a, int b);
         
-        // faults if b is false
+        /// <summary>
+        /// Faults iff b is false
+        /// </summary>
         [OpCode(OpCode.THROWIFNOT)]
         public extern static void Assert(this bool b);
         
+        /// <summary>
+        /// Converts and ensures parameter source is sbyte (range 0x00 to 0xff); faults otherwise.
+        /// Examples: 255 -> fault; -128 -> [0x80]; 0 -> []; 10 -> [0x0a]; 127 -> [0x7f]; 128 -> fault
+        /// </summary>
         public static sbyte AsSbyte(this BigInteger source)
         {
             Assert(source.Within(-128, 128));
             return (sbyte) source;
         }
         
+        /// <summary>
+        /// Converts and ensures parameter source is sbyte (range 0x00 to 0xff); faults otherwise.
+        /// Examples: 255 -> fault; -128 -> [0x80]; 0 -> []; 10 -> [0x0a]; 127 -> [0x7f]; 128 -> fault
+        /// </summary>
         public static sbyte AsSbyte(this int source)
         {
             Assert(source.Within(-128, 128));
             return (sbyte) source;
         }
         
+        /// <summary>
+        /// Converts and ensures parameter source is byte (range 0x00 to 0xff); faults otherwise.
+        /// Examples: 255 -> fault; -128 -> [0x80]; 0 -> []; 10 -> [0x0a]; 127 -> [0x7f]; 128 -> fault
+        /// </summary>
         public static byte AsByte(this BigInteger source)
         {
             Assert(source.Within(-128, 128));
             return (byte) source;
         }
-        
+
+        /// <summary>
+        /// Converts and ensures parameter source is byte (range 0x00 to 0xff); faults otherwise. 
+        /// Examples: 255 -> fault; -128 -> [0x80]; 0 -> []; 10 -> [0x0a]; 127 -> [0x7f]; 128 -> fault
+        /// </summary>
         public static byte AsByte(this int source)
         {
             Assert(source.Within(-128, 128));
             return (byte) source;
         }
         
+        /// <summary>
+        /// Converts parameter to sbyte from (big)integer range 0-255; faults if out-of-range. 
+        /// Examples: 256 -> fault; -1 -> fault; 255 -> -1 [0xff]; 0 -> 0 []; 10 -> 10 [0x0a]; 127 -> 127 [0x7f]; 128 -> -128 [0x80]
+        /// </summary>
         public static sbyte ToSbyte(this BigInteger source)
         {
             if(source > 127)
@@ -59,6 +100,10 @@ namespace Neo.SmartContract.Framework
             return (sbyte) source;
         }
         
+        /// <summary>
+        /// Converts parameter to sbyte from (big)integer range 0-255; faults if out-of-range. 
+        /// Examples: 256 -> fault; -1 -> fault; 255 -> -1 [0xff]; 0 -> 0 []; 10 -> 10 [0x0a]; 127 -> 127 [0x7f]; 128 -> -128 [0x80]
+        /// </summary>
         public static sbyte ToSbyte(this int source)
         {
             if(source > 127)
@@ -67,6 +112,10 @@ namespace Neo.SmartContract.Framework
             return (sbyte) source;
         }
         
+        /// <summary>
+        /// Converts parameter to byte from (big)integer range 0-255; faults if out-of-range. 
+        /// Examples: 256 -> fault; -1 -> fault; 255 -> -1 [0xff]; 0 -> 0 []; 10 -> 10 [0x0a]; 127 -> 127 [0x7f]; 128 -> -128 [0x80]
+        /// </summary>
         public static byte ToByte(this BigInteger source)
         {
              if(source > 127)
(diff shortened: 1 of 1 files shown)


Neo.SmartContract.Framework/Helper.cs
```diff
@@ -14,6 +14,19 @@ public static class Helper
 
         [Nonemit]
         public extern static byte[] AsByteArray(this string source);
+        
+        public static byte AsByte(this BigInteger source)
+        {
+            byte[] b = source.AsByteArray();
+            return b[0];
```
shargon: When a big number is converted (like 2000), this method should throw and exception.

Neo.SmartContract.Framework/Helper.cs
```diff
@@ -14,6 +14,19 @@ public static class Helper
...
+        }
+        
+        public static byte AsByte(this int source)
+        {
+            BigInteger bigSource = source;
+            byte[] b = bigSource.AsByteArray();
+            return b[0];
+        }
```
shargon: same here

Neo.SmartContract.Framework/Helper.cs
igormcoelho: I was avoiding to generate jumps and more complicated stuff in such a simple method... this way, a byte is always generated, independent of the number. We can rename the method too, like "FirstByte", which works the same way for small numbers.

igormcoelho: In fact, my fear is that an empty bytearray is passed (usually interpreted as zero I guess), and it will break, instead of returning zero. Do you think we need to check the whole thing @shargon ? Safety comes first I guess :)

Neo.SmartContract.Framework/Helper.cs
igormcoelho: Redirected this method to the other, to avoid duplication.

igormcoelho:
Feeling safe now @shargon ? :)

shargon:
Now is better :P

igormcoelho:
No Shargon... two extra bytes on short circuit OR xD but less computing... ;)

igormcoelho:
@erikzhang this one works fine even when values are over 128.. when they become ff00, I just take the first byte of the byte array. I think it's handy to have on hands.

Neo.SmartContract.Framework/Helper.cs
```diff
@@ -14,6 +14,19 @@ public static class Helper
 
         [Nonemit]
         public extern static byte[] AsByteArray(this string source);
+        
+        public static byte AsByte(this BigInteger source)
+        {
+            if((source > 255) || (source < 0))
+                throw new Exception();
+            return source == 0 ? (byte)0x00 : source.AsByteArray()[0];
```
erikzhang: Can this be compiled?

igormcoelho: Perhaps you're wondering how this avm work, so this is a simpler example:
```cs
        public static byte Main()
        {
            BigInteger x = 255;
            return x.AsByte();
        }
```
```
00c56b02ff0061650700616c756651c56b6c766b00527ac46c766b00c302
ff00a0630d006c766b00c3009f64050061f06c766b00c3009c630f006c76
6b00c300517f616c756600616c7566
```
```
00 PUSH0 
c5 NEWARRAY 
6b TOALTSTACK 
02 PUSHBYTES2 ff00 # creates 255
61 NOP 
65 CALL 0700  # invoke .AsByte(ff00)
61 NOP 
6c FROMALTSTACK 
75 DROP 
66 RET 
# BEGIN CALL .AsByte(ff00)
51 PUSH1  
c5 NEWARRAY 
6b TOALTSTACK 
6c FROMALTSTACK 
76 DUP 
6b TOALTSTACK 
00 PUSH0 
52 PUSH2 
7a ROLL  
c4 SETITEM # local variables is: [ ff00 ]
6c FROMALTSTACK  
76 DUP  
6b TOALTSTACK 
00 PUSH0  
c3 PICKITEM  # load ff00 on stack
02 PUSHBYTES2 ff00 # 255 upper limit
a0 GT  # Returns 1 if value is greater than 255, 0 otherwise.
63 JMPIF 0d00 # will not jump here
6c FROMALTSTACK  
76 DUP 
6b TOALTSTACK  
00 PUSH0  
c3 PICKITEM  # loaded ff00 again on stack
00 PUSH0  # 0 is lower limit
9f LT  # Returns 1 if input value is less than 0, 0 otherwise.
64 JMPIFNOT 0500 # will jump here and skip exception!
61 NOP  
f0 THROW  # will skip this part
6c FROMALTSTACK  # continue here
76 DUP  
6b TOALTSTACK  
00 PUSH0  
c3 PICKITEM  # load ff00 on stack
00 PUSH0  # 0 border case (avoiding empty bytearrays)
9c NUMEQUAL  # check if it's zero
63 JMPIF 0f00 # will not jump here
6c FROMALTSTACK  # continue here
76 DUP  
6b TOALTSTACK  
00 PUSH0  
c3 PICKITEM  # load ff00 on stack
00 PUSH0  
51 PUSH1  
7f SUBSTR  # Returns only first part of bytearray ff00 -> ff
61 NOP  
6c FROMALTSTACK  
75 DROP  
66 RET  # finishes with ff
00 PUSH0  
61 NOP  
6c FROMALTSTACK 
75 DROP  
66 RET 
```
If you're worried about `[0]` part, it works because it uses `SUBSTR` opcode, loading position 0.

igormcoelho: Can you point which part do you think it's a trouble Erik?

Neo.SmartContract.Framework/Helper.cs
```diff
@@ -14,6 +14,19 @@ public static class Helper
 
         [Nonemit]
         public extern static byte[] AsByteArray(this string source);
+        
+        public static byte AsByte(this BigInteger source)
```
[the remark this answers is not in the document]
igormcoelho: Good idea Erik! Use case is related to REVERT and SETITEM on byte[], so we guarantee a single byte is actually passed for attribution.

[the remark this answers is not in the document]
igormcoelho: Good idea Erik! Use case is related to REVERT and SETITEM on byte[], so we guarantee a single byte is actually passed for attribution.

erikzhang:
```
return source > 127 ? (byte)(source - 256) : (byte)source;
```

If you test it well, I'll approve it.

igormcoelho:
> If you test it well, I'll approve it.

Don't worry, I'm always doing my best to test all scenarios. Anyway, I think I have a better proposal now, which is cleaner, using sbyte.

igormcoelho:
Just a short info:
ToByte() -> converts int/BigInteger to byte format
ToSbyte() -> converts int/BigInteger to sbyte format
AsSbyte() -> casts int/BigInteger to sbyte format (with Exception verification)
AsByte() -> casts int/BigInteger to byte format (with Exception verification)

erikzhang:
Looks good to me. Waiting for you testing.

igormcoelho:
Ok Erik, I tested again, on many scenarios, and it works.

However... the inefficiency of this code is bothering me :joy: 
Look, a simple code like this:
```cs
        public static void Main()
        {
            int x = 10;
            byte y = x.AsByte();
            Runtime.Notify(y);
        }
```
is generating 279 bytes avm
```
51c56b5a61656f006c766b00527ac451c576006c766b00c3c46168124e65
6f2e52756e74696d652e4e6f74696679616c756651c56b6c766b00527ac4
6c766b00c361650700616c756651c56b6c766b00527ac46c766b00c3017f
a0630e006c766b00c301809f64050061f06c766b00c3616c756651c56b6c
766b00527ac46c766b00c36165aeff616c756651c56b6c766b00527ac46c
766b00c36165aeff616c756651c56b6c766b00527ac46c766b00c3017fa0
6411006c766b00c3020001946a00527ac46c766b00c361657eff616c7566
51c56b6c766b00527ac46c766b00c36165c0ff616c756651c56b6c766b00
527ac46c766b00c36165a9ff616c756651c56b6c766b00527ac46c766b00
c36165c2ff616c7566
```
This is unnaceptable to me :)

let's keep these method signatures, which are reasonable, and I'll try to improve (a lot) the implementation.

igormcoelho:
Ok, now it's 82 bytes... improved a little bit.
```
51c56b5a61652d006c766b00527ac451c576006c766b00c3c46168124e65
6f2e52756e74696d652e4e6f74696679616c756651c56b6c766b00527ac4
6c766b00c30180028000a5f1756c766b00c3616c7566
```

I will test each of the possible numbers to make sure this works.

igormcoelho:
Ok, tested all possible configurations and wrote the expected values on method documentation. In the end, this sounded simple, but it's amazing now to see how simple things can make byte operations much safer :)

Neo.SmartContract.Framework/Helper.cs
```diff
@@ -6,17 +6,135 @@ namespace Neo.SmartContract.Framework
...
+        /// Converts byte[] to string. Examples: [0x68656c6c6f] -> "hello"; [] -> ""; [0x4e656f] -> "Neo"
+        /// </summary>
         [Nonemit]
         public extern static string AsString(this byte[] source);
+        
+        /// <summary>
+        /// Returns true iff a <= x && x < b. Examples: x=5 a=5 b=15 is true; x=15 a=5 b=15 is false
+        /// </summary>
+        [OpCode(OpCode.WITHIN)]
+        public extern static bool Within(this BigInteger x, int a, int b);
```
erikzhang: Why not `bool Within(this BigInteger x, BigInteger a, BigInteger b)`?

igormcoelho: Ok, updated it, just tested and it works: `return 12.Within(10,15);`

igormcoelho:
Ok, I think this is ready now.
I had to sum zero on all cases to guarantee that non-standard integer inputs (such as 0x000000, or empty array ) are actually converted to a single byte in output. This also helps resolving other strange situations where biginteger may have been unsafely converted, ou unsafely invoked to contract parameter.

igormcoelho:
Ok, I think this is ready now.
I had to sum zero on all cases to guarantee that non-standard integer inputs (such as 0x000000, or empty array ) are actually converted to a single byte in output. This also helps resolving other strange situations where biginteger may have been unsafely converted, ou unsafely invoked to contract parameter.

igormcoelho:
Please, accept this, or I will spend the rest of my life trying to optimize it :joy:

AsByte() and AsSbyte() are currently issuing only 4 opcodes (the function CALL we can inline later with neo-avm-optimizer project):
```
c0 ARRAYSIZE 
51 PUSH1  
9c NUMEQUAL  
f1 THROWIFNOT 
```
ToByte() and ToSbyte() are a little longer:
```
00 PUSH0  #An empty array of bytes is pushed onto the stack
02 PUSHBYTES2 0001 # 
a5 WITHIN  # Returns 1 if x is within the specified range (left-inclusive), 0 otherwise.
f1 THROWIFNOT  #
6c FROMALTSTACK  # Puts the input onto the top of the main stack. Removes it from the alt stack.
76 DUP  # Duplicates the top stack item.
6b TOALTSTACK  # Puts the input onto the top of the alt stack. Removes it from the main stack.
00 PUSH0  #An empty array of bytes is pushed onto the stack
c3 PICKITEM  #
01 PUSHBYTES1 7f # 
a0 GT  # Returns 1 if a is greater than b, 0 otherwise.
64 JMPIFNOT 1100 # 17
6c FROMALTSTACK  # Puts the input onto the top of the main stack. Removes it from the alt stack.
76 DUP  # Duplicates the top stack item.
6b TOALTSTACK  # Puts the input onto the top of the alt stack. Removes it from the main stack.
00 PUSH0  #An empty array of bytes is pushed onto the stack
c3 PICKITEM  #
02 PUSHBYTES2 0001 # 
94 SUB  # b is subtracted from a.
6a DUPFROMALTSTACK  #
00 PUSH0  #An empty array of bytes is pushed onto the stack
52 PUSH2  # The number 2 is pushed onto the stack.
7a ROLL  # The item n back in the stack is moved to the top.
c4 SETITEM  #
6c FROMALTSTACK  # Puts the input onto the top of the main stack. Removes it from the alt stack.
76 DUP  # Duplicates the top stack item.
6b TOALTSTACK  # Puts the input onto the top of the alt stack. Removes it from the main stack.
00 PUSH0  #An empty array of bytes is pushed onto the stack
c3 PICKITEM  #
00 PUSH0  #An empty array of bytes is pushed onto the stack
93 ADD  # a is added to b.
```
But that's fine, I guess :)

PR #30, opened 2018-11-04
Simplified programming using Put, Get and Delete

igormcoelho:
When I first met Neo, I was scared for long time of seing this:
```cs
        public static void Main()
        {
            Storage.Put(Storage.CurrentContext, "Hello", "World");
        }
```
Now my dream will come true, Hello World example will be just:
```cs
        public static void Main()
        {
            Storage.Put("Hello", "World");
        }
```


igormcoelho:
Truth is, I dont have a C# compiler easily, and dont know C# very well :P I usually edit directly on github... so, the template on find was the biggest chance of doing a mistake. But I'll include Find ;)

igormcoelho:
image
Man, look at this prettyness... it works ;)

But one thing I didn't like... `Storage.Put("Hello","World")` generates this:
```
00c56b0548656c6c6f05576f726c64617c650700616c756652c56b6c766b
00527ac46c766b51527ac46168164e656f2e53746f726167652e47657443
6f6e746578746c766b00c36c766b51c3615272680f4e656f2e53746f7261
67652e507574616c7566
```
And `Storage.Put(Storage.CurrentContext, "Hello","World")` generates this:
```
00c56b6168164e656f2e53746f726167652e476574436f6e746578740548
656c6c6f05576f726c64615272680f4e656f2e53746f726167652e507574
616c7566
```
Any hopes we avoid this function call? Inlining it somehow?

igormcoelho:
Well, spend some time trying to manually inline it, invoke two SYSCALL simultaneously (but I'd need to change compiler...), and my conclusion is that: "hello world" users will prefer a simple code that generates bigger avm, than to start from a more complex coding. I think we should open doors to more people start quickly one Neo, and this PR helps in this sense :)

igormcoelho:
I think this is ready to go now @belane

igormcoelho:
I think this is ready to go now @belane

igormcoelho:
For me that's fine to go @vncoelho... one thing I noticed is that AVM will be bigger using Put( "x", "y" ) than Put( context, "x", "y") ... Because the later is not actually invoking the function, the compiler is converting opcode directly, but in simple Put, it will actually call a function. Anyway, I think that's a huge benefit for newbies, because experts can decide to use context to make avm shorter, but for those who arrive, it doesn't matter AVM size, it matters being simple! ;)

shargon:
Did you tested this outside `Main` method, and with dynamic invocations?

belane:
Any chance of producing the same avm? or we need to change the compiler for that?

igormcoelho:
> Did you tested this outside `Main` method, and with dynamic invocations?

Man, go back to your vacations, don't mess with my PR :joy:  No, I didn't test on these situations... but I don't think it should matter, because Put(x,y) corresponds to a Put(getcontext, x, y). Unless CurrentContext is broken after dynamic invoke, it should work fine. But I will test!

> Any chance of producing the same avm? or we need to change the compiler for that?

If you really want us to do that beautifully, this is how we could do it. We change SyscallAttribute on devpack to receive the string as it is and an optional array of syscall parameters. Default will be empty, as usual.

https://github.com/neo-project/neo-devpack-dotnet/blob/master/Neo.SmartContract.Framework/SyscallAttribute.cs#L10-L13

Then, we go to the compiler, and get parameters here together with syscall here:

https://github.com/neo-project/neo-compiler/blob/8cf1c019dfe5dcbb2c10c03c3ed61a3d4716325c/neon/MSIL/Conv_Multi.cs#L193

We take parameter out here:

https://github.com/neo-project/neo-compiler/blob/8cf1c019dfe5dcbb2c10c03c3ed61a3d4716325c/neon/MSIL/Conv_Multi.cs#L422

And here we add  the extra conversions:

https://github.com/neo-project/neo-compiler/blob/8cf1c019dfe5dcbb2c10c03c3ed61a3d4716325c/neon/MSIL/Conv_Multi.cs#L764-L782

However, we will have another problem... the parameter will be added on last position (and we need it to be the first), so probably we will need to inform compiler how to put it in correct position (example: parameter 3 of 5), so we push it to stack position with ROLLS and SWAPS. So, in my mind, this is possible, but not very beautiful from compiler point of view... perhaps there's some straightforward way that I don't know, to make C# don't generate it as method.

shargon:
If you use a static property for it, maybe has problems with dynamic invoke, because will share the same variable for context.

PR #32, opened 2018-11-27
Safe attribution for byte[] element

igormcoelho:
Proposed attribution for byte[] elements. Current byte[] v attribution v[ index ] = b will fail, because compiler considers byte[] a reference type (just like object[]), but neo-vm treats it as a value type.
This PR allows the following attribution to work:
```cs
   byte[] v = // ... initialization
   v[ index ] = b; // compiles, but fails on practice (value is not stored)
   v = v.Set(index, b); // works
```

### DIFF
--- a/Neo.SmartContract.Framework/Helper.cs
+++ b/Neo.SmartContract.Framework/Helper.cs
@@ -7,6 +7,18 @@ namespace Neo.SmartContract.Framework
     public static class Helper
     {
         /// <summary>
+        /// Converts byte to byte[].
+        /// </summary>
+        [Nonemit]
+        public extern static byte[] AsByteArray(this byte source);
+
+        /// <summary>
+        /// Converts sbyte to byte[].
+        /// </summary>
+        [Nonemit]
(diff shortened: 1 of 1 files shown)


erikzhang:
```c#
```

Is this helpful in making these methods inline?

igormcoelho:
For some time I try this AggressiveInlining, but it never worked for me on C#... in fact, it seems there's no way to force a inline, or at least this is the case for all functions I've tried :joy:  I believe we can inline that on AVM level, directly on neo-avm-optimizer project.

erikzhang:
At the end of the article:

> However, methods that couldn’t be inlined previously because of other criteria are still not inlined. I checked the following, and neither of these methods was inlined...

- Recursive method
- Virtual method (even if the static type of the receiver variable is sealed)
- Method with exception handling (representing a “complicated flowgraph”)

igormcoelho:
Well, we still didn't get to CALL handling, but I think it's not far from neoresearch.io/smacco and neoresearch.io/dappgram projects :)
image
image

PR #33, opened 2018-11-28
attribute to issue multiple opcodes

igormcoelho:
Using this attribute it will be possible to issue multiple opcodes simulateously.
Example:
```cs
        [OpCodes(OpCode.PUSH1, OpCode.ADD)]
        public extern static int add1(int x);
```


igormcoelho:
@erikzhang using this we can inline everything, manually if necessary :)

igormcoelho:
Perhaps we keep Nonemit just for clarity and compatibility purposes? I know OpCode empty willl do the job too haha :joy:

PR #34, opened 2018-11-29
Fix: removed two missing nonemit

igormcoelho:


igormcoelho:
@shargon can you take a look?Nonemit does not exist anymore, only [OpCode] now

shargon:
I will merge it because i think this is forgotten from the last Pull Request.

https://github.com/neo-project/neo-devpack-dotnet/commit/9af6e177a0d7c9f8f3b4c7d629e8fe6f81b2c4a0#diff-87cef985c6c94106b4b7f4fc1f22daccR24

PR #35, opened 2018-11-29
Inlining basic operations

igormcoelho:
Now we can manually inline basic operations and guarantee they will be efficient.

AsByte and AsSbyte are more simple:
```
AsSbyte(this int x)  # stack: [ x, ... ]
DUP        # stack top -> [ x, x, ... ]
ARRAYSIZE  # stack top -> [ |x|, x, ... ]
PUSH1      # stack top -> [ 1, |x|, x, ... ]
NUMEQUAL   # stack top -> [ 1 == |x|, x, ... ]
THROWIFNOT # stack top -> [ x, ... ]
```

Set was more complex, but doable with much less opcodes than original.
```
Set(this byte[] v, int x, sbyte b)  # stack: [ b , x , v , ... ]
PUSH02    # stack top -> [ 2, b, x, v, ... ]
PICK      # stack top -> [ v, b, x, v, ... ]
PUSH02    # stack top -> [ 2, v, b, x, v, ... ]
PICK      # stack top -> [ x, v, b, x, v, ... ]
LEFT      # stack top -> [ left_v, b, x, v, ... ]
SWAP      # stack top -> [ b, left_v, x, v, ... ]
CAT       # stack top -> [ left_v_b, x, v, ... ]
ROT       # stack top -> [ v, left_v_b, x, ... ]
ROT       # stack top -> [ x, v, left_v_b, ... ]
OVER      # stack top -> [ v, x, v, left_v_b, ... ]
ARRAYSIZE # stack top -> [ |v|, x, v, left_v_b, ... ]
PUSHM1    # stack top -> [ -1, |v|, x, v, left_v_b, ... ]
ADD       # stack top -> [ |v|-1, x, v, left_v_b, ... ]
SWAP      # stack top -> [ x, |v|-1, v, left_v_b, ... ]
SUB       # stack top -> [ |v|-1-x, v, left_v_b, ... ]
RIGHT     # stack top -> [ right_v, left_v_b, ... ]
CAT       # stack top -> [ left_v_b_right_v, ... ]
```

By having more flexible PACK and UNPACK we can do Set operation with three or four opcodes, in the future.


Neo.SmartContract.Framework/Helper.cs
```diff
@@ -174,10 +179,11 @@ public static byte ToByte(this int source)
         /// <summary>
         /// Safely performs attribution v[x] = b. Faults if x < 0 or x >= v.Length
         /// </summary>
-        public static byte[] Set(this byte[] v, int x, sbyte b)
-        {
-            return v.Take(x).Concat(b.AsByteArray()).Concat(v.Last(v.Length - x - 1));
-        }
+        [OpCode(OpCode.PUSH2, OpCode.PICK, OpCode.PUSH2, OpCode.PICK, OpCode.LEFT, OpCode.SWAP, OpCode.CAT, OpCode.ROT, OpCode.ROT, OpCode.OVER, OpCode.ARRAYSIZE, OpCode.DEC, OpCode.SWAP, OpCode.SUB, OpCode.RIGHT, OpCode.CAT)]
+        public extern static byte[] Set(this byte[] v, int x, sbyte b);
```
shargon: Rename this parameters for made easy to developers please :)

shargon: this is the same as `v[x]=b`?

erikzhang:
It would be better if we can use `v[x]=b;` grammar.

Neo.SmartContract.Framework/Helper.cs
erikzhang: What is your suggestion about renaming? @shargon

Neo.SmartContract.Framework/Helper.cs
```diff
@@ -172,12 +177,13 @@ public static byte ToByte(this int source)
...
         /// <summary>
-        /// Safely performs attribution v[x] = b. Faults if x < 0 or x >= v.Length
+        /// Safely performs attribution v[index] = value. Faults if index < 0 or index >= v.Length
         /// </summary>
-        public static byte[] Set(this byte[] v, int x, sbyte b)
-        {
-            return v.Take(x).Concat(b.AsByteArray()).Concat(v.Last(v.Length - x - 1));
-        }
+        [OpCode(OpCode.PUSH2, OpCode.PICK, OpCode.PUSH2, OpCode.PICK, OpCode.LEFT, OpCode.SWAP, OpCode.CAT, OpCode.ROT, OpCode.ROT, OpCode.OVER, OpCode.ARRAYSIZE, OpCode.DEC, OpCode.SWAP, OpCode.SUB, OpCode.RIGHT, OpCode.CAT)]
+        public extern static byte[] Set(this byte[] v, int index, sbyte value);
```
erikzhang: And `v` to `source`?

igormcoelho: Good one!

igormcoelho:
Erik, I think the best hope is to do something like this:
```cs
    class ByteArray
    {
       private byte[] arr;
       public byte this[int i]
       {
          get { return arr[i]; }
          set { arr[i] = value; }
       }
    }
```
Also Nonemit: ` public extern static ByteArray ToByteArray(this byte[] source);`
I still don't have the precise knowledge, but it generates special MSIL instructions, that we can perhaps capture :smiling_imp: 
`IL_0008: callvirt System.Void Neo.SmartContract.ByteArray::set_Item(System.Int32,System.Byte)`
`IL_000f: callvirt System.Byte Neo.SmartContract.ByteArray::get_Item(System.Int32)`
It's promising, also promising to finally build `Signature`, `PublicKey`, `Hash256` and `Hash160` structures :)

shargon:
Why not use a `DUP` and `SETITEM` ? instead of `CAT`?

igormcoelho:
The problem with SETITEM is that it only works for arrays,  and the main problem is that it CONSUMES the array during the attribution... so DUP SETITEM works because the array shares the reference, but bytearray stack item is copy-based, not reference-based. So, DUP won't have any side effect on the original object. We could try to reimplement SETITEM for byte arrays, but as we discussed in other thread some time ago, Erik correctly pointed out that we would have two different semantics, in one situation it would return a stack item, and in other one it wouldn't. 
I took some time to understand the problem, but I think now we can finally solve it beautifully. MSIL/CIL thinks/defines that all array-based items are reference-based, so it generates stack operations based on that, and it's nearly impossible to predict on compile level where is the actual object when you get  a SETITEM, because at that point, DUP already happened, followed by a series of stack movements.

Now I'm pretty sure we can do a nice ByteArray class (interface, no implementation really needed), together with Signature, Hash256 and Hash160 classes, so we finish the programming pack with all basic features.

shargon:
I think that the best solution is to accept `ByteArray` in `SETITEM` opcode

igormcoelho:
I think I'm nearly getting it.... if we use class or byte[], we are doomed at C# programming level (because it's reference type).
```

  IL_0000: ldarg.0
  
  IL_0001: call Neo.SmartContract.ByteArrayX Neo.SmartContract.Help::ToByteArray(System.Byte[])
  Neo.SmartContract.ByteArrayX Neo.SmartContract.Help::ToByteArray(System.Byte[])
  
  IL_0006: dup
  
  IL_0007: ldc.i4.2
  
  IL_0008: ldc.i4.s 10
  10
  
  IL_000a: callvirt System.Void Neo.SmartContract.ByteArrayX::set_Item(System.Int32,System.Byte)
  System.Void Neo.SmartContract.ByteArrayX::set_Item(System.Int32,System.Byte)
  
  IL_000f: ldc.i4.2
  
  IL_0010: callvirt System.Byte Neo.SmartContract.ByteArrayX::get_Item(System.Int32)
  System.Byte Neo.SmartContract.ByteArrayX::get_Item(System.Int32)
  
  IL_0015: ret
```

Struct may have a different way (value type):
```

  IL_0000: ldarg.0
  
  IL_0001: call Neo.SmartContract.ByteArrayX Neo.SmartContract.Help::ToByteArray(System.Byte[])
  Neo.SmartContract.ByteArrayX Neo.SmartContract.Help::ToByteArray(System.Byte[])
  
  IL_0006: stloc.0
  
  IL_0007: ldloca.s V_0
  V_0
  
  IL_0009: ldc.i4.2
  
  IL_000a: ldc.i4.s 10
  10
  
  IL_000c: call System.Void Neo.SmartContract.ByteArrayX::set_Item(System.Int32,System.Byte)
  System.Void Neo.SmartContract.ByteArrayX::set_Item(System.Int32,System.Byte)
  
  IL_0011: ldloca.s V_0
  V_0
  
  IL_0013: ldc.i4.2
  
  IL_0014: call System.Byte Neo.SmartContract.ByteArrayX::get_Item(System.Int32)
  System.Byte Neo.SmartContract.ByteArrayX::get_Item(System.Int32)
  
  IL_0019: ret
```

igormcoelho:
Too bad... even struct set item [ ] also operates with side effects... now I feel defied, the challenge has been accepted!! 
I'll try to fix native byte[ ] set directly, without any extra classes or structs :muscle:

igormcoelho:
Ok guys, managed to fix most a general case of byte[] attribution directly on compiler: https://github.com/neo-project/neo-compiler/pull/140

So, Set method can be abandoned :)
Other inlines look good.

PR #36, opened 2018-12-01
Allowing multiple syscalls

igormcoelho:
This allows multiple syscalls, example:
```cs
        [Syscall("Neo.Storage.GetContext", "Neo.Storage.Put")]
        Put(string key, string value);
```

Now, Storage Put for hello world will be simple and efficient :)


igormcoelho:
This works, but we will make it better.

PR #38, opened 2018-12-02
Created Inline attribute to mix OpCode/Syscalls/etc

igormcoelho:
See: https://github.com/neo-project/neo-compiler/pull/142

### DIFF
--- a/Neo.SmartContract.Framework/InlineAttribute.cs
+++ b/Neo.SmartContract.Framework/InlineAttribute.cs
@@ -3,22 +3,32 @@ using System;
 
 namespace Neo.SmartContract.Framework
 {
-    // OpCode with data in hex or ascii
+    /// <summary>
+    /// OpCode with data in hex or ascii
+    /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor, AllowMultiple = true)]
     public sealed class InlineAttribute : Attribute
     {
-        // opcode
-        public OpCode opCode { get; }
-        // opcode data (can be hex "ab01ab" or ascii "Runtime.Notify")
-        public string opData { get; }
-        // if opcode data is Hex or ascii
-        public bool isHex { get; }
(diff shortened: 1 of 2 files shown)


igormcoelho:
Example:
```cs
Storage.Put("Hello", "World");
```
generates:
```
01 PUSHBYTES1 61 # a
01 PUSHBYTES1 62 # b
61 NOP  # Does nothing.
7c SWAP  # The top two items on the stack are swapped.
68 SYSCALL 164e656f2e53746f726167652e476574436f6e74657874 # Neo.Storage.GetContext
68 SYSCALL 0f4e656f2e53746f726167652e507574 # Neo.Storage.Put
```

But it can be used with other opcodes too:
```cs
        [Inline(OpCode.SYSCALL, "Neo.Blockchain.GetHeight")]
        [Inline(OpCode.INC)]
        public extern static int HeightPlus1();
```
Generates:
```
68 SYSCALL 184e656f2e426c6f636b636861696e2e476574486569676874 # Neo.Blockchain.GetHeight
8b INC  # 1 is added to the input.
```

We cannot mix Inline attribute with others... example (fails):
```cs
        [Inline(OpCode.SYSCALL, "Neo.Blockchain.GetHeight")]
        [OpCode(OpCode.INC)]
        public extern static int HeightPlus1();
```
or: 
```cs
        [Syscall("Neo.Blockchain.GetHeight")]
        [Inline(OpCode.INC)]
        public extern static int HeightPlus1();
```

It's working for SYSCALL (including parameters) and OpCode (without parameters)... but it could be extended in the future to any OpCode, including APPCALLS, etc. But for now it's enough, I hope :)

erikzhang:
I think you can just change the `OpCodeAttribute` and `SyscallAttribute` to `AllowMultiple = true` without adding `InlineAttribute`.

igormcoelho:
I have to agree, the idea is nice. must how to handle these orders :)

igormcoelho:
Will rethink this.

PR #40, opened 2018-12-05
Mix attributes (OpCall/Syscall)

igormcoelho:
This change allows to mix OpCode attribute with Syscall attribute: https://github.com/neo-project/neo-compiler/pull/144

Example (prints Height and Height+1):
```cs
        [Syscall("Neo.Blockchain.GetHeight")]
        [OpCode(OpCode.INC)]
        public extern static int NextHeight();
        
        public static void Main()
        {
            uint x = Blockchain.GetHeight();
            Runtime.Notify(x);
            Runtime.Notify(NextHeight());
        }
```

### DIFF
--- a/Neo.SmartContract.Framework/OpCodeAttribute.cs
+++ b/Neo.SmartContract.Framework/OpCodeAttribute.cs
@@ -1,16 +1,31 @@
-﻿using Neo.VM;
+using Neo.VM;
 using System;
 
 namespace Neo.SmartContract.Framework
 {
-    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor, AllowMultiple = true)]
     public class OpCodeAttribute : Attribute
     {
-        public OpCode[] OpCodes { get; }
+        /// <summary>
+        /// opcode
+        /// </summary>
+        public OpCode OpCode { get; }
 
-        public OpCodeAttribute(params OpCode[] opcodes)
+        /// <summary>
+        /// opcode data (can be hex "ab01ab" or ascii "Runtime.Notify")
+        /// </summary>
+        public string OpData { get; }
+
+        /// <summary>
+        /// if opcode data is Hex or ascii
+        /// </summary>
+        public bool IsHex { get; }
+
+        public OpCodeAttribute(OpCode opCode, string opData = "", bool isHex = false)
         {
-            this.OpCodes = opcodes;
+            OpCode = opCode;
+            OpData = opData;
+            IsHex = isHex;
         }
     }
 }
--- a/Neo.SmartContract.Framework/Helper.cs
+++ b/Neo.SmartContract.Framework/Helper.cs
@@ -9,32 +9,32 @@ namespace Neo.SmartContract.Framework
         /// <summary>
         /// Converts byte to byte[].
         /// </summary>
-        [OpCode]
+        [Nonemit]
         public extern static byte[] AsByteArray(this byte source);
 
         /// <summary>
         /// Converts sbyte to byte[].
         /// </summary>
-        [OpCode]
+        [Nonemit]
         public extern static byte[] AsByteArray(this sbyte source);
 
         /// <summary>
         /// Converts sbyte[] to byte[].
         /// </summary>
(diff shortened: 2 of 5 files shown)


igormcoelho:
Sorry to dismiss your review Shargon, I needed to make sure this example would also work:

Example 2:
```cs
        [OpCode((OpCode)0x21, "031a6c6fbbdf02ca351745fa86b9ba5a9452d785ac4f7fc2b7548ca2a46c4fcf4a")]
        [OpCode(OpCode.SWAP)]
        [OpCode(OpCode.CHECKSIG)]
        public extern static bool VerifyPubKey(byte[] sig);

        [Syscall("Neo.Blockchain.GetHeight")]
        [OpCode(OpCode.INC)]
        public extern static int NextHeight();

        [OpCode(OpCode.SYSCALL, "Neo.Blockchain.GetHeight", false)]
        [OpCode(OpCode.INC)]
        public extern static int NextHeight2();
```

Basically, this gives us the capability to create highly optimized codes now. If IsHex default was false, that would overcomplicate things (now default is true) ;)

erikzhang:
I have an idea. Why don't we create a `ScriptSnippetAttribute`?

```c#
[ScriptSnippet("68184e656f2e426c6f636b636861696e2e4765744865696768748b")]
public extern static int NextHeight();
```

igormcoelho:
Do you think it's an advantage Erik? I think its less legible than current approach... but for bigger scripts surely would help.

erikzhang:
What about change `Nonemit` to `[OpCode(OpCode.NOP)]`?

igormcoelho:
In fact you are right, ScriptAttribute is a nice idea. We can use [Script("") ] or [Script] to solve this.
In a few minutes I'll update the proposal ;)

igormcoelho:
Nice Erik! Now AsByte() operation is just `[Script("76c0519cf1")]` xD

igormcoelho:
Strange, one test failed here:
`  EXEC : Convert error : System.Exception: error:System.Object Neo.SmartContract.ICO_Template::Main(System.String,System.Object[])::IL_0031 Call System.Byte[] Neo.SmartContract.Framework.Helper::AsByteArray(System.String) ---> System.Exception: neomachine Cannot mix OpCode/Syscall/Script attributes with others!`
I need to verify that.

igormcoelho:
Ok, this is ready now :)

Neo.SmartContract.Framework/OpCodeAttribute.cs
```diff
@@ -1,16 +1,31 @@
...
-        public OpCodeAttribute(params OpCode[] opcodes)
+        /// <summary>
+        /// opcode data (can be hex "ab01ab" or ascii "Runtime.Notify")
+        /// </summary>
+        public string OpData { get; }
+
+        /// <summary>
+        /// if opcode data is Hex (default) or ascii
+        /// </summary>
+        public bool IsHex { get; }
```
erikzhang: We can remove this. If `OpData` can be parsed into a hexadecimal string, then it is hex, otherwise it is ascii.

igormcoelho: Nice! I'll remove IsHex. By the way, I think.compiler is doing Utf8 conversion, but ascii is better, right?

Neo.SmartContract.Framework/OpCodeAttribute.cs
```diff
@@ -1,16 +1,25 @@
...
+        /// <summary>
+        /// opcode
+        /// </summary>
+        public OpCode OpCode { get; }
 
-        public OpCodeAttribute(params OpCode[] opcodes)
+        /// <summary>
+        /// opcode data (can be hex "ab01ab" or ascii "Runtime.Notify")
+        /// </summary>
+        public string OpData { get; }
```
erikzhang: `Operand`?

igormcoelho: I think OpData (Operation Data or OpCode Data) is fine, because it varies.. sometimes operands are on stack (ADD, INC, MUL, ...). But if you prefer, feel free to change, no problem for me.

igormcoelho:
Can we approve this and launch a new version?

lightszero:
I agree. how about change [script] to [avm]

erikzhang:
`avm` is a name that may be modified in the future. Please keep it `Script`.

PR #41, opened 2019-04-13
Change support to dotnet 2.0

igormcoelho:
dotnet 2.0 is the only official support we have for linux (not 4.0), could we adopt this as the standard?

### DIFF
--- a/Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj
+++ b/Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj
@@ -5,7 +5,7 @@
     <AssemblyTitle>Neo.SmartContract.Framework</AssemblyTitle>
     <Version>2.9.3</Version>
     <Authors>The Neo Project</Authors>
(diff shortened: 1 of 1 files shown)


erikzhang:
You need to learn about what's the difference between .NET Standard, .NET Core, and .NET Framework.

igormcoelho:
I agree that I must learn that.. after many months seeing this every day, I still don't have a clue about it.

Anyway, this is how we manage to make things compile on neocompiler.io. 
And now we are porting it to regular users on VS Code too: https://github.com/NeoResearch/neo-unified-compilers

I will put a regular expression to replace the default 1.6/4.0 frameworks to 2.0, so we can compile fine without changing official devpack project.

igormcoelho:
But Erik, from the bottom of my heart, I thank you very much for all these things I've learning, because of Neo and all the great minds around this project. It's really a unique and amazing opportunity. Many good things are coming now, we will move this thing faster and faster!

PR #47, opened 2019-05-20
fix unittest to load muli dll by rosyln

lightszero:
add a project testcontract.builder

compile every single cs file in project testcontract to a standalone dll.


Neo.Compiler.MSIL.UnitTests/UnitTest1.cs
```diff
@@ -41,6 +44,10 @@ public void GetAllILFunction()
         [TestMethod]
         public void TestDumpAFunc()
         {
+            var srcfile = System.IO.Path.GetFullPath("../../../../testdll/TestClass1.dll");
```
shargon: Maybe we could use resource files

TestContract/TestContract.csproj
```diff
@@ -51,6 +52,8 @@
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
     <PostBuildEvent>copy "$(TargetPath)" "$(SolutionDir)Neo.Compiler.MSIL.UnitTests\$(OutDir)netcoreapp2.1"
-copy "$(TargetDir)Neo.SmartContract.Framework.dll" "$(SolutionDir)Neo.Compiler.MSIL.UnitTests\$(OutDir)netcoreapp2.1"</PostBuildEvent>
+copy "$(TargetDir)Neo.SmartContract.Framework.dll" "$(SolutionDir)testdll"
+
+"$(SolutionDir)testcontract.builder\bin\Debug\netcoreapp3.0\testcontract.builder" $(ProjectDir)  $(SolutionDir)testdll\</PostBuildEvent>
```
shargon: netcore3

testcontract.builder/testcontract.builder.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
shargon: We could upgrade all projects.

Neo.Compiler.MSIL.UnitTests/UnitTest1.cs
lightszero: laster will use a json for unittest define.

TestContract/TestContract.csproj
erikzhang: Copy to root solution directory？

testcontract.builder/testcontract.builder.csproj
erikzhang: `netcoreapp3.0` is still in preview, right?

TestContract/TestContract.csproj
lightszero: create a path for gennerate standalone dll.
solutiondir\testdll

testcontract.builder/testcontract.builder.csproj
lightszero: yes,or i will change this to netframework for now.

erikzhang: `2.1` is okey.

lightszero: net core 2.1 need publish to get a exe.
in that way,need to publish it first before run unittest.

Neo.Compiler.MSIL.UnitTests/UnitTest1.cs
erikzhang: Why not use `System.IO.Directory.EnumerateFiles()`?

lightszero: getfiles and enumeratefiles？
getfiles can find file in different subdirs by once call.

testcontract.builder/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+            var ref2 = MetadataReference.CreateFromFile("needlib" + System.IO.Path.DirectorySeparatorChar + "System.dll");
+            var ref3 = MetadataReference.CreateFromFile("needlib" + System.IO.Path.DirectorySeparatorChar + "System.Numerics.dll");
+            var ref4 = MetadataReference.CreateFromFile(frameworkfile);
+            var comp = Microsoft.CodeAnalysis.CSharp.CSharpCompilation.Create("aaa.dll", new[] { tree },
+               new[] { ref1, ref2, ref3, ref4 }, op);
+
+            var file = System.IO.Path.GetFileNameWithoutExtension(filename);
+            var outdllname = System.IO.Path.Combine(outpath, file + ".dll");
+            var outpdbname = System.IO.Path.Combine(outpath, file + ".pdb");
+            var result = comp.Emit(outdllname, outpdbname);
```
erikzhang: We can emit the IL codes into a `byte[]` and pass it to unit test, without exe. Or even we can do the compilation in unit tests directly.

lightszero: Yes we can,but that will waste too must time on il compiler.
so i do il compiler before unittest.
only avm compiler in unittest.

erikzhang: We should re-compile all test contracts before running unit tests. Otherwise, it is possible to forget to compile after modifying the test contracts in the future, and not found the errors in the unit test.

lightszero: now use a post build script on TestContract.
and config unittest depends on this project.

when modifying contracts, it will cause a post build script again.

erikzhang: The downside is that files are generated on disk, and the benefit is savings of 1 to 2 seconds. . .

lightszero: ok,i will try to build dlls to memory in unittest

erikzhang:
@lightszero I removed the `needlib` directory.

PR #49, opened 2019-05-23
Reuse opcodes from Neo.VM

shargon:
We think that we should use the Neo.VM operation codes, if we add new operation codes, they could be outdated if we do not do this.


lightszero:
no,neo.smartcontract.framework is defined for contract.

if you include neo.vm on it,contract will got many unuse class and functions be defined.

and it is be compiled to net40 is for the most simple IL opcode.

erikzhang:
@shargon We can use unit test to ensure that they are the same. For example:

https://github.com/neo-project/neo/blob/1a415b48c48e06fd8f167e7d02d92febe73b0844/neo.UnitTests/UT_OpCodePrices.cs#L12-L16

shargon:
Ok, then we will make this kind of unit test

PR #51, opened 2019-05-24
add json file to config unittest

lightszero:
add json file to config unittest.

### DIFF
--- a/Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
+++ b/Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
@@ -10,6 +10,17 @@
         []
       ],
       "testresult": "(bytes)01020304"
+    },
+    "pickarray": {
+      "build": [ "./TestClasses/Contract2.cs" ],
+
+      "testmethod": "execute",
+      "entryscript": "./TestClasses/Contract2.cs",
+      "testparams": [
+        "(str)test",
+        []
+      ],
+      "testresult": 3
     }
   }
 }
--- a/Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
+++ b/Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
@@ -1,0 +1,15 @@
+{
+  "tests": {
+    "newarray": {
+      "build": [ "./TestClasses/Contract1.cs" ],
+
+      "testmethod": "execute",
+      "entryscript": "./TestClasses/Contract1.cs",
+      "testparams": [
+        "(str)test",
+        []
+      ],
+      "testresult": "(bytes)01020304"
+    }
+  }
+}
--- a/Neo.Compiler.MSIL.UnitTests/Utils/BuildScript.cs
+++ b/Neo.Compiler.MSIL.UnitTests/Utils/BuildScript.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Neo.Compiler.MSIL.Utils
+{
+    class BuildScript
+    {
+        public bool IsBuild
+        {
+            get;
+            private set;
+        }
+        public Exception Error
+        {
+            get;
+            private set;
+        }
+        public ILModule modIL
+        {
+            get;
+            private set;
+        }
+        public ModuleConverter converterIL
+        {
+            get;
+            private set;
+        }
+        public byte[] finalAVM
+        {
+            get;
+            private set;
+        }
+
+        public BuildScript()
+        {
+        }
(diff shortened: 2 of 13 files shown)


Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
```diff
@@ -0,0 +1,15 @@
+{
+  "tests": {
+    "newarray": {
+      "build": [ "./TestClasses/Contract1.cs" ],
+
+      "testmethod": "execute",
+      "entryscript": "./TestClasses/Contract1.cs",
+      "testparams": [
+        "(str)test",
```
shargon: {"type":"String","value":"test"}

Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
```diff
@@ -0,0 +1,15 @@
...
+        []
+      ],
+      "testresult": "(bytes)01020304"
```
shargon: same as neo-vm test will be better

Neo.Compiler.MSIL.UnitTests/TestJsons/test_newarray.json
lightszero: I think this

[
   "(str)transfer",
   ["(address)fromaddrxxxx", 
     "(address)toaddrxxxx",
     12345555]
]

is more clearfiy that

[
     {"type":"string","value":"transfer"},
     [
          {"type":"address","value":"fromaddrxxxx"},
          {"type":"address","value":"toaddrxxxx"},
          {"type":"number","value":12345555},
    ]
]

more params,more clearfiy

shargon: Is more small, but for parsers is better the biggest one. But we should decide what we should use, but the same in all projects.

lightszero: small and keep same structure with the stackitem.but both way is fine for me.
@neo-project/core what do you think?

erikzhang:
What's the difference between using JSON files and just using `Directory.GetFiles()` to get all *.cs files from `TestClasses` directory?

lightszero:
> What's the difference between using JSON files and just using `Directory.GetFiles()` to get all *.cs files from `TestClasses` directory?

json config the testparam and testresult,not for build only.

Neo.Compiler.MSIL.UnitTests/Utils/TestEngine.cs
```diff
@@ -0,0 +1,68 @@
...
+    class TestEngine : ExecutionEngine
+    {
+        IDictionary<string, BuildScript> contracts;
+        public TestEngine(IDictionary<string, BuildScript> contracts)
+        {
+            this.contracts = contracts;
+        }
+        public RandomAccessStack<StackItem> ExecuteTestCase(BuildScript entryscript, StackItem[] _params)
+        {
+            //var engine = new ExecutionEngine();
```
lock9: Remove dead code

Neo.Compiler.MSIL.UnitTests/Utils/base58.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Linq;
+using System.Numerics;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ThinNeo.Cryptography.Cryptography
+{
+    public static class Base58
+    {
```
lock9: Is this class really required? Can't we use some other class that already exists? Or does this class does something different?

Neo.Compiler.MSIL.UnitTests/Utils/BuildScript.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.modIL = new ILModule(log);
+            try
+            {
+                modIL.LoadModule(fs, fspdb);
+            }
+            catch (Exception err)
+            {
+                log.Log("LoadModule Error:" + err.ToString());
+                this.Error = err;
+                return;
```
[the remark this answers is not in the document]
lightszero: This pr is too old,I had read  your suggest.
I will fix this pr soon.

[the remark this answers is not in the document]
lock9: Sorry for the inconvenience @lightszero, I can assure you things will start moving faster.

shargon:
Conflicts

lock9:
@lightszero  we can close this, right? #121

meevee98:
I've tested this pull request and it's working as expected.
The tests are available [here](https://simpli.testquality.com/project/9154/plan/15601/run/122277/result/1678126).

PR #52, opened 2019-05-24
add static var support for neo.compiler.msil

lightszero:
add static var support for neo.compiler.msil

### DIFF
--- a/Neo.Compiler.MSIL/MSIL/Converter.cs
+++ b/Neo.Compiler.MSIL/MSIL/Converter.cs
@@ -124,6 +124,13 @@ namespace Neo.Compiler.MSIL
                         };
                         outModule.mapEvents[ae.name] = ae;
                     }
+                    else if (e.Value.field.IsStatic)
+                    {
+                        var _fieldindex = outModule.mapFields.Count;
+                        var field = new NeoField(e.Key, e.Value.type, _fieldindex);
+                        outModule.mapFields[e.Value.field.FullName] = field;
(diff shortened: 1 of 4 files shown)


Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -1016,17 +1035,43 @@ private int ConvertCode(ILMethod method, OpCode src, NeoMethod to)
...
+
+                            _Insert1(VM.OpCode.PICKITEM,"",to);
+
+                            //throw new Exception("Just allow defined a static variable with readonly." + d.FullName);
                         }
                     }
                     break;
+                case CodeEx.Stsfld:
+                    {
+                        _Convert1by1(VM.OpCode.NOP, src, to);
```
erikzhang: Why `NOP`?

lightszero: insert a nop pos for quick find avm when i debug it.will remove later

shargon: Maybe we can detect if is debug or release?

lightszero: @shargon that idea is good.could be useful

erikzhang: #if DEBUG
    _Convert1by1(VM.OpCode.NOP, src, to);
#endif

PR #53, opened 2019-05-24
unittest for staticvar support

lightszero:
unittest for staticvar support


lightszero:
will add another version without jsonconfig

Issue #56, opened 2019-06-04
Use the same folder structure as neo-vm


shargon:
We should use the same folder structure as `neo-vm`.

```
src/
test/
templates/
```

It is easy to configure travis with this configuration, and is more organized

PR #57, opened 2019-06-04
Reorder folders

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/56


src/Installer/Installer.csproj
```diff
@@ -15,7 +16,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Neo.SmartContract.Installer</RootNamespace>
     <AssemblyName>Neo.SmartContract.Installer</AssemblyName>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
erikzhang: Why change the target framework version?

shargon: it was my fault, thanks!

PR #60, opened 2019-06-04
Travis config

shargon:


templates/Template.CSharp/Template.CSharp.csproj
```diff
@@ -98,5 +98,4 @@
     </BootstrapperPackage>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />
```
erikzhang: Is there no problem with this? I think this should be something that the VS plugin project needs.

shargon: Travis threw an exception because can't find this import, do you think that is needed?

src/Installer/Installer.csproj
```diff
@@ -67,12 +67,11 @@
     </ProjectReference>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />
```
erikzhang: After removing this, the `.vsix` file is no longer automatically generated.

shargon:
I have problems with `Microsoft.VsSDK.targets` and travis ... looks like they are not friends

PR #67, opened 2019-06-21
Branch autogen entrypoint

lightszero:
if you do not write a Main function in smart contract.
now we can add one automatic:

```
    class Contract_autoentrypoint : SmartContract.Framework.SmartContract
    {
        //there is no main here, it can be auto generation.
        //object Main(string name,object[])
        //{
        //    if(name=="call01")
        //    {
        //        return call01();
        //    }
        //    if(name=="call02")
        //    {
        //        call02(_params[0],params[1]);
        //        return 0; //null or "" or 0 or false , if call02 is void,will add return 0;
        //    }
        //    throw //if no method was found,make vm fault.
        //}
        public static byte[] call01()
        {
            var nb = new byte[] { 1, 2, 3, 4 };
            return nb;
        }
        public static void call02(string a,int b)
        {
            Neo.SmartContract.Framework.Services.Neo.Runtime.Log(a);
        }
    }
```

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
@@ -320,8 +320,29 @@ namespace Neo.Compiler.MSIL
                 _Insert1(VM.OpCode.SETITEM, null, to);
             }
         }
+        private void _insertBeginCodeEntry(NeoMethod to)
+        {
+            _InsertPush(2, "begincode", to);
+            _Insert1(VM.OpCode.NEWARRAY, "", to);
+            _Insert1(VM.OpCode.TOALTSTACK, "", to);
+            //移动参数槽位
+            for (var i = 0; i < 2; i++)
+            {
+                //getarray
+                _Insert1(VM.OpCode.FROMALTSTACK, "set param:" + i, to);
+                _Insert1(VM.OpCode.DUP, null, to);
+                _Insert1(VM.OpCode.TOALTSTACK, null, to);
 
-        private void _insertEndCode(ILMethod from, NeoMethod to, OpCode src)
+                _InsertPush(i, "", to); //Array pos
+
+                _InsertPush(2, "", to); //Array item
+                _Insert1(VM.OpCode.ROLL, null, to);
+
+                _Insert1(VM.OpCode.SETITEM, null, to);
+            }
+        }
+
+        private void _insertEndCode(NeoMethod to, OpCode src)
         {
             ////占位不谢
             _Convert1by1(VM.OpCode.NOP, src, to);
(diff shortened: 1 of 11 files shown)


shargon:
wait please

shargon:
Very nice work @lightszero ! i love it! my only concern is about verification, how we can deal with this, i think that it should return false in verification.

shargon:
What about...

```csharp
public static bool onVerification()
{
  // Logic here?
  return false;
}
```

lightszero:
> This idea is really good. Congratulation Lights.
> 
> I reviewed most of this PR in the other PR... but I didn't like that one 😂 hahaha we need testing in both ways: standard and non-standard entrypoints.

this idea is from eric.

lightszero:
> Very nice work @lightszero ! i love it! my only concern is about verification, how we can deal with this, i think that it should return false in verification.

err...,that is a question.

as eric said all neo3 contract is Main(string,object[])

but how about verification?
@erikzhang

igormcoelho:
Congratulations for the idea @erikzhang :joy:

if all is unified, fine.

lightszero:
> Congratulations for the idea @erikzhang 😂
> 
> if all is unified, fine.
> 
> I think your case is not a problem Shargon, because the method would push0 , and conversion already knows its a bool type, and ot would be returned normally. The only "edge" case is a void metgod, which is being treated as "return false", right lights?
> 
> This one is just when you dont declare your main, so it creates for you. The other PR is more problematic.
object Main(string methodname,object[] params）
{
    if(methodname=="xxx")
    {
        xxx();//if is a void xxx(), will add a return 0 too;
       return 0;
    }
}

lightszero:
I think we can merge this first.
and discuss about verification more.

shargon:
I was thinking exactly in this, an attribute!

erikzhang:
For verification, you can write the `Main()` function manually.

erikzhang:
We should also update the templates.

shargon:
> For verification, you can write the Main() function manually.

But by default, should `return false`

igormcoelho:
I think that unused functions should FAULT on Application.

lightszero:
already fault，but fault is not have param in neovm. we can not set a fault reason.

igormcoelho:
I think that FAULT reason is more important for a better Exception handling... in the future.

lightszero:
maybe add a notify before throw.

igormcoelho:
Just to clarify, the problem is the THROW itself... sometimes you want to invoke something, and if it fails, you proceed with other alternative. So, we actually need exception handling. I'm just mentioning here because it would be useful together with this feature here... but perhaps it's time to start thinking on strategies to allow exception handling on vm.

PR #68, opened 2019-06-21
change the way how unittest execute contract

lightszero:
because all contract entrypoint is:  
Main(string methodname, params object[] params);

change unittest call method from

```
ExecuteTestCase(new stackitem[]{methodname,new stackitem[]{param1,param2,...});
```

to

```
GetMethod(methodname).Run(param1,param2,...);
```

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
@@ -320,8 +320,29 @@ namespace Neo.Compiler.MSIL
                 _Insert1(VM.OpCode.SETITEM, null, to);
             }
         }
+        private void _insertBeginCodeEntry(NeoMethod to)
+        {
+            _InsertPush(2, "begincode", to);
+            _Insert1(VM.OpCode.NEWARRAY, "", to);
+            _Insert1(VM.OpCode.TOALTSTACK, "", to);
+            //移动参数槽位
+            for (var i = 0; i < 2; i++)
+            {
+                //getarray
+                _Insert1(VM.OpCode.FROMALTSTACK, "set param:" + i, to);
+                _Insert1(VM.OpCode.DUP, null, to);
+                _Insert1(VM.OpCode.TOALTSTACK, null, to);
 
-        private void _insertEndCode(ILMethod from, NeoMethod to, OpCode src)
+                _InsertPush(i, "", to); //Array pos
+
+                _InsertPush(2, "", to); //Array item
+                _Insert1(VM.OpCode.ROLL, null, to);
+
+                _Insert1(VM.OpCode.SETITEM, null, to);
+            }
+        }
+
+        private void _insertEndCode(NeoMethod to, OpCode src)
         {
             ////占位不谢
             _Convert1by1(VM.OpCode.NOP, src, to);
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
@@ -81,7 +81,6 @@ namespace Neo.Compiler.MSIL
                 _code.debugILCode = src.code.ToString();
             }
 
-
             addr++;
 
             _code.code = code;
@@ -170,7 +169,8 @@ namespace Neo.Compiler.MSIL
             {
                 var call = from.body_Codes[next];
                 if (call.tokenMethod == "System.Numerics.BigInteger System.Numerics.BigInteger::op_Implicit(System.UInt64)")
-                {//如果是ulong转型到biginteger，需要注意
+                {
+                    //如果是ulong转型到biginteger，需要注意
                     ulong v = (ulong)i;
                     outv = v;
                     _ConvertPush(outv.ToByteArray(), src, to);
@@ -178,13 +178,10 @@ namespace Neo.Compiler.MSIL
                 }
             }
 
-
-            {
-                _ConvertPush(i, src, to);
-                return 0;
-            }
-
+            _ConvertPush(i, src, to);
+            return 0;
         }
+
         private int _ConvertPushI4WithConv(ILMethod from, int i, OpCode src, NeoMethod to)
         {
             var next = from.GetNextCodeAddr(src.addr);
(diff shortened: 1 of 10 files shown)


igormcoelho:
@lightszero I think its the opposite, we need to remove 'params' and let it be: string, OBJECT[]
not string, object, object, object, ....

igormcoelho:
The reason is two: we cannot know how many items will be read (because its not an array); many languages will have difficulties in representing "params object[] param"

igormcoelho:
Lights, we also use neon to compile verification contracts, see: neoresearch.io/smacco
not all verifications will have standard entry point I guess (for application perhaps)

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -247,6 +248,94 @@ public NeoModule Convert(ILModule _in, ConvOption option = null)
...
+                        //add params;
+                    }
+                    //call and return it
+                    var c = _Insert1(VM.OpCode.CALL, "", to, new byte[] { 0, 0 });
+                    c.needfixfunc = true;
+                    c.srcfunc = m.Key;
+                    if (m.Value.returntype == "System.Void")
+                    {
+                        _Insert1(VM.OpCode.PUSH0, "", to);
+                    }
```
igormcoelho: `void` will push `empty`.. fine for me, I would do the same :)

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_autoentrypoint.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            var nb = new byte[] { 1, 2, 3, 4 };
+            return nb;
+        }
+        public static void call02(string a,int b)
+        {
+            Neo.SmartContract.Framework.Services.Neo.Runtime.Log(a);
+        }
+    }
+}
```
igormcoelho: Very interesting change.
@rodoufu this change here from @lightszero will allow us to have automatic methods... an implicit `Main` entrypoint.

@lightszero perhaps if not found a method, it should `THROW`... but we can improve in the future too.

tests/Neo.Compiler.MSIL.UnitTests/UnitTest1.cs
```diff
@@ -56,11 +56,10 @@ public void Test_ByteArray_New()
             testengine.AddEntryScript("./TestClasses/Contract1.cs");
 
 
-            StackItem[] _params = new StackItem[] { "testfunc", new StackItem[0] };
-            var result = testengine.ExecuteTestCase(_params);
+            var result = testengine.GetMethod("testfunc").Run();
```
igormcoelho: @lightszero this part I disagree... We still need to be able to test non standard entrypoints... Because of pure verifications, like I told you. See https://neoresearch.io/smacco

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -247,6 +248,94 @@ public NeoModule Convert(ILModule _in, ConvOption option = null)
...
+
+            return name;
+
+        }
+        private void FillEntryMethod(NeoMethod to)
+        {
+            this.addr = 0;
+            this.addrconv.Clear();
+
+            _Insert1(VM.OpCode.NOP, "this is a debug code.", to);
```
shargon: #if DEBUG

src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
```diff
@@ -320,8 +320,29 @@ private void _insertBeginCode(ILMethod from, NeoMethod to)
...
+            _InsertPush(2, "begincode", to);
+            _Insert1(VM.OpCode.NEWARRAY, "", to);
+            _Insert1(VM.OpCode.TOALTSTACK, "", to);
+            //移动参数槽位
+            for (var i = 0; i < 2; i++)
+            {
+                //getarray
+                _Insert1(VM.OpCode.FROMALTSTACK, "set param:" + i, to);
+                _Insert1(VM.OpCode.DUP, null, to);
+                _Insert1(VM.OpCode.TOALTSTACK, null, to);
```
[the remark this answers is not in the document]
lightszero: that's comment,only show in map file,not in avm

tests/Neo.Compiler.MSIL.UnitTests/UnitTest1.cs
lightszero: ok,i will keep them both.

src/Neo.Compiler.MSIL/MSIL/Converter.cs
lightszero: yes, thank you i will add that.

lightszero:
keep both call test way...
infact nothing had be changed,so close this one.

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_autoentrypoint.cs
lightszero: throw,you got it, in other pr.

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -247,6 +248,94 @@ public NeoModule Convert(ILModule _in, ConvOption option = null)
...
+            foreach (var m in this.outModule.mapMethods)
+            {
+                if (m.Value.inSmartContract && m.Value.isPublic)
+                {//add a call;
+                    //get name
+                    calladdrbegin.Add(this.addr);
+                    _Insert1(VM.OpCode.DUPFROMALTSTACK, "get name", to);
+                    _InsertPush(0, "", to);
+                    _Insert1(VM.OpCode.PICKITEM, "", to);
+                    _InsertPush(System.Text.Encoding.UTF8.GetBytes(m.Value.displayName), "", to);
```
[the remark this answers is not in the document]
lightszero: i think we should use utf8 in whole system.
if we need to limit chars in string,we must add some limit code.

PR #73, opened 2019-07-01
Sourcemap [Draft]

melanke:
This is the first attempt working with source map, and I don't think the code is working properly yet, but I am studying [Mozilla implementation](https://github.com/mozilla/source-map) and replicated the most important methods.
The most important method is `SerializeMappings` inside `SourceMapGenerator`.
I am having some problems to make it work with the neo-debugger-tools because I can't find a way to get the original source code relation with the compiled resulting code.

### DIFF
--- a/src/Neo.Compiler.MSIL/SourceMap/Mapping.cs
+++ b/src/Neo.Compiler.MSIL/SourceMap/Mapping.cs
@@ -1,0 +1,12 @@
+namespace Neo.Compiler.SourceMap
+{
+    public class Mapping
+    {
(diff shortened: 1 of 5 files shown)


lightszero:
not finish yet.

lock9:
I think we should pass this to Harry, NGD Seattle is focused in the development of tools. I will send it to him to see if he can add it to his code (I've talked with him and he is doing good progress in the VS Code tool). He is also making changes in the original compiler. We need at least that he send his PR first. I will take care of this.

PR #74, opened 2019-07-27
gennerate sourcemap

lightszero:
add code for gennerate sourcemap（js format） file

https://github.com/neo-project/proposals/pull/48

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/ILModule.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/ILModule.cs
@@ -251,6 +251,7 @@ namespace Neo.Compiler.MSIL
                         {
                             c.debugcode = sp.Document.Url;
                             c.debugline = sp.StartLine;
+                            c.debugcol = sp.StartColumn;
                         }
                         c.InitToken(code.Operand);
                         this.body_Codes.Add(c.addr, c);
@@ -542,6 +543,7 @@ namespace Neo.Compiler.MSIL
         public int addr;
         public CodeEx code;
         public int debugline = -1;
+        public int debugcol = -1;
         public string debugcode;
         public object tokenUnknown;
         public int tokenAddr_Index;
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Common.cs
@@ -17,6 +17,7 @@ namespace Neo.Compiler.MSIL
             {
                 _code.debugcode = comment;
                 _code.debugline = 0;
+                _code.debugcol = 0;
             }
 
             addr++;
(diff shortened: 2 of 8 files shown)


shargon:
i will review it soon, a big step for debugging!

lock9:
Hello @devhawk, would you mind checking this PR to see if you can use it?
Thanks!

devhawk:
Note, we will need a version of this code for master-2.x branch too.

src/Neo.Compiler.MSIL/MSIL/ILModule.cs
```diff
@@ -251,6 +251,7 @@ public ILMethod(ILType type, Mono.Cecil.MethodDefinition method, ILogger logger
                         {
                             c.debugcode = sp.Document.Url;
                             c.debugline = sp.StartLine;
+                            c.debugcol = sp.StartColumn;
```
devhawk: We should only be saving sequence point info if the sp.IsHidden property is false

devhawk:
I've reviewed the output of this change and updated my debugger work to consume the source map format instead of my custom debug into format. I recommend that we **DO NOT** accept this PR at this time for reasons I will detail below. 

To be crystal clear, there is _nothing_ wrong with the implementation that @lightszero has developed. The issue is that the source map format is not sufficient for the needs of building a good debugging experience for NEO smart contracts. I will be leaving a copy of this comment on neo-project/proposals#48. I think that is the best place for discussing what information the debugger needs and what format we should use to store it. 

The core problem is that the source map format isn't really designed to store the information that a NEO-VM debugger needs. Source maps are primarily designed for mapping across text formats, such as mapping between original and minimized JS or from TypeScript to compiled JS. For binary instruction to source mapping, especially with types, the source map format is missing a bunch information that the debugger needs to provide a good experience. 

For example:

* A source map has no way to record parameter/variable information. This means there is no way for the debugger to map alt stack values back to local symbols that will be meaningful to the user. 
* A source map has no way to record any type information. This means the debugger can only rely on the type information the VM has for displaying variable values. However, there are cases where a type in the VM can be interpreted different ways. For example, since integers greater that 16 are stored as byte arrays, there's no way for the debugger to know if the developer is expecting a byte array value to be shown as an integer or not. 
* A PDB sequence point maps an IL instruction to a range of text in a source code file (start line/column and end line/column). Source maps can only map to a specific position (line/column) in the source code file. This means the debugger cannot highlight the range of text the instruction represents the way a C# developer would expect. 
* There is no way for the debugger to know what functions instructions map to if there isn't a mapping entry for that specific instruction. If an instruction falls between two map points, each associated with a different function, how is the debugger supposed to tell which function it's in? It needs to know in order to be able to select the next mapping entry to step to.

devhawk:
Can we close this?

Issue #86, opened 2019-08-21
Add support to 'long' switch statements


shargon:
There is a problem with the `switch` instruction, it seems that works if we have less entries than 6, with more, it doesn't works.

I made a pull request for test this issue at, https://github.com/neo-project/neo-devpack-dotnet/pull/87

if you comment some entries, the test will work

erikzhang:
It is because the CLR uses hash table to optimize the `switch` segment. And there is no native support for that hash table in NeoVM.

shargon:
There is not possible to convert this hash table into a jump table?

Pseudocode:

```csharp
// With hash table
hash= [ 0x0a, 0x0b, 0x0c ] 
JMP(hash[2]);

// Without hash table

0x00 NOP
0x01 JMP 0x0a
0x02 JMP 0x0b
0x03 JMP 0x0c

JMP 0x02


```

shargon:
dnSpy decompile the source as:
```csharp
using System;
using Neo.SmartContract.Framework;

namespace Neo.Compiler.MSIL.TestClasses
{
	// Token: 0x02000017 RID: 23
	internal class Contract_Switch : SmartContract
	{
		// Token: 0x0600005E RID: 94 RVA: 0x00003C78 File Offset: 0x00001E78
		public static object Main(string method, object[] args)
		{
			if (method != null)
			{
				uint num = <PrivateImplementationDetails>.ComputeStringHash(method);
				if (num <= 518729469u)
				{
					if (num <= 434841374u)
					{
						if (num <= 350953279u)
						{
							if (num != 334175660u)
							{
								if (num == 350953279u)
								{
									if (method == "19")
									{
										return 20;
									}
								}
							}
							else if (method == "18")
							{
								return 19;
							}
						}
						else if (num != 401286136u)
						{
							if (num != 418063755u)
							{
								if (num == 434841374u)
								{
									if (method == "16")
									{
										return 17;
									}
								}
							}
							else if (method == "15")
							{
								return 16;
							}
						}
						else if (method == "14")
						{
							return 15;
						}
					}
					else if (num <= 468396612u)
					{
						if (num != 451618993u)
						{
							if (num == 468396612u)
							{
								if (method == "10")
								{
									return 11;
								}
							}
						}
						else if (method == "17")
						{
							return 18;
						}
					}
					else if (num != 485174231u)
					{
						if (num != 501951850u)
						{
							if (num == 518729469u)
							{
								if (method == "13")
								{
[...]

PR #87, opened 2019-08-21
Switch unit test

shargon:
Test for https://github.com/neo-project/neo-devpack-dotnet/issues/86


shargon:
I changed the test for pass, if you want to test it, just uncomment

erikzhang:
Let's add the support to `switch`.

Issue #88, opened 2019-08-21
StorageMap optimizations


shargon:
In every get, put, or delete, we convert the `string` to `byte[]` and concat it with the prefix `0x00`, is better to concat this at the initialization, only once.

Also should be added two overloads, for allow one `byte` and `byte[]` parameters

https://github.com/neo-project/neo-devpack-dotnet/blob/cbb3ca1518a7d3757456cced98f1d45b0e2b1b02/src/Neo.SmartContract.Framework/Services/Neo/Helper.cs#L18-L72

erikzhang:
Why we need those `0`s?

shargon:
We don't need it, i just copied it, but we can remove this 0 too

shargon:
Shall i remove these 0s?

erikzhang:
If they are useless, we should remove them.

shargon:
Done

PR #89, opened 2019-08-21
Optimized StorageMap

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/88


erikzhang:
What's this?

src/Neo.SmartContract.Framework/Services/Neo/Helper.cs
```diff
@@ -9,67 +9,85 @@ public static StorageMap CreateMap(this StorageContext context, string prefix)
...
+                Prefix = prefix
+            };
+        }
+
+        public static StorageMap CreateMap(this StorageContext context, byte prefix)
+        {
+            return new StorageMap
+            {
+                Context = context,
+                Prefix = new byte[] { prefix }
```
erikzhang: Can this be compiled?

shargon: I will add some unit test

shargon:
As @erikzhang said, changes was required for byte to byte[] conversion, also was upgraded the `TestEngine` for allow to test the storage access, useful for the `NEP5` template tests

https://github.com/neo-project/neo-devpack-dotnet/pull/90

shargon:
I changed the unit test to not send the prefix as an argument because it is more real, the byte is considered as an integer, so 0x00 is added at the end.

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -8,7 +8,8 @@ public static class Helper
         /// <summary>
         /// Converts byte to byte[].
         /// </summary>
-        [Script]
+        [OpCode(OpCode.PUSH1)]
+        [OpCode(OpCode.LEFT)]
         public extern static byte[] AsByteArray(this byte source);
```
shargon: This method was wrong, because it returns 0x00 at the end, because is considered as a BigInteger. Shall we fix this on Neo2? @neo-project/core ?

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -8,7 +8,8 @@ public static class Helper
         /// <summary>
         /// Converts byte to byte[].
         /// </summary>
-        [Script]
+        [OpCode(OpCode.PUSH1)]
+        [OpCode(OpCode.LEFT)]
```
erikzhang: `ToByteArray()`?

src/Neo.SmartContract.Framework/Helper.cs
erikzhang: I think it is the expected result for `AsByteArray()`. We can add a new method `ToByteArray()` for converting.

src/Neo.SmartContract.Framework/Services/Neo/Helper.cs
```diff
@@ -9,67 +9,85 @@ public static StorageMap CreateMap(this StorageContext context, string prefix)
...
+        public static StorageMap CreateMap(this StorageContext context, byte[] prefix)
+        {
+            return new StorageMap
+            {
+                Context = context,
+                Prefix = prefix
+            };
+        }
+
+        public static StorageMap CreateMap(this StorageContext context, byte prefix)
```
shargon: I can't find a optimized version of the conversion from `byte` to `byte array` inside the VM, should we remove this method?

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -6,11 +6,17 @@ namespace Neo.SmartContract.Framework
     public static class Helper
     {
         /// <summary>
-        /// Converts byte to byte[].
+        /// Converts byte to byte[] considering the byte as a BigInteger (0x00 at the end)
         /// </summary>
-        [Script]
```
erikzhang: Don't forget the `[Script]`.

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -6,11 +6,17 @@ namespace Neo.SmartContract.Framework
...
         public extern static byte[] AsByteArray(this byte source);
 
+        /// <summary>
+        /// Converts byte to byte[].
+        /// </summary>
+        [OpCode(OpCode.PUSH1)]
+        [OpCode(OpCode.LEFT)]
+        public extern static byte[] ToByteArray(this byte source);
```
erikzhang: Can we move this method to the end of the file so that it can be together with other `To` prefixed methods?

src/Neo.SmartContract.Framework/Helper.cs

PR #93, opened 2019-08-24
ReadOnly methods

shargon:
Required for https://github.com/neo-project/neo/pull/1052

Fix some issues:
- `entrypoint` to `entryPoint` (from string to method)
- `functions` to `methods`
- `returntype` to `returnType`
- Unit test for cover all the abi


src/Neo.Compiler.MSIL/FuncExport.cs
```diff
@@ -81,16 +81,18 @@ public static MyJson.JsonNode_Object Export(NeoModule module, byte[] script)
             outjson.SetDictValue("hash", sb.ToString());
 
             //entrypoint
-            outjson.SetDictValue("entrypoint", "Main");
+            outjson.SetDictValue("entryPoint", "Main");
```
igormcoelho: why change this?

src/Neo.Compiler.MSIL/NeoModule.cs
```diff
@@ -1,267 +1,306 @@
...
-            get;
-            private set;
-        }
-        public override string ToString()
-        {
-            return type + " " + name;
-        }
-    }
-}
+﻿using Mono.Cecil;
```
igormcoelho: We should clean this first in another PR @shargon... is there anything changed here in this class?

src/Neo.Compiler.MSIL/FuncExport.cs
shargon: because in the neo/manifest/abi we read these vars like this

src/Neo.Compiler.MSIL/NeoModule.cs
shargon: Yes, something was changed

shargon: I prefer to clean after this, because this PR is needed and is waiting flr 2 days now. We can clean the files later, we need these changes for preview1

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_ABI.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    [TestClass]
+    public class UnitTest_ABI
+    {
+        [TestMethod]
+        public void Test_ABI()
+        {
+            var testengine = new TestEngine();
+            testengine.AddEntryScript("./TestClasses/Contract_Abi.cs");
+            string expectABI = @"{""hash"":""0x77811b3127dea2df1a18230f91396fbcf8c648f4"",""entryPoint"":""Main"",""methods"":[{""name"":""Main"",""parameters"":[{""name"":""method"",""type"":""String""},{""name"":""args"",""type"":""Array""}],""returnType"":""Void""},{""name"":""readOnlyTrue"",""parameters"":[],""returnType"":""Void""},{""name"":""readOnlyFalse1"",""parameters"":[],""returnType"":""Void""},{""name"":""readOnlyFalse2"",""parameters"":[],""returnType"":""Void""}],""readOnlyMethods"":[""readOnlyTrue""],""events"":[{""name"":""transfer"",""parameters"":[{""name"":""arg1"",""type"":""ByteArray""},{""name"":""arg2"",""type"":""ByteArray""},{""name"":""arg3"",""type"":""Integer""}]}]}";
```
igormcoelho: could we put this in multiple lines? 
it's very hard to read and understand this without formatting...

shargon: Is exported without line break so i need to put the same format

src/Neo.Compiler.MSIL/NeoModule.cs
igormcoelho: Let's merge the other before 24 hours.

igormcoelho:
I'll have to reproduce your commits one by one... the conflicting file is exactly the most complicated one........ another excellent reason for doing 94 before, it would be impossible to review.

igormcoelho:
Can you merge @shargon? It's about 5 blocks of code... for me I'm blind, have to understand everything step by step to fix this... perhaps you can do it quicker, as you know the changes.

shargon:
@igormcoelho is done

src/Neo.Compiler.MSIL/NeoModule.cs
```diff
@@ -26,13 +35,7 @@ public byte[] Build()
             return bytes.ToArray();
             //将body链接，生成this.code       byte[]
             //并计算 this.codehash            byte[]
-        }
-        public string mainMethod;
-        public ConvOption option;
-        public Dictionary<string, NeoMethod> mapMethods = new Dictionary<string, NeoMethod>();
-        public Dictionary<string, NeoEvent> mapEvents = new Dictionary<string, NeoEvent>();
-        public Dictionary<string, NeoField> mapFields = new Dictionary<string, NeoField>();
-        //public Dictionary<string, byte[]> codes = new Dictionary<string, byte[]>();
```
shargon: Was moved all the variables at the top of the class

igormcoelho: Yes, let's undo this.

src/Neo.Compiler.MSIL/NeoModule.cs
```diff
@@ -224,10 +267,6 @@ public MyJson.JsonNode_ValueString GenJson()
             }
             return new MyJson.JsonNode_ValueString(info);
         }
-
-        public void FromJson(MyJson.JsonNode_Object json)
-        {
-        }
```
igormcoelho: Really removed?

src/Neo.Compiler.MSIL/NeoModule.cs
shargon: Why?

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_ABI_Event.cs
```diff
@@ -1,25 +0,0 @@
-using Microsoft.VisualStudio.TestTools.UnitTesting;
```
igormcoelho: What happened to this file?

src/Neo.Compiler.MSIL/NeoModule.cs
```diff
@@ -118,10 +120,6 @@ public MyJson.JsonNode_Object GenJson()
             return json;
         }
 
-        public void FromJson(MyJson.JsonNode_Object json)
-        {
-        }
-
```
igormcoelho: Really removed?

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_ABI_Event.cs
shargon: Was renamed

src/Neo.Compiler.MSIL/NeoModule.cs
shargon: This methods are not used

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_ABI.cs
shargon: It would be nice to parse this using the neo class, for ensure that fit the json standard

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_ABI_Event.cs
shargon: Before was checking only the events, now whole the abi

src/Neo.Compiler.MSIL/NeoModule.cs
```diff
@@ -26,13 +35,7 @@ public byte[] Build()
...
-        public string mainMethod;
-        public ConvOption option;
-        public Dictionary<string, NeoMethod> mapMethods = new Dictionary<string, NeoMethod>();
-        public Dictionary<string, NeoEvent> mapEvents = new Dictionary<string, NeoEvent>();
-        public Dictionary<string, NeoField> mapFields = new Dictionary<string, NeoField>();
-        //public Dictionary<string, byte[]> codes = new Dictionary<string, byte[]>();
+        } //public Dictionary<string, byte[]> codes = new Dictionary<string, byte[]>();
         //public byte[] GetScript(byte[] script_hash)
         //{
         //    string strhash = "";
```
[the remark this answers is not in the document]
shargon: I leave it because maybe @lightszero require it in the future, but a general clean will come later, i need to order this classes xD

shargon:
ABI was fixed because it did not fit to the standard

tests/Neo.Compiler.MSIL.UnitTests/Neo.Compiler.MSIL.UnitTests.csproj
```diff
@@ -1,27 +1,28 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
shargon: @igormcoelho ,This file is supposed that was fixed ?

erikzhang:
It modifies the format of ABI?

shargon:
Here was different to neo repository, here entryPoint was a string, and in neo is a methodDescriptor

erikzhang:
Can you give an example of the output json from the changes?

shargon:
This is the final format (was adjusted considering that neo have the good one)
```csharp
{
    "hash": "0x77811b3127dea2df1a18230f91396fbcf8c648f4",
    "methods": [
        {
            "name": "readOnlyTrue",
            "parameters": [],
            "returnType": "Void"
        },
        {
            "name": "readOnlyFalse1",
            "parameters": [],
            "returnType": "Void"
        },
        {
            "name": "readOnlyFalse2",
            "parameters": [],
            "returnType": "Void"
        }
    ],
    "readOnlyMethods": [
        "readOnlyTrue"
    ],
    "entryPoint": {
        "name": "Main",
        "parameters": [
            {
                "name": "method",
                "type": "String"
            },
            {
                "name": "args",
                "type": "Array"
            }
        ],
        "returnType": "Void"
    },
    "events": [
        {
            "name": "transfer",
            "parameters": [
                {
                    "name": "arg1",
                    "type": "ByteArray"
                },
                {
                    "name": "arg2",
                    "type": "ByteArray"
                },
                {
                    "name": "arg3",
                    "type": "Integer"
                }
            ]
        }
    ]
}
```

erikzhang:
I think we don't have `readOnlyMethods` in ABI specification.

shargon:
These changes are precisely for this https://github.com/neo-project/neo/issues/927

erikzhang:
Maybe we can output in this format:

```json
{
    "methods": [
        {
            "name": "readOnlyTrue",
            "parameters": [],
            "returnType": "Void",
            "readonly": true,
        }
    ]
}
```

Or just not modify the ABI, we keep `readonly` in the manifest.

shargon:
I prefer to modify the abi, is more readable, i like your way

erikzhang:
Okay. Let's amend the NEP-3 specification first.

igormcoelho:
Fully agree on `"readonly": true`. Just one camel case standard... shouldn't it be `readOnly:true` ? Let's ammend NEP-3.

shargon:
@igormcoelho could amend the NEP-3? and as you said, yes, must be in camel case

shargon:
Changes done, require to update NEP-3 and NEO manifest before merge

PR #94, opened 2019-08-25
Check dotnet format in travis

shargon:
This will prevent to push any format error

Related to https://github.com/neo-project/neo/issues/982


shargon:
Works as expected

shargon:
@igormcoelho it seems that CRLF is checked here, isn't it?

igormcoelho:
Let's format this before 93... it's too complicated to review. This code was never properly indented :joy:

igormcoelho:
@shargon can you reopen this in your branch? I cannot write over neo directly, I don't know why...

igormcoelho:
Ok, I created another repo clone with write access here. Let's fix travis before 94, so changes will be much easier to track (compiler is tricky).

igormcoelho:
Man, this code is so non-formatted... I don't know what to do here. Let's just pass these basic changes for now... and we format it in time perhaps...

shargon:
> If tests pass, we should merge this ASAP (24hrs and counting).

image

shargon:
We should unify **in a future** the name of the folders across all the projects

igormcoelho:
Yes.. in fact this one I like more: `src` and `tests`. Just a typo causing headache here :joy:

shargon:
PASSED!

igormcoelho:
double-check is always good precaution :+1:
:joy:

igormcoelho:
Sorry guys, we need to review 93. This is blocking us.

PR #95, opened 2019-08-28
Added new syscalls - GetInvocationCounter and GetNotifications

igormcoelho:
@belane you will need this for NEP-5


igormcoelho:
Can you please try to use it @belane? To see if it's responding correctly... like first invocation of kind.. second invocation.. etc.

igormcoelho:
Ah, it's breaking by format now... could you run a `dotnet format` for us @belane? when you have time please, during your tests.

shargon:
Also, could you add the notifications?

shargon:
`GetNotifications` was added, but require to be tested because i am not sure if it works xD

igormcoelho:
> Also, could you add the notifications?

Yes.. I'll have to open another PR, because this one started on the browser, it's not on branch, it's hard for me (and everybody) to keep editing this. This one is simpler, the other may be more complex...

igormcoelho:
On `GetNotifications`, best solution now, in my opinion, is to return an `object[]`, where first element is scripthash, and second is state. This corresponds to the internals used right now. A better solution is, in the future, create an Interop object just for notifications, so we can improve devpack a little bit more.

src/Neo.SmartContract.Framework/Services/Neo/Notification.cs
```diff
@@ -0,0 +1,8 @@
+namespace Neo.SmartContract.Framework.Services.Neo
+{
+    public class Notification : IScriptContainer
```
igormcoelho: I don't think this works, and it doesn't makes sense to inherit from IScriptContainer, bacause Notification is not a ScriptContainer.

src/Neo.SmartContract.Framework/Services/Neo/Notification.cs
```diff
@@ -0,0 +1,8 @@
...
+    {
+        public byte[] ScriptHash { get; }
+        public object State { get; }
```
igormcoelho: On the future we can do this, but we need to create an Interop class on Neo, so that we can actually access the fields in a named format, like this. Right now, it's just an Array of two elements.

erikzhang: I think it is working.

src/Neo.SmartContract.Framework/Services/Neo/Runtime.cs
```diff
@@ -14,6 +14,15 @@ public static class Runtime
...
         }
 
+        public static extern uint InvocationCounter
+        {
+            [Syscall("System.Runtime.GetInvocationCounter")]
+            get;
+        }
+
+        [Syscall("System.Runtime.GetNotifications")]
+        public static extern Notification[] GetNotifications(byte[] hash);
```
erikzhang: If I want to get all the notifications, I have to pass `new byte[20]` to the method? Can I pass `null`?

igormcoelho: Right now, it cannot be null Erik.. I agree, it should support null, thus not expecting an UInt160 at the other side, but a byte[]. I think it's better than requiring a bunch of zeroes on UInt160.

igormcoelho: https://github.com/neo-project/neo/pull/1074/files

erikzhang:
@igormcoelho I think we can revert the last commit. I believe that returning `Notification`s is also working. Let's test it.

igormcoelho:
Erik, I agree it could work, but it's a very unsafe side effect to rely on. If we want Notification to have its own life, we should create an Interop object for it, not treat it as an Array.

For me it's good to revert, I'm just saying, it's not 100% correct in my opinion.

erikzhang:
What risk do you mean?

We convert `class` and `array` into `Array` in NeoVM, and convert `struct` into `Struct`.

igormcoelho:
> We convert class and array into Array in NeoVM

I never thought this was official, but a side effect. In this case I'll revert the commit.

shargon:
We need to deal with the class, because is easier for devs.

shargon:
Maybe need to be a struct?

igormcoelho:
@belane please confirm that this works for NEP-5 template... as Erik said, if class and array are equivalent for NeoVM, this should be fine.

shargon:
I will make some unit test for both syscalls

shargon:
Added unit test, it not works :( but you can play with it xD
@lightszero any idea ?

src/Neo.SmartContract.Framework/Services/Neo/Runtime.cs
shargon: We should to a new method for this null, 
```
public static extern Notification[] GetNotifications();
```
is possible to define a default parameter?

shargon:
The problem was because property is a method, with variables, works perfect!

igormcoelho:
Are tests still broken? I'll take a look.

shargon:
Formatted done!

vncoelho:
Still something:
```cs
 Error Message:
   CollectionAssert.AreEqual failed. (Different number of elements.)
  Stack Trace:
     at Neo.SmartContract.Framework.UnitTests.UT_OpcodeTest.TestAllOpcodes() in /home/travis/build/neo-project/neo-devpack-dotnet/tests/Neo.SmartContract.Framework.UnitTests/UT_OpcodeTest.cs:line 17
```

shargon:
https://github.com/neo-project/neo-devpack-dotnet/pull/99/files solve the issue, but i think that UTF8 was not cleaned in this project

shargon:
Is fixed (i think) Thanks @vncoelho for the format

erikzhang:
> We used to consider that struct is also a class when programming in C++. How is it in C#?

They are quite different in C#.

src/Neo.SmartContract.Framework/Services/Neo/Runtime.cs
erikzhang:
```suggestion
        public static extern Notification[] GetNotifications(byte[] hash = null);
```

src/Neo.SmartContract.Framework/Services/Neo/Notification.cs
```diff
@@ -0,0 +1,15 @@
+namespace Neo.SmartContract.Framework.Services.Neo
+{
+    public class Notification : IApiInterface
+    {
+        /// <summary>
+        /// Sender script hash
+        /// </summary>
+        public byte[] ScriptHash;
```
erikzhang: `readonly`?

src/Neo.SmartContract.Framework/Services/Neo/Notification.cs
```diff
@@ -0,0 +1,15 @@
...
+
+        /// <summary>
+        /// Notification's state
+        /// </summary>
+        public object State;
```
erikzhang: `readonly`?

src/Neo.SmartContract.Framework/OpCode.cs
```diff
@@ -640,18 +642,8 @@ public enum OpCode : byte
         /// </summary>
         WITHIN = 0xA5,
 
-
-        // Crypto
-        //RIPEMD160 = 0xA6, // The input is hashed using RIPEMD-160.
-        /// <summary>
-        /// The input is hashed using SHA-1.
-        /// </summary>
-        SHA1 = 0xA7,
```
vncoelho: Onhhh, here it goes.

vncoelho: @shargon, maybe leave the SHA's for being modified in another PR, no?
Because it may be needed to add a syscall for them, yes?

shargon: We need to remove this opcodes for unit test pass

igormcoelho: Ok, we take them back in a next PR.

PR #99, opened 2019-08-29
Remove Crypto opcodes from dev-pack

shargon:


shargon:
@igormcoelho please check why all changed because is without BOM :S

shargon:
Is formated with dotnet-format

PR #102, opened 2019-09-01
Update nugets

shargon:
There are projects with different version of dotnet or .net standard

### DIFF
--- a/src/Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
+++ b/src/Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
@@ -5,7 +5,7 @@
     <AssemblyTitle>Neo.Compiler.MSIL</AssemblyTitle>
     <Version>2.4.1</Version>
     <Authors>The Neo Project</Authors>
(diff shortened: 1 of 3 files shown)


erikzhang:
I think it's meaningless.

shargon:
I think that all projects should share the same version of the framework

erikzhang:
As a library, the lower the version of the framework, the better.

shargon:
Then, we need to review if any security issue was fixed in any of this upgrades

erikzhang:
> Then, we need to review if any security issue was fixed in any of this upgrades

No need to review. The framework version is not the runtime version.

shargon:
Let's just change it for only update the nugets

erikzhang:
I can't see the benefits of upgrading the dependencies of unit tests.

shargon:
I ussally have the obsession of have the last packages 😂

erikzhang:
Let's upgrade the dependencies only when necessary. Because old dependencies work very well, upgrades can have unpredictable compatibility issues. (although this probability is very low)

Issue #83, opened 2019-08-09
There should not be separate versions of NEON for NEO 2 and 3


devhawk:
having separate versions of NEON targeting NEO 2's VM and NEO 3's VM is going to be difficult for developers to manage. Instead of separate binaries, there should be a command-line switch specifying which VM version the developer is targeting. 

As a suggestion, we could follow the model the C# compiler uses for specifying versions. The [langversion](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/langversion-compiler-option) command-line switch allows users to specify an exact version like "7.3" or versions that can change between releases such as "Latest".

lock9:
I agree. Maybe we can read it from the manifest file (not sure if we already added that information there)

igormcoelho:
We need to see if this is feasible @devhawk... perhaps changes for Neo 3 will require too many changes on compiler, even structural ones. This will depend on next steps on VM side.

devhawk:
@lightszero and I discussed this issue over lunch. We agreed between the two of us that the best developer experience would be to have a single NEON tool that worked for NEO 2 and 3. This does make the NEON tool more complex than two separate tools, but provides a better experience for the developer using the tool.

My thinking is that by default NEON would target NEO 2 and developers could opt into NEO 3 via a command line switch (to be designed). This approach is used in other compilers :

* [C# langversion option](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/langversion-compiler-option)
* [MSVC std option](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version)
* [Clang std option](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-std)

@igormcoelho Even if there are significant changes in the VM, we can still build a single tool to target both NEO 2 and 3. Even if NEO VM 2 and 3 diverge drastically, we could still have a single front end command line tool with a back end code generator for each version of NEO. I'm using a similar architecture in neo-express so that I will be able to add NEO 3 support w/o having separate versions of the tool.

shargon:
But you are teaching different things to the developers, you are teaching, for the compiler you can use the same tool, for the node don't, for the explorer don't, for the wallet don't... strange, i think that all separated is not a bad idea.

If the user need to change the expected behavior according to the project, is confusing

devhawk:
[...]

PR #103, opened 2019-09-03
dotnet new templates

devhawk:
.NET core introduces a new [template engine](https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates) for new projects and items. This PR adds a project for creating a NEO smart contract project NuGet package for the new template engine


erikzhang:
Should we remove the vsix templates?

devhawk:
Personally, I don't think we should remove the VSIX templates. Do we have any numbers on how many developers have installed them?

erikzhang:
https://marketplace.visualstudio.com/items?itemName=neo.AntSharesSmartContract-18032

This is the plugin page.

devhawk:
As long as the neo contract vsix is available from marketplace, we probably shouldn't remove the associated source code. Can we consider vsix template removal as a separate task from this PR?

templates/dotnet-templating/neo-contract-templates.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageType>Template</PackageType>
+    <PackageVersion>1.0.0</PackageVersion>
```
erikzhang: Should this version be the same as the smart contract framework?

templates/dotnet-templating/templates/csharp/Directory.Build.targets
```diff
@@ -0,0 +1,13 @@
...
+  <!-- 
+  Note, because of a bug in the NEON compiler, neon-de can only run during the publish target
+  https://github.com/neo-project/neo-devpack-dotnet/issues/78  
+  <Target Name="NEONBuld" AfterTargets="Build" Inputs="$(TargetPath)" 
+          Outputs="$(TargetDir)$(TargetName).avm;$(TargetDir)$(TargetName).abi.json;$(TargetDir)$(TargetName).md.json">
+    <Exec WorkingDirectory="$(TargetDir)" Command="neon-de $(TargetFileName)"/>
+  </Target> -->
+  <Target Name="NEONPublish" AfterTargets="Publish" Inputs="$(PublishDir)" 
+          Outputs="$(PublishDir)$(TargetName).avm;$(TargetDir)$(TargetName).abi.json;$(TargetDir)$(TargetName).md.json">
+    <Exec WorkingDirectory="$(PublishDir)" Command="neon-de $(TargetFileName)"/>
```
erikzhang: What is "neon-de"?

devhawk: Neon-de is a temporary fork of the neon compiler that also emits information used to drive the debugger experience. Once the debug info format is finalized, this fork will be merged into master.

templates/dotnet-templating/neo-contract-templates.csproj
devhawk: I don't think so. If we add new services, we would have to update the framework version but probably wouldn't update the template

templates/dotnet-templating/templates/csharp/Directory.Build.targets
devhawk: I can update the template to default to neon, but will also provide an opt-in mechanism to use neon-de instead

devhawk:
FYI, there is an open issue tracking the use of the new templates in Visual Studio: https://github.com/dotnet/templating/issues/481. When this work is complete, we can probably remove the VSIX templates

templates/dotnet-templating/templates/csharp/Directory.Build.targets
devhawk: with the latest update, `dotnet new neo-contract` will create a project using neon. `dotnet new neo-contract -de` will create a project using neon-de

erikzhang: Will `dotnet new neo-contract` create c# or vb project?

devhawk: there are C# and VB versions of the neo-contract template. dotnet new defaults to C#, but the developer can specify they want VB via the `-lang VB` command line option. So `dotnet new neo-contract` uses the C# template and `dotnet new neo-contract -lang VB` uses the VB contract

erikzhang:
Do we also need to update the .sln file?

devhawk:
which sln file?

erikzhang:
https://github.com/neo-project/neo-devpack-dotnet/blob/master/neo-devpack-dotnet.sln

devhawk:
I don't think we need solution files for the templates. Well, I don't know much about vsix templates, but I'm not sure what additional value sln files provide for the dotnet new templates over and above the existing project files.

devhawk:
I would recommend removing the VSIX template projects and the VSIX template installer project out of the root SLN file. If you're working on the compiler, you don't really care about building templates + these project require VS extensiblity workload to be installed. We can either create a new SLN file at the root or in the templates directory

devhawk:
SInce this PR has been open so long, I created a new repo to host these templates https://github.com/ngdseattle/neo-templates

shargon:
Sorry for the delay, I will test it asap, I think that it's a very important feature.

templates/dotnet-templating/templates/csharp/Directory.Build.targets
shargon: The problem is that in this repository there are no `neon-de`

templates/dotnet-templating/templates/visualbasic/Directory.Build.targets
```diff
@@ -0,0 +1,15 @@
...
+  Note, because of a bug in the NEON compiler, neon can only run during the
+  publish target when using dotnet standard
+  https://github.com/neo-project/neo-devpack-dotnet/issues/78  
+  -->
+  <Target Name="NEONPublish" AfterTargets="Publish" Inputs="$(PublishDir)" 
+          Outputs="$(PublishDir)$(TargetName).avm;$(TargetDir)$(TargetName).abi.json;$(TargetDir)$(TargetName).md.json">
+<!--#if (EnableDebugEnhancements) -->
+    <Exec WorkingDirectory="$(PublishDir)" Command="neon-de $(TargetFileName)"/>
+<!--#else -->
+    <Exec WorkingDirectory="$(PublishDir)" Command="neon $(TargetFileName)"/>
```
shargon: I prefer to use the same exe with extra parameter, in order to be able to add this feature in a future

devhawk: I don't think this is a big deal. neon-de is a superset of neon for neo2 and will stay that way as long as neo2 is supported. Once this debug info gets finalized, the code will merged into the main neon for neo2 codebase. We can add a warning to neon-de telling customers to move back to neon, but as long as neon-de generates the same byte code as neon, it doesn't matter. Or we can just continue forward with the neon-de name since that's the only way developers can get the neon compiler without building it themselves from source. 

Since Neo3 is in preview, we should add debug info generation to neon master branch even though debug info isn't finalized. I will do that once I start working on Neo3 support for the [smart contract debugger](https://github.com/neo-project/neo-debugger).

devhawk: sticking with neon-de as a name also means we can reserve "neon" as a name for neo3 (or for a single tool that can support neo2 and neo3 as I suggested in #83)

shargon: neonde is for neo2 or for allow debugging in neo3?

devhawk: neon-de is a fork of neon in the master-2.x branch

shargon: I don't understand why we use different executables, if we want to add `.debug.json` for me we have two ways (both better than a new executable)

- Include this file every time we compile.
- Create this file if you specify the debug argument (`--debug` or something like that)

devhawk: I agree @shargon that we should create the debug.json file every time we compile. The issue is that the format of debug.json hasn't been finalized. I don't think we want to merge the change to emit debug.json into the master-2.x branch of NEON until that format is finalized. Hence the fork

devhawk:
Closing this PR. The dotnet new templates are maintained here: https://github.com/ngdseattle/neo-templates

PR #104, opened 2019-09-05
remove dependecny on neo from neo.compiler.msil

devhawk:
these changes remove the dependency on NEO.dll and NEO.VM.dll from NEON. This is the first step towards having a single NEON tool that targets NEO 2 and 3.

Note, this is PR is not ready to merge. I'm opening it as a draft PR in order to facilitate discussion.


shargon:
But for example we want to test the manifest and also the abi, now we are getting these classes from this library, if we remove the dependencies we will need to replicate a lot of code.

Is not better to use neo2x as a dll, and if you want to compile in 2x, just load this assembly and execute it?

PR #90, opened 2019-08-21
[WIP] NEP5 Template for NEO3

belane:
This is the first draft for the NEP5 template for NEO3 (https://github.com/neo-project/neo/issues/973).
It's a work in progress, still missing notifications, unit tests, ...

PR #106, opened 2019-09-11
Update Neo.SmartContract.Framework according to last neo3 changes

shargon:
This changes pretend update the syscalls according to the last changes for neo3. A lot of changes was done and the compiler was not updated

Require: https://github.com/neo-project/neo/pull/1081 https://github.com/neo-project/neo-vm/issues/215
Required for: https://github.com/neo-project/neo-devpack-dotnet/pull/90

- Added unit tests for all syscalls
- A unit test was included for ensure that this will be updated acording to the last neo nuget


shargon:
Waiting for neo-project/neo#1081

src/Neo.SmartContract.Framework/Services/Neo/Crypto.cs
```diff
@@ -0,0 +1,20 @@
...
+        [Syscall("Neo.Crypto.Hash160")]
+        public extern static byte[] Hash160(byte[] data);
+
+        [Syscall("Neo.Crypto.Hash256")]
+        public extern static byte[] Hash256(byte[] data);
+
+        [Syscall("Neo.Crypto.CheckSig")]
+        public extern static bool VerifySignature(byte[] signature, byte[] pubkey);
+
+        [Syscall("System.Crypto.Verify")]
```
vncoelho: Should we move `Verify` as well?

shargon: I will wait until tomorrow, because the new nuget will come tomorrow. A lot of syscalls must be moved

shargon:
@erikzhang The native contracts expose the `Main` method here

https://github.com/neo-project/neo/blob/4112df22fa5c5080dc0c39469e298db7fa4f85ef/neo/SmartContract/Native/NativeContract.cs#L88-L89

But at the same time, they don't allow direct calls and i try to understand why, because at the beginning i thought that the declaration of this syscall should be like this.

https://github.com/neo-project/neo-devpack-dotnet/pull/106/files#diff-b88b415c97eb9e65791b85a3abc1e887R5-R12

src/Neo.SmartContract.Framework/Services/Neo/Crypto.cs
shargon: Did you mean to `Neo` namespace?

shargon:
Ready to review @neo-project/core

PR #110, opened 2019-09-18
Fix OOM abusing large arrays (3x)

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/112

Fixes:

- `MaxArraySize` was added in order to prevent big arrays that the VM can't handle
- Constructor detection was improved because is bypaseable using `cctor` as class name

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
@@ -6,7 +6,9 @@ namespace Neo.Compiler.MSIL
 {
     class CctorSubVM
     {
-        static Stack<object> calcStack;
+        private const ushort MaxArraySize = ushort.MaxValue;
+        private static Stack<object> calcStack;
+
         public static object Dup(object src)
         {
             if (src.GetType() == typeof(byte[]))
@@ -35,6 +37,7 @@ namespace Neo.Compiler.MSIL
                 return null;
             }
         }
+
         public static byte[] HexString2Bytes(string str)
         {
             byte[] outd = new byte[str.Length / 2];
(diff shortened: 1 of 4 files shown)


src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
```diff
@@ -100,6 +104,7 @@ public static void Parse(ILMethod from, NeoModule to)
                             if ((src.tokenType == "System.Byte") || (src.tokenType == "System.SByte"))
                             {
                                 var count = (int)calcStack.Pop();
+                                if (count > MaxArraySize) throw new ArgumentException("MaxArraySize found");
```
vncoelho: Nice and simple, @shargon, nice catch.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -985,17 +985,12 @@ private bool TryInsertMethod(NeoModule outModule, Mono.Cecil.MethodDefinition me
             try
             {
                 NeoMethod nm = new NeoMethod();
-                if (method.FullName.Contains(".cctor"))
+                if (method.IsConstructor)
```
vncoelho: This is another simplification, right? Is this related to the same issue?

lightszero: IsConstructor is for both cctor and ctor,and we
==“.cctor" is more clear.

CctorSubVM is only for cctor

shargon: I exploited using this bypass, so i think that require to be fixed too

shargon:
@lightszero please take a look again

image

PR #113, opened 2019-10-18
Calculate Coverage

shargon:
Require https://github.com/neo-project/neo/pull/1178


shargon:
@erikzhang We need to configure the environment variable for codecov as we did in the past in other neo projects. Could you do it please?

shargon:
Done, the problem is that SmartContract.Framework could not be instrumented by the coverage engine because it's executed on the VM and the test will fail. Although i fix in the compiler the instrumentation issues, it never be executed because is executed inside the neo VM. 

So the coverage is only for Neo.Compiler.MSIL

shargon:
We should remove OSX here too? or not because is not a library? @erikzhang

erikzhang:
Just remove it.

shargon:
Ok, then it's done

erikzhang:
I think there is no need to upgrade the target frameworks of these projects.

shargon:
Is not better to have all the projects unified?

erikzhang:
When we are going to upgrade the target framework, either because the new framework fixes a serious bug or because we want to use the new features in the new framework. Otherwise, the lower the version of the framework, the better, because there will be better compatibility.

shargon:
Then leave it only for coverage

PR #115, opened 2019-10-19
implement ISNULL

shargon:
Close https://github.com/neo-project/neo-vm/issues/215


shargon:
> made some fix.
"==null"
always had 2 codes in IL
LDNULL
CEQ

But it works as `==null` too?

lightszero:
> > made some fix.
> > "==null"
> > always had 2 codes in IL
> > LDNULL
> > CEQ
> 
> But it works as `==null` too?
```
a==null
a is null
null==a
```
all workd

shargon:
It works perfect @lightszero

shargon:
@lightszero could you approve it?

PR #116, opened 2019-10-20
Support long string switch

lightszero:
fixed https://github.com/neo-project/neo-devpack-dotnet/issues/86

add long string switch support.

find IL code pattern for switch,and delete hashpart,it work for current unittest code.

```c#
        {
            switch (method)
            {
                case "0": return 1;
                case "1": return 2;
                case "2": return 3;
                case "3": return 4;
                case "4": return 5;
                case "5": return 6;
                case "6": return 7;
                case "7": return 8;
                case "8": return 9;
                case "9": return 10;
                case "10": return 11;
                case "11": return 12;
                case "12": return 13;
                case "13": return 14;
                case "14": return 15;
                case "15": return 16;
                case "16": return 17;
                case "17": return 18;
                case "18": return 19;
                case "19": return 20;
                case "20": return 21;

                default: return 99;
            }
        }
```


erikzhang:
I think we should create `ComputeStringHash()` in NeoContract Interop Services first.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -964,6 +957,132 @@ private int _ConvertCall(OpCode src, NeoMethod to)
...
+                        isjumptable = true;
+                        jumptableaddr = method.GetNextCodeAddr(code3.addr);
+                        skipcount += 3;
+                        jumptablecount++;
+                    }
+                    else
+                    {
+                        isjumptable = false;
+                        //順便看看是不是ldstr 段
+                        if (bLdLoc1 && code1.tokenI32 == last.tokenI32 && code2.code == CodeEx.Ldstr && (code3.code == CodeEx.Call || code3.code == CodeEx.Callvirt) && code3.tokenMethod.Contains("String::op_Equality"))
```
shargon: could be replaced `Contains` by a regex? @lightszero

lightszero:
> I think we should create `ComputeStringHash()` in NeoContract Interop Services first.

https://github.com/neo-project/neo-vm/pull/15/files
long long ago

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
lightszero: i am worse about regex

erikzhang:
@lightszero I mean create it in NeoContract Interop Services, not in NeoVM.

shargon:
Maybe create it in the compiler only if is needed is better, because maybe in different frameworks dotnet3, dotnet2, it could change the behavior.

lightszero:
but this pr is skip opcode for " ComputeStringHash" 
 I think for switch we do not need to implentment ComputeStringHash
@erikzhang

shargon:
I agree, if we can detect and remove the `ComputeStringHash` we can do it, because every time after this check they check again the right value

erikzhang:
That's great. Please resolve the conflicts.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
shargon: I will do it

shargon: https://github.com/neo-project/neo-devpack-dotnet/pull/138 @lightszero

lightszero: tks a lot,  megre this first .
I am learning regex too.

lightszero:
why travis always error?
@neo-project/core 
help

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
shargon: It's a PR to your branch

shargon:
@lightszero it was because it's required to execute `dotnet format` first in order to follow all of us the same standards.

shargon:
I think that it require a `rebase`, a lot of files looks as changed, when are not

lightszero:
i had do some change on unittest, for test contract on  debug and release.so got many file changes

shargon:
I don't know why appear files as changed, when are not changed :S

lock9:
@lightszero  @shargon  please check #141 . I rebased this PR and removed some problematic merges.

Issue #123, opened 2019-11-04
NEON should raise a compile error when encountering an event with a return value


devhawk:
As per [NEP-3](https://github.com/neo-project/proposals/blob/master/nep-3.mediawiki#event), in smart contracts do not have return types. NEON should fail to compile contracts that have event delegates with return values

PR #124, opened 2019-11-05
Prevent events with return (3x)

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/123


devhawk:
looks good to me

PR #122, opened 2019-11-04
fix abi bug for 2.x 

lightszero:
fixed https://github.com/neo-project/neo-devpack-dotnet/issues/120

PR #125, opened 2019-11-05
Prevent events with return (2x)

shargon:
Port of https://github.com/neo-project/neo-devpack-dotnet/pull/124 to 2x

Wait for: https://github.com/neo-project/neo-devpack-dotnet/pull/122


Neo.Compiler.MSIL/NeoModule.cs
```diff
@@ -147,7 +149,19 @@ public class NeoEvent
         public string name;
         public string displayName;
         public List<NeoParam> paramtypes = new List<NeoParam>();
-        public string returntype;
```
lock9: I think we should keep the return type for Neo 2, for compatibility purposes.

lapetina9:
## Neo PR Test Report

**Repository:** neo-devpack-dotnet  
**Number:** 125  
**Title:** Prevent events with return (2x)  
**Change description:** Port of #124 to 2x


## Code version / branch

**Neo-devpack-dotnet:**  shargon:close-123-2x


## Test Results

| Test Name | Network | OS | Test Result | Comments |
| ------- |:-------:| --:| -----------:| --------:|
| Initial UT | UT | Localhost | OK | 
| Additional unit test to delegate event with arguments | UT | Localhost | OK | |

## Additional info
The code works but removes a feature unecessarily. We recommend putting back the return type, even if it is always Void.
We merged #122 to test these changes.

## **Test List**

**1. Initial UT**  
Run solution unit tests.

**2. Additional unit test to event with return type**  
  
Create Smart Contract with event return type int.

Expected result:
- While reading Smart Contract, engine throws a NotSupportedException.

shargon:
> Put the return type back in the ABI, even when it is void. I can't see gains for this change in Neo 2.

This pr is precisely for this hahha

lock9:
Hi @shargon , but why are we removing the property? Isn't the purpose of this PR block events with return type *different* than void? 
I think that removing a property is a breaking change, I thought that the idea of this PR was to prevent invalid events, not to remove them.
What is your opinion? Do you think we really should remove it or just allow void return types?

shargon:
> Isn't the purpose of this PR block events with return type different than void?

The purpose is throw an error when one event return different than void, because is not in NEP3

lock9:
Hi @shargon,  can't we do that without changing the JSON file? We should keep the exception.
I'm going to remove my request for changes. Since #122 is not merged yet, there will be no difference.

shargon:
Why? it's a new compilation, why you need to compile a wrong SC?

Neo.Compiler.MSIL/NeoModule.cs
shargon: Is not a breaking change, in the case that you want to recompile your SC, you need to fit to the new rules, but your SC still working on the chain.

shargon:
Travis still broken 😔

erikzhang:
Please re-create the PR from a new branch in neo-project. Don't create PRs from personal branches. Then we can restart travis easily.

cloud8little:
> Travis still broken 😔

There is no .travis.yml file for branch: master-2.x.

shargon:
> Don't create PRs from personal branches.

The problem is that i can't use gitkraken with the neo-project branchs, i have no rights with gitkraken :(

shargon:
Closed in favor of https://github.com/neo-project/neo-devpack-dotnet/pull/136

PR #121, opened 2019-11-04
add unittest framework for neo-devpack-dotnet branch 2.x.

lightszero:
Add unittest framework for master-2.x.
fix bug will be more simple.

PR #127, opened 2019-11-05
DevPack dependencies update + general fixes (2x)

### LINKED ISSUE #126: Neo 2 devpack not passing tests
**How to reproduce:**
- Clone neo-devpack-dotnet
- Checkout master-2.x
- Open the solution
- Run MSIL Unit tests

Result:
image

I managed to fix some tests (the problem was the dll name was misconfigured), but now I have a much bigger concern: the code is actually broken.
This works:

```Csharp
public static string UnitTest_003()
{
       return "Unit Test";
}
```
But this doesn't:

```Csharp
public static byte UnitTest_001()
{
     var nb = new byte[] { 1, 2, 3, 4 };
     return nb[2];
}
```

@devhawk

lock9:
Hi, this PR includes:

- Fixing copy scripts for test contract
- Updating dependencies to use latest .Net Framework, .Net Standard and .Net Core
- 'Fixing'  compiler tests and adding a new one
- Removing dead code
- Configuring project to use only one framework (.Net Standard or .Net Core or .Net Framework)
- Migrating TestContract to Neo.TestContract and making usage of .Net Standard instead of .Net Framework

For some weird reasons, it fixes all UT (closes #126)


Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj
```diff
@@ -5,7 +5,7 @@
     <AssemblyTitle>Neo.SmartContract.Framework</AssemblyTitle>
     <Version>2.9.3.1</Version>
     <Authors>The Neo Project</Authors>
-    <TargetFrameworks>netstandard1.6;net40</TargetFrameworks>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
devhawk: unless we are using features from netstandard 2.1 (span/memory in particular) we should leave this on netstandard 2.0

lock9: @devhawk  shouldn't we always use the latest version? At least in mobile development, this is a rule, I don't know about MS stuff

lock9: But I have to confess that I don't understand how MS name their stuff. Isn't 2.1 stable?

Neo.Compiler.MSIL.App/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Neo.Compiler.MSIL.App
+{
+    class Program
```
devhawk: What is the point of creating a seperate project for the app?

Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
```diff
@@ -5,8 +5,8 @@
     <AssemblyTitle>Neo.Compiler.MSIL</AssemblyTitle>
     <Version>2.4.1.1</Version>
     <Authors>The Neo Project</Authors>
-    <TargetFrameworks>netcoreapp2.0;netstandard2.0</TargetFrameworks>
-    <OutputType>Exe</OutputType>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
devhawk: unless we are using features from netstandard 2.1 (span/memory in particular) we should leave this on netstandard 2.0

Neo.ConvertTask/Neo.ConvertTask.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Neo</RootNamespace>
     <AssemblyName>Neo.ConvertTask</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
```
devhawk: When can we kill this project entirely? There's not really a good need for it. the exec task should do just fine

Neo.Compiler.MSIL.App/Program.cs
lock9: When we use 2 TargetFrameworks, VS doesn't work properly. Also, the compiler is a library, but the other project was configured to be a console app. 
Putting it in a different project, it is now .Net standard only (VS friendly) and it is now a library, not a console app

Neo.ConvertTask/Neo.ConvertTask.csproj
devhawk: Example of using exec in a PR I'm still waiting for @erikzhang to approve: https://github.com/neo-project/neo-devpack-dotnet/blob/3826fc89b4324817a2556d0aed636c2c67cae581/templates/dotnet-templating/templates/csharp/Directory.Build.targets

lock9: @devhawk Once your debugger is released I think we can. This is used in the project templates.

Neo.Compiler.MSIL.App/Program.cs
lock9: I have discussed with @devhawk  and we agreed that this is a console application and not a library.

Neo.ConvertTask/Neo.ConvertTask.csproj
erikzhang: It has been kill in master branch.

Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj
```diff
@@ -5,28 +5,21 @@
     <AssemblyTitle>Neo.SmartContract.Framework</AssemblyTitle>
     <Version>2.9.3.1</Version>
     <Authors>The Neo Project</Authors>
-    <TargetFrameworks>netstandard1.6;net40</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
erikzhang: Why upgrade the target frameworks?

Neo.ConvertTask/Neo.ConvertTask.csproj
lock9: @erikzhang  do we need to keep this in this branch? Are we replacing the VS Extension to a VS Code extension? Because I believe that NGD Seattle code is Neo 2 ready

Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj
lock9: It seems that using 2.0 is the most reasonable choice, according to @devhawk.
I also read the docs and they recommend to target 2.0 too (we could target 1.6 and 2.0 but I see no reasons for that).
https://docs.microsoft.com/en-us/dotnet/standard/net-standard#which-net-standard-version-to-target

lock9: @erikzhang  I can move it back to 1.6 if you prefer. Do you want me to put 1.6 back?

erikzhang: https://github.com/neo-project/neo-devpack-dotnet/pull/113#issuecomment-545783446

lock9:
Closing in favor of #121

PR #156, opened 2019-12-10
Fix abi method name

erikzhang:


src/Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
```diff
@@ -19,6 +19,7 @@
     <Product>Neo.Compiler.MSIL</Product>
     <Description>Neo.Compiler.MSIL</Description>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <LangVersion>latest</LangVersion>
```
erikzhang: We don't need `LangVersion` in `netcoreapp3.1` or `netstandard2.1`.

shargon: I was not able to compile without this line

image

erikzhang: Are you using the latest version of visual studio?

erikzhang: My version is 16.4.0

shargon: 16.3.10, I will update it now

shargon: With the last version it's not needed, if you want we can remove the line

vncoelho:
Ok here as well 

image

image

dotnet-sdk-3.1

Issue #158, opened 2019-12-10
Unify master-2.x contract metadata attributes with master branch


devhawk:
Master branch has added `FeatureAttribute` while master-2.x has ContractHasStorage/ContractHasDynamicInvoke/ContractIsPayable. Need to change master-2.x to match master branch

devhawk:
@shargon I'd appreciate your thoughts on this issue. I started to do this, but I ran into a few incompatibilities:

* FeaturesAttribute is applied to a class while all of the contract attributes I added are applied to the assembly. As far as I understand, NEON doesn't support multiple contracts in a single managed dll - at least for 2.x. If that's the same for NEON 3.x, should FeaturesAttribute be applied to the assembly not the class? Or should all the 2.x attributes apply to class instead of assembly. 
  * My preference is to attach these attributes to assembly not class. What does it mean to attach these attributes to classes *other* than the main contract class?
* FeaturesAttribute uses the ContractFeatures enum, but the 2.x branch already has the ContractPropertyState enum. I don't think we should define to enums with the same values and we can't change the existing ContractPropertyState enum without breaking code. I'm thinking the 2.x branch version of FeaturesAttribute should just take a ContractPropertyState value in its constructor, even though it's not 100% compatible with 3.x branch.

shargon:
> should FeaturesAttribute be applied to the assembly not the class?

Yes, I think so.

> even though it's not 100% compatible with 3.x branch.

Don't worry about that, it's ok for me

shargon:
Seems weird to me with smart contracts with one file.

```csharp
using Neo.SmartContract.Framework;
using Neo.SmartContract.Framework.Services.Neo;

[assembly: Features(ContractFeatures.HasStorage)]

namespace Neo.Compiler.MSIL.TestClasses
{
    class Contract_Storage : SmartContract.Framework.SmartContract
    {
```

shargon:
Please @devhawk , Take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/161

devhawk:
Not all contracts are a single file: https://github.com/Moonlight-io/moonlight-ico-template
[...]

PR #161, opened 2019-12-10
Merge assembly attributes with the SmartContract Attributes

shargon:
Related to https://github.com/neo-project/neo-devpack-dotnet/issues/158

- Also was moved the `FeatureAttribute` to `AssemblyAttribute`

### DIFF
--- a/src/Neo.SmartContract.Framework/FeaturesAttribute.cs
+++ b/src/Neo.SmartContract.Framework/FeaturesAttribute.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace Neo.SmartContract.Framework
 {
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false)]
     public class FeaturesAttribute : Attribute
     {
         /// <summary>
--- a/src/Neo.Compiler.MSIL/MSIL/Converter.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Converter.cs
@@ -70,7 +70,7 @@ namespace Neo.Compiler.MSIL
         {
             this.inModule = _in;
             //logger.Log("beginConvert.");
-            this.outModule = new NeoModule(this.logger)
+            this.outModule = new NeoModule(this.logger, _in)
             {
                 option = option ?? ConvOption.Default
             };
--- a/src/Neo.Compiler.MSIL/MSIL/ILModule.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/ILModule.cs
@@ -12,7 +12,9 @@ namespace Neo.Compiler.MSIL
         public Mono.Cecil.ModuleDefinition module = null;
         public List<string> moduleref = new List<string>();
         public Dictionary<string, ILType> mapType = new Dictionary<string, ILType>();
+        public List<Mono.Cecil.CustomAttribute> attributes = new List<Mono.Cecil.CustomAttribute>();
         public ILogger logger;
+
         public ILModule(ILogger _logger = null)
         {
             this.logger = _logger;
@@ -20,6 +22,12 @@ namespace Neo.Compiler.MSIL
         public void LoadModule(System.IO.Stream dllStream, System.IO.Stream pdbStream)
         {
             this.module = Mono.Cecil.ModuleDefinition.ReadModule(dllStream);
+
+            if (module.Assembly.HasCustomAttributes)
+            {
+                attributes.AddRange(module.Assembly.CustomAttributes);
+            }
+
             //#if WITHPDB
             if (pdbStream != null)
             {
@@ -52,7 +60,6 @@ namespace Neo.Compiler.MSIL
                         foreach (var nt in t.NestedTypes)
                         {
                             mapType[nt.FullName] = new ILType(this, nt, logger);
-
                         }
                     }
                 }
--- a/src/Neo.Compiler.MSIL/NeoModule.cs
+++ b/src/Neo.Compiler.MSIL/NeoModule.cs
@@ -9,8 +9,6 @@ namespace Neo.Compiler
 {
     public class NeoModule
     {
-        public NeoModule(ILogger logger) { }
-
         public string mainMethod;
         public ConvOption option;
         public List<CustomAttribute> attributes = new List<CustomAttribute>();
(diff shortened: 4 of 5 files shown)


devhawk:
Note, I have a branch ready to PR for master-2.x branch that will mirror FeatureAttribute from 3.x branch: https://github.com/devhawk/neo-devpack-dotnet/tree/devhawk/FeaturesAttribute

erikzhang:
Applying attributes on assembly is not a good idea if we allow multiple contracts in a single assembly in the future.

devhawk:
> Applying attributes on assembly is not a good idea if we allow multiple contracts in a single assembly in the future.

It's not clear to me what developer scenario would be enabled by supporting multiple contracts in a single assembly. And if one did become clear, we could change the attributes to support applying these attributes to classes without breaking existing code.

erikzhang:
I think applying the attribute to the class is a good choice. Many DApps now deploy multiple contracts. NEO3's manifest file now supports contract groups, allowing multiple contracts to be in the same group and trusting each other. So in the future we will hope to allow the development of these related contract groups in one assembly. Therefore, an assembly may contain multiple contracts with different functions.

devhawk:
Is there some developer value to co-locating the contracts in a single assembly? For example, to make it easier to code a contract that dynamically invokes another contract? 

Today, it is possible to have multiple classes in a smart contract project (example: [Moonlight ICO template](https://github.com/Moonlight-io/moonlight-ico-template)). In this scenario, how would the compiler handle contact metadata attributes attached to these other classes? What if these attributes supply conflicting values? This seems likely to confuse developers in the short term without providing any benefit until the compiler is updated to handle multiple contracts in a single assembly - which may or may not happen.

Since we can update these attributes to support class attachment in the future without breaking backwards compatibility, it would seem wise to implement the contracts for the way the compiler works today and address updating the smart contract framework if and when we decide to support the multiple contract / single assembly scenario.

Note, regardless of what is decided for the master branch, I plan to submit a PR to add an assembly attached FeaturesAttribute to master-2.x branch. Given that we are limiting updates to master-2.x branch, it seems unlikely that we will add multi-contact support to neon for neo 2

erikzhang:
The contract class is inherited from the base class `SmartContract`. So the compiler is easy to handle the attributes on the contract class.

devhawk:
> The contract class is inherited from the base class `SmartContract`. So the compiler can easily handle the attributes on the contract class.

Does the compiler enforce only a single class in a smart contract project can inherit from SmartContract? In the Moonlight ICO template, every class in the project except for StorageKeys inherits from SmartContract.

erikzhang:
If there are multiple non-abstract subclasses inheriting from `SmartContract`, the compiler should generate a NEF file for each subclass. But I don't think it currently supports it. Currently it only handle the first contract class found.

devhawk:
@erikzhang if you think that supporting multiple contracts in a single assembly is an important scenario, I would recommend that you open a new issue to track that change. I would also suggest that we merge this PR making FeaturesAttribute assembly attached *for now* with the expectation that it would change FeaturesAttribute to be class attached if and when NEON gets updated to support multiple contracts. That way, the developer experience is consistent end to end, regardless if NEON is updated to support multiple contracts or not.

src/Neo.SmartContract.Framework/FeaturesAttribute.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Neo.SmartContract.Framework
 {
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false)]
```
erikzhang: This attribute was originally applied to classes. Now that we know that we may allow multiple contract classes in a single assembly in the future, why do we change it to apply it to assemblies? This is a very strange to me.

devhawk: Since it is unclear if and when NEON will be updated to support multiple contract classes in a single assembly, this attribute should target assemblies instead of classes in order to have a consistent developer experience. 

If NEON is updated to support multiple contract classes in a single assembly in the future, then this attribute should be updated at that time.

shargon:
i think that if in the future we will have multiple contracts in one assembly, os better like now, per class

devhawk:
> i think that if in the future we will have multiple contracts in one assembly, os better like now, per class

As I said to @erikzhang, I think it would be better for the behavior of the compiler + the attributes to be consistent. Since the compiler today doesn't support multiple contracts in one assembly, the contract metadata attributes should apply to the assembly. If in the future we change the compiler, we can also change the attributes. 

Looking at the NEON master branch code, I notice two things:

1. While the attributes are attached to classes in the .net metadata, NEON [reads them from the NeoModule instance](https://github.com/neo-project/neo-devpack-dotnet/blob/63e90e567b50c41d1f4e1d621830ebc4020cbf9f/src/Neo.Compiler.MSIL/Program.cs#L237). 
2. The NeoModule attributes field is populated by collecting all the custom attributes from the first class in the project that inherits from `SmartContract`. In a project like Moonlight, every class inherits from `SmartContract` so the class NEON picks to collect attributes from may not match the class with the contract entry point or the class the developer has attached the `FeaturesAttribute` to.

This behavior may lead to unexpected behavior that a developer not familiar with the internals of the NEON compiler would not be able to diagnose. Making `FeaturesAttribute` assembly attached is the easiest way in the short term to make the developer experience consistent. Making `FeaturesAttribute` class attached can be done in the future if the work is done to make NEON support multiple contracts in a single assembly.

Frankly, if we want to leave this in an inconsistent state for now, that's fine by me. Someone from the core team should file an issue to track the work to make NEON support multiple contracts in a single assembly.

shargon:
But today is not the same as tomorrow, you propose to change it again in a future. 

I think that it'ss easier to detect if we have more than one SmartContract in the assembly and throw an error, instead of changing the smart contract tomorrow again, because it will not work as before, because the Attribute will change from the `Assembly` to the` Class' again. So we will need to tell the developers that now he need to change the attribute to the class.

So, if we have plans to allow more than one in the future, I think it is reasonable to keep it as before.

shargon:
We currently prevent this

https://github.com/neo-project/neo-devpack-dotnet/blob/d92e60a37c78ac7c42a13b00ab3e0380fa1626b3/src/Neo.Compiler.MSIL/MSIL/Converter.cs#L199

devhawk:
That code only checks to make sure only one smart contract class has a `Main` method. it is still possible  to have multiple classes that inherit from SmartContract so long as only one has a main method. 

Can we update #164 to ensure the contract attributes are retrieved from the class that contains the main method entrypoint?

shargon:
Closed in favor of #164

PR #164, opened 2019-12-11
Test multiple entry points in the same assembly

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/pull/161 https://github.com/neo-project/neo-devpack-dotnet/issues/158

- Prevent mistakes with multiples SmartContracts in one assembly 
- Deterministic behaviour with the `FeatureAttribute`


src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -196,7 +196,7 @@ public NeoModule Convert(ILModule _in, ConvOption option = null)
                             if (l.Value == m)
                             {
                                 if (mainmethod != "")
-                                    throw new Exception("Have too mush EntryPoint,Check it.");
+                                    throw new Exception("The smart contract contains multiple entryPoints, please check it.");
```
devhawk: Do we check somewhere else to make sure there is at least one entry point? If the developer writes a contract with zero Main entrypoints, will the compiler raise that as a compile error?

shargon:
> That code only checks to make sure only one smart contract class has a Main method. it is still possible to have multiple classes that inherit from SmartContract so long as only one has a main method.
> Can we update #164 to ensure the contract attributes are retrieved from the class that contains the main method entrypoint?

Done

devhawk:
> > That code only checks to make sure only one smart contract class has a Main method. it is still possible to have multiple classes that inherit from SmartContract so long as only one has a main method.
> > Can we update #164 to ensure the contract attributes are retrieved from the class that contains the main method entrypoint?
> 
> Done

I don't see where this is done. ModuleConverter.Convert is [still populating outModule.attributes](https://github.com/neo-project/neo-devpack-dotnet/blob/6989af1d3517035364c36993033b9e9925f09b50/src/Neo.Compiler.MSIL/MSIL/Converter.cs#L226) based on the first smart contract class it finds, rather than from the smart contract class containing the entrypoint. 

Out of curiosity, if there are multiple classes that inherit from smart contract and none of them have an entrypoint, which will be used for the auto insert entrypoint?

shargon:
> I don't see where this is done. ModuleConverter.Convert is still populating outModule.attributes based on the first smart contract class it finds, rather than from the smart contract class containing the entrypoint.

Because if there are more than one, it will throw an exception. So at now, is not possible to have more than one

devhawk:
> Because if there are more than one, it will throw an exception. So at now, is not possible to have more than one
My read of the code is that you throw an exception if you find more than one `Main` method, independent of what parent class they have. But then the outModule.attributes is populated by this bit of code:

``` csharp
var attr = outModule.mapMethods.Values
    .Where(u => u.inSmartContract)
    .Select(u => u.type.attributes.ToArray())
    .FirstOrDefault();
if (attr?.Length > 0)
{
    outModule.attributes.AddRange(attr);
}
```

If there are more than one class in the assembly that inherits from SmartContract, this code will pick whichever one is first in the mapMethods.Values collection and use its containing type's attributes. I don't think that is deterministic.

shargon:
> I don't think that is deterministic.

It is because you can't build it if there are two smart contracts, This issue that you say is not related to this PR and should be done when we accept more than one SmartContract in the same assembly.

devhawk:
> It is because you can't build it if there are two smart contracts, This issue that you say is not related to this PR and should be done when we accept more than one SmartContract in the same assembly.

Please add a test with two classes that inherit from `SmartContract` but only one that has a Main entrypoint function.

shargon:
Both will have entrypoints because if not it will be auto-generated, in which do you want the attribute?

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_EntryPoints.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Neo.Compiler.MSIL
+{
+    [TestClass]
+    public class UnitTest_EntryPoints
+    {
+        [TestMethod]
+        public void Test_MultipleContracts()
+        {
+            var testengine = new TestEngine();
+            Assert.AreEqual(2, Assert.ThrowsException<EntryPointException>(() => testengine.AddEntryScript("./TestClasses/Contract_multilecontracts.cs")).Count);
```
shargon: @devhawk is not this UT?

erikzhang:
@shargon

shargon:
> what problem had be fixed

Only was added an specific exception if there are more than one entry points or none. And some Unit test for test different scenarios, with two sc, without sc, with attributes, without...

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_EntryPoints.cs
devhawk: No, I'm asking about a test case that looks like this:

``` csharp
namespace Neo.Compiler.MSIL.TestClasses
{
    class Contract_a : SmartContract.Framework.SmartContract
    {
        {
            return 'a';
        }
    }

    class Contract_b : SmartContract.Framework.SmartContract
    {
        public static object AMethodNotNamedMain(string method, object[] args)
        {
            return 'b';
        }
    }
}
```

shargon: Yes, It will take the defined one `Contract_a`. It will not fail.

shargon: Please @devhawk , Check it again

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_EntryPoints.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    [TestClass]
+    public class UnitTest_EntryPoints
+    {
+        [TestMethod]
+        public void Test_MultipleContracts()
+        {
+            using (var testengine = new TestEngine())
+            {
+                Assert.AreEqual(2, Assert.ThrowsException<EntryPointException>(() => testengine.AddEntryScript("./TestClasses/Contract_multilecontracts.cs")).Count);
```
devhawk: nitpick: multiple is misspelled in this file name (and again below)

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_EntryPoints.cs
```diff
@@ -0,0 +1,30 @@
...
+            }
+
+            using (var testengine = new TestEngine())
+            {
+                Assert.AreEqual(2, Assert.ThrowsException<EntryPointException>(() => testengine.AddEntryScript("./TestClasses/Contract_multilecontracts2.cs")).Count);
```
devhawk: To clarify, having multiple classes in a project that inherit from `SmartContract` but where only one has a Main method is allowed today. I'm OK with changing this - behavior in smart contracts is mostly implemented in static methods so inheriting from SmartContract is primarily a marker anyway. But we should be explicit that we are changing this. If we have a doc listing all of the 2 -> 3 changes, this should be in it.

shargon: Now, only one SmartContract is allowed in one assembly

shargon:
Summary of changes:

- Prevent mistakes with multiples SmartContracts in one assembly (not possible now)
- Deterministic behaviour with the `FeatureAttribute`

src/Neo.Compiler.MSIL/EntryPointException.cs
```diff
@@ -0,0 +1,22 @@
...
+        public int Count { get; }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="count">Number of entry point founds</param>
+        /// <param name="message">Message</param>
+        public EntryPointException(int count, string message) : base(message)
+        {
+            Count = count;
```
lock9: HI @shargon, if possible, I would prefer to have the URL of the main methods than only having the Count. This way we can present a better error message.

erikzhang:
We have supported auto-generated entry points long ago.

lock9:
@erikzhang  I know, but it is not mandatory. In my opinion, we should _block_ user-generated entry-points.

erikzhang:
Some contracts have no method other than the entry point.

lock9:
I think that is ok too. You can have only one method. The difference is that you are not calling it Main, but use a more domain-specific name. 
Just to be clear: I think we shuold auto-generate the entry-point function to ensure that the ABI is respected.  That is my main argument, I'm looking for a solution to that.
Today the ABI never matches the SmartContract methods, you know the method names but you don't know how to access them.

shargon:
> we should block user-generated entry-points.

Why? what are the benefits?

lock9:
@shargon  ensure the ABI is being respected. 
I found out that problem when I was using my own version of neo-cli. 

I thought that by having the ABI I could easily improve the user interaction with the contract, since we have the method names and parameter types. However, I found out that the ABI is not respected. And this is by default.

On code, you implement `Transfer`, on the ABI, it will be ´Transfer´, but at code level, the switch-case will use `transfer`, or `safeTransfer`. In any case, the name, never matches, only if the user disrespect C# naming conventions.

shargon:
I think that this is error could be made because unknown the technology, for me is not an error that deserve to be fixed avoiding the main method. But it's my opinion

If a developer made something wrong with his design, is not our fault, he can do a lot of things wrong, like alter the ABI manually.

lock9:
@shargon  I'm not against this PR. I think we should block if there are two, but in my opinion, in the future, we could block even if it has only one (because it would be auto-generated). About the developer, the situation is the opposite: the interface is not for the current SC developer, but to other developers and systems. 

From wikipedia:
> In computing, an interface is a shared boundary across which two or more separate components of a computer system exchange information.

However, this is not true for us.  You cannot use the ABI to define how our messages should be written. It feels weird for me (and caused me problems) to have an interface for a contract that doesn't actually work. 

If there is an agreement that the ABI doesn't have to be respected, I can delete this from my mind. I just want to make sure we are on the same page.

Issue #143, opened 2019-12-02
Need to remove conditional compilation


erikzhang:
https://github.com/neo-project/neo-devpack-dotnet/blob/63e90e567b50c41d1f4e1d621830ebc4020cbf9f/src/Neo.Compiler.MSIL/MSIL/Converter.cs#L46

https://github.com/neo-project/neo-devpack-dotnet/blob/63e90e567b50c41d1f4e1d621830ebc4020cbf9f/src/Neo.Compiler.MSIL/MSIL/Converter.cs#L1145

https://github.com/neo-project/neo-devpack-dotnet/blob/63e90e567b50c41d1f4e1d621830ebc4020cbf9f/src/Neo.Compiler.MSIL/MSIL/ILModule.cs#L23

We should remove these conditional compilations.

lock9:
@erikzhang  do you think we should remove  #if NET47 from all projects? Is it done already? I don't remember if we removed .net framework 4.7 completely or not.

erikzhang:
I'm sure we have no NET47 any more. The last .NET Framework project is neo-gui, which has been ported to .NET Core 3.0.

devhawk:
Even if neo-gui was still on .NET Framework, that shouldn't block NEON from targeting .NET core only

PR #165, opened 2019-12-11
Remove compiler directives

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/143

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Converter.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Converter.cs
@@ -43,23 +43,6 @@ namespace Neo.Compiler.MSIL
                 logger = new DefLogger();
             }
             this.logger = logger;
-#if NET47
-            try
-            {
-                var assm = System.Reflection.Assembly.GetAssembly(typeof(System.Action));
-                var name = System.IO.Path.GetFileName(assm.Location);
-                if (name.ToLower() == "mscorlib.dll")
-                {
-                    var path = System.IO.Path.GetFullPath(".");
-                    System.IO.File.Copy(assm.Location, System.IO.Path.Combine(path, name));
-                }
-            }
-            catch
-            {
-
-            }
-            //assm.Location
-#endif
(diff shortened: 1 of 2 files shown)


src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -1152,12 +1135,8 @@ private int ConvertCode(ILMethod method, OpCode src, NeoMethod to)
                     break;
 
                 default:
-#if WITHPDB
                     logger.Log("unsupported instruction " + src.code + "\r\n   in: " + to.name + "\r\n");
-                    break;
-#else
                     throw new Exception("unsupported instruction " + src.code + "\r\n   in: " + to.name + "\r\n");
```
erikzhang: This line should be removed too?

shargon: Maybe the log, but not the exception, isn't it? @lightszero

shargon: @lightszero we should throw the exception or only log it?

shargon: For me, it have sense to stop the conversion if one instruction is unsupported.

lightszero: I think should throw for here,skip this could be got more error..

cloud8little:
smoke test passed.

neo-devpack-dotnet: master dda641a4030c0562937391e68336a16b3f7c5950
neo-cli: master 51cd29fbe21abb9e1f17f64e5c6d21bc7decbbb9
neo: master ab4830c
neo-vm: master be2ac36bf35a3033d828e0ba0630d390599c487d
1. build project neo-devpack-dotnet. [Pass]
2. build helloworld smart contract.  [Pass]
3. deploy helloworld contract. [Pass]
4. deploy nep5 contract. [Pass]

Issue #163, opened 2019-12-11
Update master 2.x version numbers for NEON and Smart Contract Framework


devhawk:
We need to ship a new version of NEON (packaged as a global tool as per #159) and the smart contract framework. As part of that release, these packages need new version numbers. Since these tools are in maintenance mode, I believe we should continue to update the version numbers manually. 

- current version of Neo.SmartContract.Framework is v2.9.3.1. I recommend moving to v2.10
  - as per [Semantic Versioning](https://semver.org/) we should bump the minor version number because we have added backward compatible changes

- current version of Neo.Compiler.MSIL is v2.4.1.1. I recommend moving to v2.6.
  - Again, as per [Semantic Versioning](https://semver.org/) we should bump the minor version number because we have added backward compatible changes
  - The current version of [Neo.neon-de](https://www.nuget.org/packages/Neo.neon-de) is v2.5.11. Even though neon-de is a separate fork, it seems reasonable to bump neon an extra minor version number to make it clear that neon is the latest and greatest version of the tool to use

Issue #82, opened 2019-08-09
Distribute neon as a .NET Global Tool


devhawk:
the current model of distributing the NEON compiler requires many steps for the developer. .NET Core supports [global tools](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) which are easy to install and automatically are added to the path. Basically, all the developer would have to do would be 

``` shell
$ dotnet tool install -g Neo.Compiler.MSIL
```

we might also want to change the package name to just "neon"

erikzhang:
It's a good idea!

shargon:
i like it

devhawk:
I'm going to make a series of changes to enable NEON to target NEO 2 and 3 + the generate debug info for neo-debugger. I will do enable NEON to be a global tool as part of that work

devhawk:
FYI, I am going to do this w/o the changes to enable targeting Neo 2 and 3 from a single NEON exe

erikzhang:
I have tried to publish with `/p:PublishSingleFile=true` and `/p:PublishTrimmed=true`, but it still generates huge exe.

devhawk:
I just built [my soon-to-pr fork](https://github.com/devhawk/neo-devpack-dotnet/tree/devhawk/neonGlobalTool) locally and the resulting nupkg is about 300kb.

PR #166, opened 2019-12-11
Package NEON as .net global tool

devhawk:
Fixes #164, #159, #82 , #163 

NEON was updated to .NET core 3.1 as part of #154. This PR adds additional changes in csproj files to enable building NEON as a [.NET Core Global Tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools). This will allow developers to install NEON from the command line instead of having to clone this repo and build manually

``` shell
$ dotnet tool install -g Neo.Neon
```

Currently, the Neo.Compiler.MSIL builds as both .NET Standard 2.0 and .NET Core 3.1 to support scenarios where someone wants to use the compiler as a library. Unfortunately, you can't build as a global tool when targeting .net standard. So I refactored the build script to allow building as .net standard and .net core global tool separately. `dotnet pack` will build Neo.Compiler.MSIL as a .NET Standard 2.0 library package while `dotnet pack /p:GlobalTool=true` builds Neo.Compiler.MSIL as a  .NET Core 3.1 global tool. To differentiate the packages, I gave the global tool package the name `Neo.Neon` instead of `Neo.Compiler.MSIL`.

Additionally, this PR bumps the version numbers. 

* The current version of Neo.SmartContract.Framework is v2.9.3.1. This PR bumps the version to v2.10 as per [Semantic Versioning](https://semver.org/) guidelines
* The current version of Neo.Compiler.MSIL is v2.4.1.1. This PR bumps the version to v2.6. [Semantic Versioning](https://semver.org/) guidelines would suggest a new version number of v2.5, but [the temporary Neo.neon-de fork](https://www.nuget.org/packages/Neo.neon-de) already shipped v2.5.x releases. Bumping Neo.Neon to v2.6 makes it clear that this is the more up-to-date package.


Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
```diff
@@ -21,6 +19,23 @@
...
+
+  <PropertyGroup Condition="'$(GlobalTool)' != true">
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <PackageId>Neo.Compiler.MSIL</PackageId>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(GlobalTool)' == true">
+    <PackAsTool>true</PackAsTool>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <PackageId>Neo.Neon</PackageId>
```
erikzhang: `neon`?

superboyiii: I think it makes sense as the name of. netcoreapp since it's always neon.exe.

devhawk: Not sure what you're asking @erikzhang. The exe is named neon, but the package is named neo.neon. All neo packages need to have the "neo" prefix in order to get associated with the [neo-project nuget org](https://www.nuget.org/profiles/neo-project)

erikzhang: If the `PackageId` doesn't match the exe name, we have to install it by a name and use it by another name. It is too confusing.

devhawk: Neo express already works this way. Exe name is `neo-express` while package name is `neo.express`. Besides, developers aren't typically going to execute neon.exe manually, it is automatically invoked by the contract build script created by the [template](https://github.com/ngdseattle/neo-templates).

Additionally, there are already a bunch of [neon packages](https://www.nuget.org/packages?q=neon) on nuget.org that we don't want to get confused with

devhawk: Many [existing dotnet tools](https://github.com/natemcmaster/dotnet-tools) have exe names that don't match their package name.

superboyiii:
@devhawk Nice feature! Harry, you missed `neo-logo-72.png` at the same path of neo-devpack-dotnet.sln. And I just use a png of the same name to test, it works on both .NET Standard 2.0 and .NET Core 3.1. But some issue appear when installing, not sure how to solve.
image
image

Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
erikzhang: What is the benefit of a PackageId and exe name mismatch?

superboyiii: I think it's used to avoid conflict of the same name for packages on nuget.

erikzhang: I searched on nuget, there is no conflict name.

devhawk:
Thanks for testing this @superboyiii! I'm not sure what you mean by 

> you missed neo-logo-72.png at the same path of neo-devpack-dotnet.sln.

that file was added in commit e7d56b8802161f5b79e908935289cdd1b9619be5. I added it at the root so it could be accessed for packages built out of ./Neo.Compiler.MSIL and ./Neo.SmartContract.Framework

devhawk:
@superboyiii I'm sorry, I provided instructions for building the neo.neon package but I forgot to include instructions for installing it!

I see you've successfully built the tool package using /p:GlobalTool=true. In order to install the tool, you need to provide the package location via --add-source. Otherwise, dotnet tool install looks in nuget.org for the tool package. Eventually that will work, but the package has not yet been published on nuget.org yet. You should be able to install the package with this command:

``` shell
> dotnet tool install neo.neon -g --add-source .\bin\Debug\
```

superboyiii:
That's cool! Thanks for your clear instruction.

shargon:
I will test it tomorrow!

Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
shargon: Can we call it Neo2.XXX ?

erikzhang: I prefer `neon` with version `2.x`.

shargon:
@devhawk I was not able to uninstall it, what i am doing wrong?
image

devhawk:
@shargon you misspelled uninstall

devhawk:
> i think that we should rename it to something related to neo2.
> `neo2.neon` or something like that.

I think there should be one version of NEON that works with both neo 2 & 3. I filed #83 to this effect a while ago.

PR #167, opened 2019-12-12
Add UnitTest for Array and fixed int Array Init  problem

lightszero:


tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_Array.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var arrobj = new int[3];
+            arrobj[1] = 88;
+            arrobj[2] = 33;
+            return arrobj[2];
+        }
+        public static object TestIntArrayInit()
+        {
+            var arrobj = new int[] { 0, 88, 33 };
+            arrobj[1] = 88;
```
shargon: Other value?

shargon:
@lightszero could you check my last UT, is not working

https://github.com/neo-project/neo-devpack-dotnet/pull/167/commits/e85eb3d2958a54ab761ea60be9e03848105a7a55

lightszero:
> @lightszero could you check my last UT?, is not working
> 
> [e85eb3d](https://github.com/neo-project/neo-devpack-dotnet/commit/e85eb3d2958a54ab761ea60be9e03848105a7a55)

fun，i will check it.

shargon:
@lightszero please take a look to the test `Test_DefaultArray`

lightszero:
> @lightszero please take a look to the test `Test_DefaultArray`, it take null instead of 0

in IL new array was filled by default value，int default value is 0
in nef new array was filled by null

shargon:
> in IL new array was filled by default value，int default value is 0
in nef new array was filled by null

I know but any idea about how to solve this?

@erikzhang `ARRAYFILL` opcode ?

erikzhang:
What about `NEWARRAY_T`? It creates a new array and fills the array with default values of the specified type.

shargon:
> What about NEWARRAY_T? It creates a new array and fills the array with default values of the specified type

Good idea, with a list of types?  i think that it's enough with `bool`, `integer` or `null`

erikzhang:
https://github.com/neo-project/neo-vm/issues/241

PR #169, opened 2019-12-12
add unittest for NullCoalescing and nullCollation

lightszero:
add unittest for NullCoalescing and nullCollation

example:
return code ?? "linux"
return  code ?.Substring(1,2);


shargon:
Please, Could you add new lines between the methods?

lightszero:
> Please, Could you add new lines between the methods?

ok ，I will format it

PR #170, opened 2019-12-12
support code in cctor

lightszero:
before only can init const value in cctor step.
``` c#
     static byte[] callscript = HexToBytes("112233aa....");
```

now we can run code like:
``` c#
  static byte[] callscript = ExecutionEngine.EntryScriptHash;
```

add unittest:Test_StaticVarInit

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
@@ -101,9 +101,9 @@ namespace Neo.Compiler.MSIL
                             }
                             else
                             {
+                                //other type mean is not a constValue
                                 constValue = false;
                                 continue;
-                                //throw new Exception("only byte[] can be defined in here.");
                             }
                         }
                         break;
--- a/src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
@@ -183,7 +183,6 @@ namespace Neo.Compiler.MSIL
                                             var bytes = NEO.AllianceOfThinWallet.Cryptography.Base58.Decode(text);
                                             var hash = bytes.Skip(1).Take(20).ToArray();
                                             calcStack.Push(hash);
-
                                         }
                                         else if (attrname == "HexToBytes")//HexString2Bytes to bytes[]
                                         {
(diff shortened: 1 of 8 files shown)


tests/Neo.Compiler.MSIL.UnitTests/UnitTest_StaticVar.cs
```diff
@@ -20,5 +20,28 @@ public void Test_StaticVar()
...
+        [TestMethod]
+        public void Test_StaticVarInit()
+        {
+            ByteArray var1;
+            ByteArray var2;
+            {
+                var testengine = new TestEngine();
+                testengine.AddEntryScript("./TestClasses/Contract_StaticVarInit.cs");
+                var result = testengine.ExecuteTestCaseStandard("staticinit");
+                //test (1+5)*7 == 42
```
shargon: What this means?

src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
```diff
@@ -100,7 +101,9 @@ public static void Parse(ILMethod from, NeoModule to)
                             }
                             else
                             {
-                                throw new Exception("only byte[] can be defined in here.");
+                                constValue = false;
+                                continue;
+                                //throw new Exception("only byte[] can be defined in here.");
```
shargon: If now it doesn't throw, we should remove the comment

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_StaticVar.cs
lightszero: call staticinit and directget and check returnvalues equal
```c#

    static byte[] callscript = ExecutionEngine.EntryScriptHash;

        {
            if (method == "staticinit")
                return testStaticInit();
            if (method == "directget")
                return ExecutionEngine.EntryScriptHash;
            return null;
        }

```

src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
lightszero: yes

tests/Neo.Compiler.MSIL.UnitTests/UnitTest_StaticVar.cs
lightszero: error comment,fix it.

tests/Neo.Compiler.MSIL.UnitTests/Utils/BuildScript.cs
```diff
@@ -59,18 +59,23 @@ public void Build(Stream fs, Stream fspdb)
 
             converterIL = new ModuleConverter(log);
             ConvOption option = new ConvOption();
+#if NDEBUG
```
shargon: Does we need this directives?

lightszero: yes it could be got more clear information when vm error

PR #162, opened 2019-12-10
FeaturesAttribute instead of individual feature attributes to match 3.x

devhawk:
Fixes #158 

When I did #148, I forgot that the master branch has already created a FeaturesAttribute in #129. This PR updates the master-2.x branch NEON + smart contract framework to us a FeaturesAttribute like master branch.

~Note, this PR is in draft waiting for #154 to be merged ~

PR #172, opened 2019-12-17
Revert master-2.x FeaturesAttribute Pull Request

devhawk:
I accidentally merged #162 without approval. This PR reverts that change.


shargon:
In neo2 it's ok for me, because it won't be changed in the future, so only one SC will be per assembly

erikzhang:
Agree with @shargon. This revert is not necessary.

PR #176, opened 2019-12-19
Using attributes to fill manifest Extra property

lock9:
This is a draft while I can't fix the GitHub workflow issue.

### DIFF
--- a/src/Neo.Compiler.MSIL/Program.cs
+++ b/src/Neo.Compiler.MSIL/Program.cs
@@ -288,7 +288,7 @@ namespace Neo.Compiler
         {
             if (extraAttributes.Count == 0)
             {
-                return null;
+                return "{}";
             }
 
             string extra = "{";
--- a/src/Neo.Compiler.MSIL/Program.cs
+++ b/src/Neo.Compiler.MSIL/Program.cs
@@ -288,7 +288,7 @@ namespace Neo.Compiler
         {
             if (extraAttributes.Count == 0)
             {
-                return "{}";
+                return "null";
             }
 
             string extra = "{";
--- a/src/Neo.Compiler.MSIL/Program.cs
+++ b/src/Neo.Compiler.MSIL/Program.cs
@@ -1,9 +1,11 @@
 using CommandLine;
+using Mono.Cecil;
 using Neo.Compiler.MSIL;
 using Neo.Compiler.Optimizer;
 using Neo.SmartContract;
 using Neo.SmartContract.Manifest;
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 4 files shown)


shargon:
Any advance @lock9 ?

Tommo-L:
@shargon @lightszero We need this feature, as neo3 have removed these properties. Sometimes, we want to contact the author of those contracts. I think we can support in our compilers. Could we put more efforts on this PR?

lightszero:
where is  this info will  write to ? abi?

shargon:
Agree too

ShawnYun:
I have updated this pr to the latest complier. And the UT passed. So I think we can convert the draft PR into formal PR.

Tommo-L:
@shargon @lightszero Could help convert this into a formal PR, since @ShawnYun  have tested, all passed.

lock9:
Hello, 
Sorry, I disabled all GitHub notifications. What change you guys want? 
The reason that there is a single and generic property, is that, according to Erik, and I agree, the user may choose to not fill any of this information, so there is no need for "mandatory fields".

Tommo-L:
Thank you @lock9 , it's a good PR.

Tommo-L:
@lock9  If you have time, could you implement it #210 , or after this PR merged, we'll do it later.

src/Neo.Compiler.MSIL/Program.cs
```diff
@@ -278,5 +283,25 @@ public static int Compile(Options options)
 
             return -1;
         }
+
+        private static string BuildExtraAttributes(List<Mono.Collections.Generic.Collection<CustomAttributeArgument>> extraAttributes)
+        {
+            if (extraAttributes.Count == 0)
+            {
+                return null;
```
shargon: It could produce a invalid ABI

shargon:
Maybe we should use constants?

Tommo-L:
> Maybe we should use constants?

？

shargon:
`[ManifestExtra(ManifestExtraAttribute.Author, "Neo")]`

shargon:
Or maybe create different arributes for this
`AuthorAttribute` `EmailAttribute`... etc

Tommo-L:
```c#
[ManifestExtraAttribute("Name", "Demo")]
[ManifestExtraAttribute("Author", "Neo")]
[ManifestExtraAttribute("Email", "dev@neo.org")]
[ManifestExtraAttribute("Description", "This is a contract example")]
```
vs
```c#
NameAttribute("Neo")
AuthorAttribute("Neo")
EmailAttribute("dev@neo.org")
DescriptionAttribute("This is a contract example")
```

 Personally, I prefer the original approach, which is clearer and more organized. 😂

shargon:
> Personally, I prefer the original approach, which is clearer and more organized. 😂

Developers will use it sometimes in lowercase or upercase...

shargon:
What do you think @neo-project/core @lock9 @neo-project/ngd-shanghai @neo-project/ngd-seattle  ?

Tommo-L:
@shargon How about this format:

```c#
@Name("Demo")
@Author("Neo")
@Email("dev@neo.org")
@Description("This is ...")
@ExtraAttribute("Key1", "value1")
public class ContractA : SmartContract
{
     //....
}
```

shargon:
> Or could we read these information from summary directly?

It's more complicated, I prefer in the IL

Tommo-L:
haha...., or let's merge first, we can optimize later. 🤣

Issue #181, opened 2020-01-10
Update neo-devpack-dotnet for new opcode and slots


lightszero:
NEOVM had some change:
1.redesign opcodes
2.delete altstack and add some slot.

I am working on it.

PR #182, opened 2020-01-11
update to support new neo

lightszero:
fixed https://github.com/neo-project/neo-devpack-dotnet/issues/181

finish it.
can merge now.

And need help for fix the native unittest?I am not familiar with that.
@neo-project/core


shargon:
I will help you with UT

src/Neo.SmartContract.Framework/OperandSizeAttribute.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Neo.VM
+{
+    [AttributeUsage(AttributeTargets.Field, AllowMultiple = false)]
+    internal class OperandSizeAttribute : Attribute
```
shargon: We can remove this class if we merge https://github.com/neo-project/neo-vm/pull/285

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -679,15 +682,15 @@ private int ConvertCode(ILMethod method, OpCode src, NeoMethod to)
                 case CodeEx.Brtrue:
                 case CodeEx.Brtrue_S:
                     {
-                        var code = _Convert1by1(VM.OpCode.JMPIF, src, to, new byte[] { 0, 0 });
+                        var code = _Convert1by1(VM.OpCode.JMPIF_L, src, to, new byte[] { 0, 0, 0, 0 });
```
shargon: We should use `JMPIF` if the address is less than one byte

tests/Neo.SmartContract.Framework.UnitTests/Services/Neo/NativeTest.cs
```diff
@@ -30,14 +32,14 @@ public void Init()
...
             using (var script = new ScriptBuilder())
             {
-                script.EmitSysCall(InteropService.Neo_Native_Deploy);
+                script.EmitSysCall(TestEngine.Native_Deploy);
                 _engine.LoadScript(script.ToArray());
-                _engine.Execute();
+                Assert.AreEqual(VMState.HALT, _engine.Execute());
             }
+
+            _engine.AddEntryScript("./TestClasses/Contract_Native.cs");
```
shargon: @lightszero fixed the Deploy method, it's needed for the test, now the failure seems to be other error

src/Neo.SmartContract.Framework/OperandSizeAttribute.cs
lightszero: we can do that on other pr

src/Neo.Compiler.MSIL/MSIL/Converter.cs
lightszero: we can do that on other pr

lightszero: we can do that on other pr

PR #183, opened 2020-01-14
fix notequal with null problem

### LINKED ISSUE #184: Two convert cases will fail
### One case

```csharp
public static bool CheckNullConvertToBigInteger(string code)
{
       var m = Storage.Get(code);
       var k = m?.AsBigInteger();  // <---- this convert will throw ex, but if we remove the `?`, it will not throw
       return k == null;
}
```

Then the compiler will throw the following exceptions 

```
 消息: 
    Initialization method Neo.Compiler.MSIL.UnitTest_NULL.Init threw exception. System.Exception: error:System.Boolean Neo.Compiler.MSIL.TestClasses.Contract_NULL::StorageAndConvertToBigInteger(System.String)::IL_0024 Call System.Boolean System.Nullable`1<System.Numerics.BigInteger>::get_HasValue() ---> System.Exception: unknown call: System.Boolean System.Nullable`1<System.Numerics.BigInteger>::get_HasValue()
       in: System.Boolean Neo.Compiler.MSIL.TestClasses.Contract_NULL::StorageAndConvertToBigInteger(System.String)
    .
  堆栈跟踪: 
    ModuleConverter._ConvertCall(OpCode src, NeoMethod to) 行 742
    ModuleConverter.ConvertCode(ILMethod method, OpCode src, NeoMethod to) 行 888
    ModuleConverter.FillMethod(ILMethod from, NeoMethod to, Boolean withReturn) 行 434
```
> It's problably we have not handle the `System.Boolean System.Nullable`1<System.Numerics.BigInteger>::get_HasValue()`

But if we rewrite the code as following, it will be compiled successful.

```csharp
public static bool CheckNullConvertToBigInteger(string code)
{
       var m = Storage.Get(code);
       var k = m?.AsBigInteger()??0;  
       return k == null;
}
```

### Another case

```csharp
public static bool CheckStorageIsNotNull(string operation, object[] args)
{
        var m = Storage.Get(code); // if code isn't exist in storage
       return m != null;  // <---- The vm will execute failed, as this code will be parsed to OpCode.GT 
 }
```
image

#### How to solve it?
One way is to add `null` support in vm `Opcode.GT`.

Another is to add a check in compiler, and convert to `OpCode.NOTEQUAL` or `OpCode.ISNULL`, which must recognize the `null` parameters.  

This is closed by #183

lightszero:
fixed #184 
fix a notequal with null problem.

in c#
```c#
a!=null
```

now convert to
```
ISNULL
NOT
```

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
@@ -1152,6 +1152,23 @@ namespace Neo.Compiler.MSIL
                 }
             }
         }
+        private int _ConvertCgt(ILMethod method, OpCode src, NeoMethod to)
+        {
+            var code = to.body_Codes.Last().Value;
+            if (code.code == VM.OpCode.PUSHNULL)
+            {
+                //remove last code
+                to.body_Codes.Remove(code.addr);
+                this.addr = code.addr;
+                _Convert1by1(VM.OpCode.ISNULL, src, to);
+                _Convert1by1(VM.OpCode.NOT, src, to);
+            }
+            else
+            {
+                _Convert1by1(VM.OpCode.GT, src, to);
+            }
+            return 0;
+        }
         private int _ConvertCeq(ILMethod method, OpCode src, NeoMethod to)
         {
             var code = to.body_Codes.Last().Value;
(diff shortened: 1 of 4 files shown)


superboyiii:
Tested, it works as expected. Test pass.

superboyiii:
@lightszero This PR solves the second issue. There's another issue inclued in https://github.com/neo-project/neo-devpack-dotnet/issues/184
The first case still has the same issue. Will you go on fixing it in this PR?

Tommo-L:
It works fine.  The C # compiler is a bit powerful, `null! = M` and `m! = Null` both put `null` as the last parameter. 😂

lightszero:
> @lightszero This PR solves the second issue. There's another issue inclued in #184
> The first case still has the same issue. Will you go on fixing it in this PR?

we should disscus that in another issue.

PR #187, opened 2020-02-01
NEP5 Template for NEO3

belane:
C Sharp version of the NEP5 template for NEO3 (https://github.com/neo-project/neo/issues/973).

As the nep-5 standard, it includes `totalSupply`,  `name`,  `symbol`,  `decimals`,  `balanceOf` and `transfer` methods, and the `transfer` event.


shargon:
We need to wait for https://github.com/neo-project/neo/pull/1432

superboyiii:
Good work, I'll test this.

superboyiii:
I tried this nep5 template on my local environment. It was deployed successfully and then I invoked some static methods such like `name` or `decimals` which all returned correct result.
However, when I invoked `deploy` method, I found tx.Cosigner is null which caused execution failed.
image
image
image
Is there any more we should adjust on compiler? Does it support this kind of style? @lightszero

shargon:
Cosigners should be [0], not null :S

doubiliu:
Format needs to be modified

templates/Template.NEP5.CSharp/NEP5.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace Template.NEP5.CSharp
 {
+    [Features(ContractFeatures.HasStorage | ContractFeatures.Payable)]
```
superboyiii: Good, this makes sense.

erikzhang:
@belane Did you see https://github.com/neo-project/examples/tree/3.0/nep-5/csharp/NEP5?

superboyiii:
> Cosigners should be [0], not null, @superboyiii please check [neo-project/neo-node#539](https://github.com/neo-project/neo-node/pull/539)

@shargon I've checked [neo-project/neo-node#539](https://github.com/neo-project/neo-node/pull/539), that's OK. But there's another issue that Cosigners is always [0] here which makes usage is always null so that checkwitness  isn't able to pass when invoke `deploy`.
image

shargon:
You received HALT, and should be FAULT, isn't it? @superboyiii

superboyiii:
> You received HALT, and should be FAULT, isn't it? @superboyiii

Yes, that's it. Now I add some logs on every step of `Deploy()`, you could see that it returns false, but finally receive HALT.
image
image

shargon:
This `CheckWitnessInternal` return false, so the execution must be `FAULT` :S

superboyiii:
Yes, that's strange.

belane:
> @belane Did you see https://github.com/neo-project/examples/tree/3.0/nep-5/csharp/NEP5?

@erikzhang yes, what do you prefer? move it, merge or close?

superboyiii:
@erikzhang  I think two different templates could exist together since they're in absolutely different repositories. We just need to ensure their methods all meet nep5 standard and work well. This template is using Object-oriented style which is valueable for testing purpose.

templates/Template.NEP5.CSharp/NEP5.Admin.cs
```diff
@@ -0,0 +1,53 @@
...
+        private static bool Deploy()
+        {
+            if (!Runtime.CheckWitness(Owner()))
+            {
+                return false;
+            }
+
+            StorageMap contract = Storage.CurrentContext.CreateMap(GetStoragePrefixContract());
+            if (contract.Get("totalSupply").Length != 0)
+                throw new Exception("Contract already deployed");
```
superboyiii: The value of contract.Get("totalSupply").Length will be `null` here when first invoke `deploy` which will cause execution `FAULT`.

doubiliu:
@erikzhang  Both templates are great works. The NEP5 template in the neo/example provides users with a basic Token template, while the Belance's is an extension to meet deeper needs, or we can modify the name of Belance's template to Crowdfunding template.

templates/Template.NEP5.CSharp/NEP5.Admin.cs
doubiliu: change to `if (contract.Get("totalSupply")!= null)`?

templates/Template.NEP5.CSharp/NEP5.Methods.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Numerics;
+
+namespace Template.NEP5.CSharp
+{
+    public partial class NEP5 : SmartContract
+    {
+        private static BigInteger TotalSupply()
+        {
+            StorageMap contract = Storage.CurrentContext.CreateMap(GetStoragePrefixContract());
```
doubiliu: Do a `Nulll `value check
```
            if (contract.Get("totalSupply") is null)
                throw new Exception("Contract is not deployed");
```

superboyiii: > Do a `Null `value check
> 
> ```
>             if (contract.Get("totalSupply") is null)
>                 throw new Exception("Contract is not deployed");
> ```

No, you shouldn't make it like this. At the beginning, contract.Get("totalSupply") is always null. It's normal. For me, it's better like this:
```
BigInteger zero = 0;
            byte[] byteZero = zero.AsByteArray();
            StorageMap contract = Storage.CurrentContext.CreateMap(GetStoragePrefixContract());
            Runtime.Log("StorageMap contract");
            if (contract.Get("totalSupply") != null)
            {
                if (contract.Get("totalSupply").AsBigInteger() != 0)
                {
                    throw new Exception("Contract already deployed");
                }
                
            } else
            {
                contract.Put("totalSupply", byteZero);
            }
```

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,70 @@
...
+            var avaliable_supply = MaxSupply() - current_supply;
+
+            var contribution = (neo * TokensPerNEO()) + (gas * TokensPerGAS());
+            if (contribution <= 0)
+                throw new Exception();
+            if (contribution > avaliable_supply)
+                throw new Exception();
+
+            StorageMap balances = Storage.CurrentContext.CreateMap(GetStoragePrefixBalance());
+            Transaction tx = Blockchain.GetCurrentTransaction();
```
superboyiii: Since https://github.com/neo-project/neo/pull/1432 is closed, should find another way to get current transaction. -_-!

shargon: @belane could you use `GetScriptContainer`

templates/Template.NEP5.CSharp/NEP5.Methods.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                balances.Delete(from);
+            }
+            else
+            {
+                balances.Put(from, fromAmount - amount);
+            }
+
+            BigInteger toAmount = balances.Get(to).AsBigInteger();
+            balances.Put(to, toAmount + amount);
```
superboyiii: `balances.Get(to)` could often be null here which could cause exception. Should be like this:
```
            if (toAmount != null)
            {
                balances.Put(to, toAmount + amount);
            } else
            {
                balances.Put(to, amount);
            }
```

Tommo-L:
```c#
BigInteger toAmount = balances.Get(to)?.AsBigInteger() ?? 0;
```

belane: Thanks, I've added the check to `BalanceOf` too.

templates/Template.NEP5.CSharp/NEP5.Methods.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private static BigInteger TotalSupply()
+        {
+            StorageMap contract = Storage.CurrentContext.CreateMap(GetStoragePrefixContract());
+            return contract.Get("totalSupply").AsBigInteger();
+        }
+
+        private static BigInteger BalanceOf(byte[] account)
+        {
+            if (!ValidateAddress(account)) throw new FormatException("The parameter 'account' SHOULD be 20-byte addresses.");
```
Tommo-L: `throw` ?  This function hasn't been supported in my mind.

Tommo-L: Got, it's already supported and vm will be suspended without error message. It's try-catch not supported.

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var contribution = (neo * TokensPerNEO()) + (gas * TokensPerGAS());
+            if (contribution <= 0)
+                throw new Exception();
+            if (contribution > avaliable_supply)
+                throw new Exception();
+
+            StorageMap balances = Storage.CurrentContext.CreateMap(GetStoragePrefixBalance());
+            Transaction tx = (Transaction)ExecutionEngine.ScriptContainer;
+            var balance = balances.Get(tx.Sender).AsBigInteger();
```
Tommo-L:
```suggestion
            var balance = balances.Get(tx.Sender)?.AsBigInteger()??0;
```

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,70 @@
...
+            if ((string)notification[0] != "Transfer") return 0;
+            // Check dest
+            if ((byte[])notification[2] != ExecutionEngine.ExecutingScriptHash) return 0;
+            // Amount
+            return (BigInteger)notification[3];
+        }
+
+        private static bool Mint()
+        {
+            if (Runtime.InvocationCounter != 1)
```
Tommo-L: I recommend to do it by this way.
```c#
private static bool Mint(byte[] from, BigInteger neoAmount)
{
       // ...
       Neo.Contract.Transfer(from, this.scriptHash, neoAmount); // pseudo code
      //....
}
``` 
(ps: If we can add `from`, `to`, `amount` fields, it'll be better and simpler)

ProDog:
if (operation == "migrate") return Migrate((byte[])args[1], (string)args[2]);
Here's have a problem? args[1].

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
belane: Not sure about this, it's gonna be easier for the user and safer for the project if we get that information from the transaction. Don't you think so?

Tommo-L: Dose user need to combile two scripts, one is neo/gas transfer, another is the `mint` script? If so, I don't think it's very user-friendly.

templates/Template.NEP5.CSharp/NEP5.cs
```diff
@@ -0,0 +1,85 @@
...
+                if (operation == "supportedStandards") return SupportedStandards();
+                #endregion
+
+                #region CROWDSALE METHODS
+                if (operation == "mint") return Mint();
+                #endregion
+
+                #region ADMIN METHODS
+                if (operation == "deploy") return Deploy();
+                if (operation == "migrate") return Migrate((byte[])args[1], (string)args[2]);
```
Tommo-L: start from args[0] ?
```suggestion
                if (operation == "migrate") return Migrate((byte[])args[0], (string)args[1]);
```

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
belane: This is how it currently works, however even if users must specify `from`, `to` and `amount` they will still need to add a contribution building this special Tx (at least until we decide Payable)

templates/Template.NEP5.CSharp/NEP5.cs
```diff
@@ -0,0 +1,85 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static string[] SupportedStandards() => new string[] { "NEP-5", "NEP-10" };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static byte[] Owner() => new byte[] { 0xf6, 0x64, 0x43, 0x49, 0x8d, 0x38, 0x78, 0xd3, 0x2b, 0x99, 0x4e, 0x4e, 0x12, 0x83, 0xc6, 0x93, 0x44, 0x21, 0xda, 0xfe };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong TokensPerNEO() => 1_000_000_000;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong TokensPerGAS() => 1;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static byte[] NeoToken() => new byte[] { 0x9b, 0xde, 0x8f, 0x20, 0x9c, 0x88, 0xdd, 0x0e, 0x7c, 0xa3, 0xbf, 0x0a, 0xf0, 0xf4, 0x76, 0xcd, 0xd8, 0x20, 0x77, 0x89 };
```
ProDog: notification.ScriptHash in contract is little-endian
```suggestion
         public static byte[] NeoToken() => new byte[] { 0x89, 0x77, 0x20, 0xd8, 0xcd, 0x76, 0xf4, 0xf0, 0x0a, 0xbf, 0xa3, 0x7c, 0x0e, 0xdd, 0x88, 0x9c, 0x20, 0x8f, 0xde, 0x9b };
```

templates/Template.NEP5.CSharp/NEP5.cs
```diff
@@ -0,0 +1,85 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static byte[] GasToken() => new byte[] { 0x8c, 0x23, 0xf1, 0x96, 0xd8, 0xa1, 0xbf, 0xd1, 0x03, 0xa9, 0xdc, 0xb1, 0xf9, 0xcc, 0xf0, 0xc6, 0x11, 0x37, 0x7d, 0x3b };
```
ProDog:
```suggestion
       public static byte[] GasToken() => new byte[] { 0x3b, 0x7d, 0x37, 0x11, 0xc6, 0xf0, 0xcc, 0xf9, 0xb1, 0xdc, 0xa9, 0x03, 0xd1, 0xbf, 0xa1, 0xd8, 0x96, 0xf1, 0x23, 0x8c };
```

ProDog:
I tested and found that Runtime.GetNotifications().ScriptHash in contract is little-endian,so NeoToken(),GasToken() should use little-endian.
Owner() too.

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,70 @@
...
+                    neo += GetTransactionAmount(notification.State);
+                }
+                else if (notification.ScriptHash == GasToken())
+                {
+                    gas += GetTransactionAmount(notification.State);
+                }
+            }
+
+            StorageMap contract = Storage.CurrentContext.CreateMap(GetStoragePrefixContract());
+            var current_supply = contract.Get("totalSupply").ToBigInteger();
```
ProDog: If contract.Get("totalSupply") is null, use AsBigInteger() will report error.
```suggestion
            var totalSupplyData = contract.Get("totalSupply");
            BigInteger current_supply = 0;
            if (totalSupplyData != null && totalSupplyData.Length > 0)
            {
                current_supply = totalSupplyData.AsBigInteger();
            }   
```

ProDog: Use AsBigInteger() will report error if contract.Get("totalSupply") is null.
```suggestion
            var totalSupplyData = contract.Get("totalSupply");
            if (totalSupplyData != null && totalSupplyData.Length > 0)
            {
                current_supply = totalSupplyData.AsBigInteger();
            }
```

belane: I did not change this so if the deployment had not been done, 'totalSupply' would fail.
Do you think it is better to return zero when is not yet supply?

belane: Anyway, I've followed your advice and added it to `Mint` and `TotalSupply` to have a uniform behavior.

templates/Template.NEP5.CSharp/NEP5.cs
```diff
@@ -0,0 +1,85 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static string Name() => "Token Name";
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static string Symbol() => "TokenSymbol";
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static byte Decimals() => 8;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong MaxSupply() => 1_000_000_000;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong InitialSupply() => 20_000_000;
```
superboyiii:
```suggestion
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static byte Decimals() => 8;
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static ulong MaxSupply() => 10_000_000_000_000_000;
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static ulong InitialSupply() => 2_000_000_000_000_000;
```
Since the decimals is 8 here, supply number should be adjusted else it will cause issues when invoke `mint`.

belane:
Done @shargon @Tommo-L @superboyiii I think it's ready to go.

superboyiii:
Now it's excellent.  Let's move on!

Tommo-L:
@superboyiii @ProDog Have we tested `Mint` method?

ProDog:
> @superboyiii @ProDog Have we tested `Mint` method?
Yes, I've tested it

tests/Neo.Compiler.MSIL.UnitTests/Utils/NeonTestTool.cs
```diff
@@ -50,11 +51,22 @@ public static byte[] HexString2Bytes(string str)
...
         {
-            var ext = System.IO.Path.GetExtension(filename);
+            return BuildScript(new string[] { filename }, releaseMode, optimizer);
+        }
+
+        /// <summary>
+        /// Build script
+        /// </summary>
+        /// <param name="filenames">Files</param>
+        /// <param name="releaseMode">Release mode (default=false)</param>
```
shargon:
```suggestion
        /// <param name="releaseMode">Release mode (default=false)</param>
        /// <param name="optimizer">Optimize script (default=false)</param>
```

tests/Neo.Compiler.MSIL.UnitTests/Utils/TestEngine.cs
```diff
@@ -48,12 +48,31 @@ public BuildScript Build(string filename, bool releaseMode = false, bool optimiz
...
+        }
+
         public void AddEntryScript(string filename, bool releaseMode = false, bool optimizer = true)
         {
             ScriptEntry = Build(filename, releaseMode, optimizer);
             Reset();
         }
 
+
+        public void AddEntryScript(string[] filenames, bool releaseMode = false)
```
shargon: optimizer as https://github.com/neo-project/neo-devpack-dotnet/pull/187/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.sln#diff-96ba01c5b09aec9a0feacba4d2a11b2dR63 ?

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Template.NEP5.CSharp
+{
+    public partial class NEP5 : SmartContract
+    {
+        private static BigInteger GetTransactionAmount(object state)
+        {
+            var notification = (object[])state;
+            // Checks notification format
+            if (notification.Length != 4) return 0;
```
shargon: <4? if in the future we add more entries we should add them at the end, and the contract could works.
```suggestion
            if (notification.Length < 4) return 0;
```

But this kind of things must be versioning in the abi, in order to don't affect other contracts. so it's good to me like now as well.

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        private static bool Mint()
+        {
+            if (Runtime.InvocationCounter != 1)
+                throw new Exception();
+
+            var notifications = Runtime.GetNotifications();
+            if (notifications.Length == 0)
+                throw new Exception();
```
shargon: only return?

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var notification = (object[])state;
+            // Checks notification format
+            if (notification.Length != 4) return 0;
+            // Only allow Transfer notifications
+            if ((string)notification[0] != "Transfer") return 0;
+            // Check dest
+            if ((byte[])notification[2] != ExecutionEngine.ExecutingScriptHash) return 0;
+            // Amount
+            return (BigInteger)notification[3];
```
shargon: Assert that it's possitive?

templates/Template.NEP5.CSharp/NEP5.Admin.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            StorageMap contract = Storage.CurrentContext.CreateMap(GetStoragePrefixContract());
+            if (contract.Get("totalSupply") != null)
+                throw new Exception("Contract already deployed");
+
+            StorageMap balances = Storage.CurrentContext.CreateMap(GetStoragePrefixBalance());
+            balances.Put(Owner(), InitialSupply());
+            contract.Put("totalSupply", InitialSupply());
+
+            OnTransfer(null, Owner(), InitialSupply());
```
shargon: Store InitialSupply in a variable in order to save gas

templates/Template.NEP5.CSharp/NEP5.Admin.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            if (!Runtime.CheckWitness(Owner()))
+            {
+                return false;
+            }
+            if (script.Length == 0 || manifest.Length == 0)
+            {
+                return false;
+            }
+            Contract.Update(script, manifest);
```
shargon: Shall we send our neo and gas to the next contract or this should be made in the core? 🤔

templates/Template.NEP5.CSharp/NEP5.Admin.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        public static bool Destroy()
+        {
+            if (!Runtime.CheckWitness(Owner()))
+            {
+                return false;
+            }
+
+            Contract.Destroy();
```
shargon: Shall we send our neo and gas to the owner or this should be made in the core? 🤔

templates/Template.NEP5.CSharp/NEP5.Admin.cs
vncoelho: Makes sense, but, in principle, it is already a variable `public static ulong InitialSupply() => 2_000_000_000_000_000;`

templates/Template.NEP5.CSharp/NEP5.Admin.cs
vncoelho: Perhaps there should be a check, if there is neo or gas it returns false.

templates/Template.NEP5.CSharp/NEP5.Admin.cs
Tommo-L: I think we should move asset to the next contract in the core, as NEO, GAS and data all are protocol layer things. For me, we can treat NEO and GAS as one special data type/properties in contract, so when we update contract or destory, we should handle them.

What do you think? @erikzhang @lightszero

templates/Template.NEP5.CSharp/NEP5.Admin.cs
Tommo-L: If one contract hasn't set the verificaiton trigger correctly, it may can't be destoryed never.

belane: These kind of controls are good, we've recommended before, but they are specific to each project.
Each project must implement the measures that fit its business model. As @Tommo-L says if a contract does not include Verification, it is enough to send 1 GAS to prevent them from migrating (or destroy) the contract.

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
belane: At the moment I'd prefer to leave it with the current notifications format, is it ok?

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
belane: What do you mean? If the contribution is not to the contract, it does not count.

templates/Template.NEP5.CSharp/NEP5.Admin.cs
belane: Not sure which is the best option. 🤔 
It would prevent losing assets ​​in hyperspace, but on the other hand, it may exceed the functions of the core.

templates/Template.NEP5.CSharp/NEP5.Admin.cs
belane: Yes, it is and has `AggressiveInlining`.  Do you think we should change it?

lightszero: 1.AggressiveInlining is for JIT engine. totally nouse for us fornow.maybe we can optimize this in future.
and
2."public static ulong InitialSupply() => 2_000_000_000_000_000; " is a function 
still slow and expensive.
=> is very exprensive

"public static ulong InitialSupply = 2_000_000_000_000_000;" is better.

templates/Template.NEP5.CSharp/NEP5.cs
```diff
@@ -0,0 +1,85 @@
...
+        public static string[] SupportedStandards() => new string[] { "NEP-5", "NEP-10" };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static byte[] Owner() => new byte[] { 0xf6, 0x64, 0x43, 0x49, 0x8d, 0x38, 0x78, 0xd3, 0x2b, 0x99, 0x4e, 0x4e, 0x12, 0x83, 0xc6, 0x93, 0x44, 0x21, 0xda, 0xfe };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong TokensPerNEO() => 1_000_000_000;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong TokensPerGAS() => 1;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static byte[] NeoToken() => new byte[] { 0x9b, 0xde, 0x8f, 0x20, 0x9c, 0x88, 0xdd, 0x0e, 0x7c, 0xa3, 0xbf, 0x0a, 0xf0, 0xf4, 0x76, 0xcd, 0xd8, 0x20, 0x77, 0x89 };
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
belane: Very good recommendation, thanks Li.

nicolegys:
We need to send notification when destroying the sc, otherwise the plugin RpcNep5Tracker won't know it, the balance of the asset will still exist in its storage.

erikzhang:
You can check the change set in `OnPersist`.

templates/Template.NEP5.CSharp/NEP5.Crowdsale.cs
shargon: This must be the quantity of NEO and GAS, so it must be a positive number.

belane: Done. very unlikely but check is worth it.

templates/Template.NEP5.CSharp/NEP5.Admin.cs
belane: done

templates/Template.NEP5.CSharp/NEP5.cs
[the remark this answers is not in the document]
belane: I've followed your advice and replaced all of them.
But I have had to keep `SupportedStandards` because it gives a compilation failure. There seems to be some kind of issue with String Arrays.

This is the error what I get in Neo.Compiler.MSIL.Converter.cs when using 
`static readonly string[] SupportedStandards = new string[] { "NEP-5", "NEP-10" };`

image

Please @lightszero can you help me and take a look on it?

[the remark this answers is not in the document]
Tommo-L: I had a try, it compiled successfully, but invoke failure.

image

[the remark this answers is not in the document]
Tommo-L: But in ut, it'll compile failure.

```c#
// Add in Conrtact_Array.cs
static readonly string[] SupportedStandardsVar = new string[] { "NEP-5", "NEP-10" };

public static object SupportedStandards()
{
       return SupportedStandardsVar;
}
```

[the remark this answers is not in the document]
superboyiii: The same error to me, a compilation failure.
```
2>Find entrypoint:System.Object Template.NEP5.CSharp.NEP5::Main(System.String,System.Object[])
2>EXEC : Convert error : System.Exception: sth error
2>   at Neo.Compiler.MSIL.ModuleConverter.LinkCode(String main) in D:\Master-smoking\2020-03-20\neo-devpack-dotnet\src\Neo.Compiler.MSIL\MSIL\Converter.cs:line 359
2>   at Neo.Compiler.MSIL.ModuleConverter.Convert(ILModule _in, ConvOption option) in D:\Master-smoking\2020-03-20\neo-devpack-dotnet\src\Neo.Compiler.MSIL\MSIL\Converter.cs:line 200
2>   at Neo.Compiler.Program.Compile(Options options) in D:\Master-smoking\2020-03-20\neo-devpack-dotnet\src\Neo.Compiler.MSIL\Program.cs:line 159
2>D:\Master-smoking\2020-03-20\neo-devpack-dotnet\templates\Template.NEP5.CSharp\Template.NEP5.CSharp.csproj(12,5): error MSB3073: The command "neon -f "D:\Master-smoking\2020-03-20\neo-devpack-dotnet\templates\Template.NEP5.CSharp\bin\Debug\netstandard2.1\Template.NEP5.CSharp.dll" -o" exited with code -1.
2>Done building project "Template.NEP5.CSharp.csproj" -- FAILED.
```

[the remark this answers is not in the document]
Tommo-L: I'll fix this issue in #211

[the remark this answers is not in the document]
belane: thanks @Tommo-L

belane:
> We need to send notification when destroying the sc, otherwise the plugin RpcNep5Tracker won't know it, the balance of the asset will still exist in its storage.

In the past we haven't used this kind of notifications, as Erik says you can use `OnPersist` better since not all projects will include these notifications. But it's something we can consider adding in the future.

superboyiii:
@lightszero Could you review it again? I think it's OK now.

PR #189, opened 2020-02-11
Optimizer for NEF scripts

shargon:
Optimize NEF before publish the script

- Remove NOPS
- Reduce _L to the small version if it's possible
- All contributions are wellcome


shargon:
> maybe need more unittest to check address

Could you list all the tests you think should be necessary?

lightszero:
1.every opcode with _L need to be test
2.appcall need to be test. 
3.Optimizer need a command line cmd to open or close.

shargon:
> Optimizer need a command line cmd to open or close.

Could you explain more this?

shargon:
We can add `--optimize`, but it will require `-f` for set the file, it's good for you?

lightszero:
> We can add `--optimize`, but it will require `-f` for set the file, it's good for you?

I mean add a if 
···

if(optimizeEnable)
{
                var optimize = new NefOptimizer(bytes).Optimize();
                log.Log("optimization succ " + (((bytes.Length / (optimize.Length + 0.0)) * 100.0) - 100).ToString("0.00 '%'"));
                bytes = optimize;
}

···

var dataa =cmd "neon.exe a.cs"
var datab =cmd "neon.exe a.cs -optimize"

shargon:
Please, @lightszero check it now

lightszero:
ok,I will do some test today.

lightszero:
@shargon I fixed unittest, add a param for optimize ,and default value=true

more than 50% test is fail. so I think we need to check it.

lightszero:
> I think it is worth checking the avm optimizer @igormcoelho created sometime ago.
> 
> I already tested it many times and we also integrated it nativelly on https://neocompiler.io
> 
> https://github.com/NeoResearch/nvm-opt
@vncoelho  maybe you can add it to this pr?

src/Neo.Compiler.MSIL/Optimizer/NefOptimizer.cs
```diff
@@ -0,0 +1,360 @@
...
+                if (instruction.Jump.Offset > 0 && instruction.Offset + instruction.Jump.Offset > offset)
+                {
+                    instruction.Jump.Offset -= lessSize;
+                }
+
+                // Recalculate negative jumps
+
+                else if (instruction.Jump.Offset < 0 && instruction.Offset + instruction.Jump.Offset < offset)
+                {
+                    instruction.Jump.Offset += lessSize;
```
Tommo-L: Is it safe?
```
0: push0
1: push0
2: push0
3: jmp -2
4: jmp_L -3
5: push0
```

If we replace `jmp_L` by `jmp`, the L3 `jmp` seems that it doesn't have to update the offset?

ShawnYun:
Maybe we need to deal with `RecalculateJumps` in different situations:

1. for ` RemoveAt` and `ins.Jump.Offset < 0`, it should be 
```
foreach (var instruction in instructions.Where(u => u.Jump != null))
            {
                if (instruction.Offset < offset && instruction.Jump.Offset > 0 && instruction.Offset + instruction.Jump.Offset > offset)
                {
                    instruction.Jump.Offset -= lessSize;
                }

                else if (instruction.Offset >= offset && instruction.Jump.Offset < 0 && instruction.Offset + instruction.Jump.Offset < offset)
                {
                    instruction.Jump.Offset += lessSize;
                }
            }
```
2. for  `ins.Jump.Offset > 0`, it should be
```
foreach (var instruction in instructions.Where(u => u.Jump != null))
            {

                if (instruction.Offset <= offset && instruction.Jump.Offset > 0 && instruction.Offset + instruction.Jump.Offset > offset)
                {
                    instruction.Jump.Offset -= lessSize;
                }
                else if (instruction.Offset > offset && instruction.Jump.Offset < 0 && instruction.Offset + instruction.Jump.Offset < offset)
                {
                    instruction.Jump.Offset += lessSize;
                }
            }
```

vncoelho:
@lightszero, I am not expert in this part as @igormcoelho.
We are currently focusing on the consensus with our team, in the final design steps and proofs for dBFT 3.0.

Perhaps the feature on https://github.com/NeoResearch/nvm-opt can come on a next version and a new PR.

lightszero:
@Tommo-L @ShawnYun @shargon 
this pr 's idea is good, but the code have some problem.
I will rewrite it thisweek.

ShawnYun:
I've modified `RecalculateJumps`. Then the Unitest passed. But maybe we need to improve it.

shargon:
@ShawnYun could you push your changes?

ShawnYun:
@shargon Sure, I will push it later.

ShawnYun:
@shargon  I have pushed to your branch.

src/Neo.Compiler.MSIL/Optimizer/NefOptimizer.cs
shargon: It should update his offset, maybe we need a test.

lightszero:
I finish my rewrite.
@shargon @Tommo-L @ShawnYun

src/Neo.Compiler.MSIL/Optimizer/Parser_DeleteDeadCode.cs
```diff
@@ -0,0 +1,81 @@
...
+                    }
+                }
+
+                //这些指令之后的指令不可能会执行 （除非由另一条指令跳转而来）
+                if (inst.OpCode == OpCode.JMP ||
+                    inst.OpCode == OpCode.JMP_L ||
+                    inst.OpCode == OpCode.RET ||
+                    inst.OpCode == OpCode.THROW ||
+                    inst.OpCode == OpCode.THROWIF ||
+                    inst.OpCode == OpCode.THROWIFNOT)
```
Tommo-L: Why `OpCode.THROWIF` and `Opcode.THROWIFNOT`? The next instructions of them may be executed. I think we should remove them.

Tommo-L:
@shargon I send a PR, could you merge it https://github.com/shargon/neo-devpack-dotnet/pull/5

shargon:
@Tommo-L I invited you, feel free to change directly this PR

Tommo-L:
> @Tommo-L I invited you, feel free to change directly this PR

Thank you!

src/Neo.Compiler.MSIL/Optimizer/Parser_DeleteDeadCode.cs
lightszero: yes

src/Neo.Compiler.MSIL/Optimizer/NefInstruction.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                uint oplen = GetOperandSize(_OpCode);
+                if (oplen > 0)
+                {
+                    if (Data == null)
+                        Data = new byte[oplen];
+                    else if (Data.Length != oplen)
+                    {
+                        byte[] newdata = new byte[oplen];
+                        Array.Copy(Data, 0, newdata, 0, Math.Min(Data.Length, oplen));
```
Tommo-L: newdata ?
Add `Data = newdata` ?

src/Neo.Compiler.MSIL/Optimizer/NefInstruction.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            uint opprefix = GetOperandPrefixSize(_OpCode);
+            if (opprefix == 0)
+            {
+                uint oplen = GetOperandSize(_OpCode);
+                byte[] newdata = new byte[oplen];
+                Data ??= new byte[0];
+                if (oplen > 0)
+                {
+                    Array.Copy(Data, 0, newdata, 0, Math.Min(Data.Length, oplen));
```
Tommo-L: @lightszero  I'm worried about here. For example, if the original has two 4-byte addresses, we have convert it to two 1-byte addresses.
```c#
TRY_L 0x00000001 0x00000002
```
If use the above code, it will be converted to 
```
TRY 0x00 0x00
```

But it should be 
```c#
TRY 0x01, 0x02
```

lightszero: dont worry. have a real link process now.

src/Neo.Compiler.MSIL/Optimizer/NefInstruction.cs
lightszero: I think you fix wrong.
```c# 
if (Data == null)
       Data = new byte[oplen];
else if (Data.Length != oplen)
{
```
the old code is good.
If data==null then newit
else data !=null then newit and copy partof data from old data to new.

src/Neo.Compiler.MSIL/Optimizer/Parser_UseShortAddress.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Neo.Compiler.Optimizer
+{
+    class Parser_UseShortAddress : IOptimizeParser
+    {
+        public bool NeedRightAddr => false;
+
+        private bool Is8BitAddress(NefInstruction inst)
+        {
+            if (inst.AddressCountInData == 9)
+                return false;
```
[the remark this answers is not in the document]
lightszero: yes,it is a bug

src/Neo.Compiler.MSIL/Optimizer/NefInstruction.cs
Tommo-L:
```c#
Data ??= new byte[0]; 

equals

Data = Data == null ? new byte[0] : Data;
```

Currently, it's okay.

src/Neo.Compiler.MSIL/Optimizer/NefInstruction.cs
Tommo-L: Okay

src/Neo.Compiler.MSIL/Optimizer/NefInstruction.cs
lightszero: new byte[oplen] not new byte[0]

lightszero: if(Data==null)
{
  Data=new byte[oplen]
  return;//?? skip this
}
else if(Data.length!=oplen)
{
  //copy here.
}
else
{
//dot not copy ?? skip this too.
}

lightszero:
@shargon deletedeadcode include deletenop,

shargon:
Very good refactor, good job guys :)

src/Neo.Compiler.MSIL/Optimizer/Parser_DeleteDeadCode.cs
```diff
@@ -30,7 +31,26 @@ private static void Touch(List<INefItem> items, List<int> reachableAddrs, int be
...
-                if (inst.OpCode == OpCode.NOP) continue; // NOP never touch
+
+                bool touch = true;
+                switch (inst.OpCode)
+                {
+                    case OpCode.NOP: touch = false; break;
+                    case OpCode.JMP:
+                        {
+                            // Jump always to the next instruction
+                            if (inst.Data[0] == 2) touch = false;
```
Tommo-L: `inst.Data[0] in {0, 1, 2}` all can skip?

shargon: This change the execution flow

Tommo-L: I'm wrong. I mean, when `inst.Data[0] in {0, 1, 2}`, we can set `touch = false`.  But currently it's enough.

Tommo-L:
@superboyiii  Could you have a test?

src/Neo.Compiler.MSIL/Optimizer/Parser_DeleteDeadCode.cs
lightszero: i think deadcode delete should not remove jmp 2.
take this function  in a new parser is better.
maybe call  deleteUselessJmp.it can delete 
1.jmp to end
2.jmp 2
3.jmp to ret

shargon:
@lightszero we don't have the offset recalculated between optimizers and it makes that the last optimizer fault, any idea?

lightszero:
> @lightszero we don't have the offset recalculated between optimizers and it makes that the last optimizer fault, any idea?

we have

lightszero:
my keyboard is bad...how to delete that milestone.....

Tommo-L:
> my keyboard is bad...how to delete that milestone.....

@lightszero  How to delete the milestone https://help.github.com/en/github/managing-your-work-on-github/creating-and-editing-milestones-for-issues-and-pull-requests#deleting-milestones

src/Neo.Compiler.MSIL/Optimizer/NefOptimizer.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        public void Optimize()
+        {
+            for (var i = 0; i < OptimizeFunctions.Count; i++)
+            {
+                var func = OptimizeFunctions[i];
+                if (i > 0 && func.NeedRightAddr)
+                    RefillAddr();
+                func.Parse(Items);
+            }
```
Tommo-L: Shoud we add a `RefillAddr()` after the `for` block?

lightszero: noneed, Link will do that.

lightszero:
@Tommo-L 
haschangedaddress is not right name.

 that value is not about if a  parser  have a  RightAddr

this value is about a parser  need a RightAddr before it Execute

Tommo-L:
> this value is about a parser need a RightAddr before it Execute

Do you mean some parser don't need the right address? Why not choose to `RefillAddr` after the parser effected the address?

lightszero:
> > this value is about a parser need a RightAddr before it Execute
> 
> Do you mean some parser don't need to care about the right address? Why not choose to `RefillAddr` after the parser effected the address?

yes, some parser do not need right address,if it do some simple work.

define needbefore, we can recalc address just for that parser need rightaddress.

I think's it is different with hadChanged.
we can have needRightAddress and willChangeAddress both.

Tommo-L:
> we can have needRightAddress and willChangeAddress both.

Okay, got your point, I'll add both them.

src/Neo.Compiler.MSIL/Optimizer/NefOptimizer.cs
Tommo-L: I see, I prefer one thing should be done in one function. we can remove `RefillAddr()` in Link process.

src/Neo.Compiler.MSIL/Optimizer/NefOptimizeTool.cs
```diff
@@ -1,41 +1,64 @@
...
+        /// <para> deletedeadcode -- delete dead code parser, default parser</para>
+        /// <para> useshortaddress -- use short address. eg: JMP_L -> JMP, JMPIF_L -> JMPIF, default parser</para>
+        /// <para> deletenop -- delete nop parser</para>
+        /// <para> deleteuselessjmp -- delete useless jmp, eg: JPM 2</para>
+        /// <returns>Optimized script</returns>
+        public static byte[] Optimize(byte[] script, string[] parsers)
+        {
+            var optimizer = new NefOptimizer();
+                
+            if (parsers.Contains("deletedeadcode"))
```
shargon: I prefer an enum @Tommo-L

src/Neo.Compiler.MSIL/Optimizer/Parser_DeleteDeadCode.cs
```diff
@@ -4,7 +4,9 @@ namespace Neo.Compiler.Optimizer
 {
     class Parser_DeleteDeadCode : IOptimizeParser
     {
-        public bool NeedRightAddr => true;
+        private uint OptimizedCount = 0;
```
shargon: This prevent the reuse of this class in other optimizers, or twice in the same if we need to pass twice.

tests/Neo.Compiler.MSIL.UnitTests/Neo.Compiler.MSIL.UnitTests.csproj
```diff
@@ -1,7 +1,10 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <LangVersion>8.0</LangVersion>
```
shargon: latest?

src/Neo.Compiler.MSIL/Optimizer/NefOptimizeTool.cs
Tommo-L: Okay

tests/Neo.Compiler.MSIL.UnitTests/Neo.Compiler.MSIL.UnitTests.csproj
Tommo-L: Always prompt me to upgrade to C#8.0 in my local, do you have the same problem?

shargon: No hahahah, with latest doesn't work for you?

Tommo-L: Yes
I'll reset here.

src/Neo.Compiler.MSIL/Optimizer/Parser_DeleteDeadCode.cs
shargon: it's ok if we have an init method before execute the optimizer

src/Neo.Compiler.MSIL/Neo.Compiler.MSIL.csproj
```diff
@@ -1,11 +1,14 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Copyright>2015-2019 The Neo Project</Copyright>
     <AssemblyTitle>Neo.Compiler.MSIL</AssemblyTitle>
     <Version>3.0.0-preview1</Version>
     <Authors>The Neo Project</Authors>
     <TargetFrameworks>netcoreapp3.1;netstandard2.1</TargetFrameworks>
+    <LangVersion>8.0</LangVersion>
+    <Nullable>enable</Nullable>
```
shargon: I think that we can add nullable in other PR, i like it, but it will require more changes than expected, what do you think @Tommo-L ?

lightszero: I think we do not need this after @Tommo-L update his netcore sdk and vs2019.
^_^.

lightszero:
found a problem
```
        public static int Main(Options options)
```
should not be named as Main here,will got a warnning.

src/Neo.Compiler.MSIL/Optimizer/OptimizeParserType.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    [AttributeUsage(AttributeTargets.Field, AllowMultiple = false)]
+    internal class OptimizeParserAttribute : Attribute
+    {
+        public string Name { get; }
+        public Type Type { get; }
+
+        public OptimizeParserAttribute(string name, Type type)
+        {
+            Name = name;
```
shargon: i think that we don't need the name

Tommo-L: Okay, I'll remove it.

Tommo-L:
> found a problem
> 
> ```
>         public static int Main(Options options)
> ```
> 
> should not be named as Main here,will got a warnning.

How about renaming it 'Compile' ？

lightszero:
@shargon @Tommo-L  we should merge this quickly,too much commit's.

Tommo-L:
Okay, I'll review it again today.

shargon:
It's ready for me

lightszero:
3 agree

PR #204, opened 2020-03-13
Add Features in Template.CSharp project.

chenzhitong:


templates/Template.CSharp/Contract1.cs
```diff
@@ -5,7 +5,8 @@
 
 namespace $safeprojectname$
 {
-	public class Contract1 : SmartContract
+    [Features(ContractFeatures.HasStorage)]
+    public class Contract1 : SmartContract
```
Tommo-L: Should we set the default value is `ContractFeatures.HasStorage` to replace this?

shargon: I don't think so, it's an extra feature

Tommo-L: We should replace here by extra features about `author`, `email`, `date` fields, as we need developers to update it. And the default value of features value is already `ContractFeatures.HasStorage`.

I think we should keep contract code as clean as possible, move configurations to `mainfest.json` as we have a lot of items to set.

vncoelho:
Perhaps we need to clean this folder still, right? There are two csprojec

PR #215, opened 2020-03-19
Resolve visual studio messages

shargon:
- **Fixed a bug with long JMPs**
- Removed `_` from private methods
- Removed some unnecessary initializations
- Removed some unnecessary usings

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Converter.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Converter.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
-using System.Text;
 
 namespace Neo.Compiler.MSIL
 {
@@ -203,10 +202,12 @@ namespace Neo.Compiler.MSIL
         private string InsertAutoEntry()
         {
             string name = "::autoentrypoint";
-            NeoMethod autoEntry = new NeoMethod();
-            autoEntry._namespace = "";
-            autoEntry.name = "Main";
-            autoEntry.displayName = "main";
+            NeoMethod autoEntry = new NeoMethod
+            {
+                _namespace = "",
+                name = "Main",
+                displayName = "main"
+            };
             autoEntry.paramtypes.Add(new NeoParam(name, "string"));
             autoEntry.paramtypes.Add(new NeoParam(name, "array"));
             autoEntry.returntype = "object";
@@ -225,10 +226,10 @@ namespace Neo.Compiler.MSIL
             this.addrconv.Clear();
 
 #if DEBUG
-            _Insert1(VM.OpCode.NOP, "this is a debug code.", to);
+            Insert1(VM.OpCode.NOP, "this is a debug code.", to);
 #endif
-            _insertSharedStaticVarCode(to);
-            _insertBeginCodeEntry(to);
+            InsertSharedStaticVarCode(to);
+            InsertBeginCodeEntry(to);
 
             bool inserted = false;
             List<int> calladdr = new List<int>();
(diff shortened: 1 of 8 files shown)


src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -352,30 +351,29 @@ private void LinkCode(string main)
                     if (c.bytes.Length > 4)
                     {
                         var len = c.bytes.Length - 4;
-                        int wantaddr = addrfunc - c.addr - len;
+                        long wantaddr = addrfunc - c.addr - len;
```
shargon: @lightszero @Tommo-L  check this because I think that it was wrong

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -352,30 +351,29 @@ private void LinkCode(string main)
...
+                        var bts = BitConverter.GetBytes((int)wantaddr);
+                        c.bytes[^4] = bts[0];
+                        c.bytes[^3] = bts[1];
+                        c.bytes[^2] = bts[2];
+                        c.bytes[^1] = bts[3];
                     }
                     else if (c.bytes.Length == 4)
                     {
-                        int wantaddr = addrfunc - c.addr;
+                        long wantaddr = addrfunc - c.addr;
```
shargon: @lightszero @Tommo-L  check this because I think that it was wrong

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -401,7 +399,6 @@ private void FillMethod(ILMethod from, NeoMethod to, bool withReturn)
                     if (src.code == CodeEx.Ret)//before return
                     {
                         if (!withReturn) break;
-                        _insertEndCode(to, src);
```
shargon: This method is not needed anymore, maybe this function could be refactored.

src/Neo.Compiler.MSIL/MSIL/Converter.cs
Tommo-L: Agree, or use `checked`?

src/Neo.Compiler.MSIL/MSIL/Converter.cs
Tommo-L: Agree, or use `checked`?

shargon: Yes, I prefer `checked`

src/Neo.Compiler.MSIL/MSIL/Converter.cs
Tommo-L: It's ok for me, vm will clear automatic.

erikzhang:
@lightszero Please review it.

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -352,30 +351,19 @@ private void LinkCode(string main)
                     if (c.bytes.Length > 4)
                     {
                         var len = c.bytes.Length - 4;
-                        int wantaddr = addrfunc - c.addr - len;
+                        int wantaddr = checked(addrfunc - c.addr - len);
```
lightszero: yes should be long,this is a bug,we can fix it

src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -352,30 +351,19 @@ private void LinkCode(string main)
...
+                        var bts = BitConverter.GetBytes(wantaddr);
+                        c.bytes[^4] = bts[0];
+                        c.bytes[^3] = bts[1];
+                        c.bytes[^2] = bts[2];
+                        c.bytes[^1] = bts[3];
                     }
                     else if (c.bytes.Length == 4)
                     {
-                        int wantaddr = addrfunc - c.addr;
+                        int wantaddr = checked(addrfunc - c.addr);
```
lightszero: should be long,not checked

src/Neo.Compiler.MSIL/MSIL/Converter.cs
lightszero: yes, should use long,not checked.

PR #216, opened 2020-03-23
Update opcodes

### LINKED ISSUE #209: Sync new opcode: ASSERT,  ABORT
waiting for https://github.com/neo-project/neo-vm/pull/300

shargon:
Close #209

- Update Assert and Abort tests
- Create UT for HexToBytes


src/Neo.SmartContract.Framework/SmartContract.cs
```diff
@@ -2,6 +2,10 @@ namespace Neo.SmartContract.Framework
 {
     public class SmartContract
     {
-
+        ///// <summary>
+        ///// Faults if b is false
+        ///// </summary>
+        //[OpCode(OpCode.ASSERT)]
+        //public extern static void Assert(bool b);
```
shargon: @lightszero could you check why I can create this call?

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Helper.cs
```diff
@@ -0,0 +1,30 @@
...
+            (value == true).Assert();
+            return 5;
+        }
+
+        public static void VoidAssertExtension(bool value)
+        {
+            (value == true).Assert();
+        }
+
+        //public static void AssertCall(bool value)
```
shargon: I want to allow both ways

Tommo-L:
close #209

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -74,10 +74,16 @@ public static class Helper
         public extern static bool Within(this int x, BigInteger a, BigInteger b);
 
         /// <summary>
-        /// Faults iff b is false
+        /// Faults if b is false
         /// </summary>
-        [OpCode(OpCode.THROWIFNOT)]
-        public extern static void Assert(this bool b);
+        [OpCode(OpCode.ASSERT)]
+        public extern static void Assert(this bool condition);
```
erikzhang: I don't think this is suitable for implementation as an extension method.

src/Neo.SmartContract.Framework/SmartContract.cs
erikzhang: We should move `Assert()` and `Abort()` to here.

shargon: Agree, @lightszero could you check it why I can't define here this method?

src/Neo.SmartContract.Framework/Helper.cs
shargon: Ok! I will remove it when the other works

shargon:
Now got this error @lightszero @erikzhang 

`Initialization method Neo.SmartContract.Framework.UnitTests.Services.Neo.IteratorTest.Init threw exception. System.TypeLoadException: Could not load type 'Neo.SmartContract.Framework.SmartContract' from assembly 'Neo.SmartContract.Framework, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null' because the method 'Assert' has no implementation (no RVA)..`

shargon:
I can fix it removing the `abstract` keyword, but i am not sure if it will add a empty CALL with empty body after the call

https://github.com/neo-project/neo-devpack-dotnet/pull/216/commits/d5982079dba4413f99e61123285b3f3c1c0e3c8b

lightszero:
i will check it later.

lightszero:
rosyln build  in unittest can't extend a class with extern method.
@shargon 
so we move Abort and Assert to another class will fix this.

src/Neo.SmartContract.Framework/Debug.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Neo.SmartContract.Framework
+{
+    public static class Debug
```
shargon: Runtime?

shargon:
@lightszero with empty body it's not possible to do that? discard the body if this method has `OpCodeAttribute`?

src/Neo.SmartContract.Framework/Debug.cs
lightszero: you can rename it.

lightszero:
> @lightszero with empty body it's not possible to do that? discard the body if this method has `OpCodeAttribute`?

is possible,but we use extern in framework everywhere.

shargon:
nice!!!

erikzhang:
Test failed.

shargon:
I will revert the last change because it will require `GasLeft`

ProDog:
I have tested it passed.
Contract:

image

Invoke result:
for assert(true):

image
image

for assert(false):

image

for abort:

image

ProDog:
plz review it and merge code.

ProDog:
Can merge into master? @erikzhang @lightszero

lightszero:
> Can merge into master? @erikzhang @lightszero

good for me

PR #220, opened 2020-03-25
Fix the number of NEWARR

### LINKED ISSUE #212: A contract compile error when Storage.Put byte[]
Contract compilations report errors when byte[] length exceeds 16,like this:
image
image

Tommo-L:
Close #212 

This reason is that when create a byte array, the size of array may over 16, which opcode is `PUSHDATA[1,24] CONVERT(Integer)` 

@lightszero @shargon @ProDog   plz, have a reivew.


src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -1153,17 +1153,20 @@ private int ConvertNewArr(ILMethod method, OpCode src, NeoMethod to)
             {
                 var code = to.body_Codes.Last().Value;
 
-                if (code.code == VM.OpCode.CONVERT)
-                    code = to.body_Codes.TakeLast(2).First().Value;
```
Tommo-L: Here, we lost one remove operation before.

ProDog:
I have tested it passed.

image

image

PR #223, opened 2020-03-27
Fix byte array reverse

### LINKED ISSUE #217: Contract compile issues for Helper.Reverse()
A contract compiler error for Helper.Reverse(this byte[] source),like:
image

### LINKED ISSUE #218: Contract invoke exceptions for byte[] Concat:
When contract included code:
```
return (new byte[] { 0x12, 0x23, 0x32 }).Concat(new byte[] { 0x55, 0x23 });
```
or
```
Runtime.Notify((new byte[] { 0x12, 0x23, 0x32 }).Concat(new byte[] { 0x55, 0x23 }));
```
it will throw exceptions when invoke contract:

image

Tommo-L:
paritly close #217 close #218 

- support Bytes.Reverse
- add ut for Helper.Concat, Helper.Ranage, Helper.Take, Helper.Last

But, we don't support to assign the element of byte array, eg: `bytes[i] = 0x01`.

@lightszero @shargon @ProDog plz, have a reivew.

### DIFF
--- a/src/Neo.SmartContract.Framework/Helper.cs
+++ b/src/Neo.SmartContract.Framework/Helper.cs
@@ -12,7 +12,7 @@ namespace Neo.SmartContract.Framework
         //const string StackItemType_Boolean = "0x20";
         const string StackItemType_Integer = "0x21";
         const string StackItemType_ByteArray = "0x28";
-        //const string StackItemType_Buffer = "0x30";
+        const string StackItemType_Buffer = "0x30";
         //const string StackItemType_Array = "0x40";
         //const string StackItemType_Struct = "0x41";
         //const string StackItemType_Map = "0x48";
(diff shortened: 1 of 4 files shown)


src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -197,21 +197,31 @@ public static byte ToByte(this int source)
...
-                byte bg = source[k]; // must store on variable before next assignment
-                byte ed = source[m]; // must store on variable before next assignment
-                source[k] = ed;
-                source[m] = bg;
-            }
-            return source;
-        }
+        /// <param name="source"></param>
+        /// <returns></returns>
+        [OpCode(OpCode.CONVERT, StackItemType_Buffer)]
```
erikzhang: I don't think that you need to convert it to `Buffer` first. Because `byte[]` is `Buffer`.

Tommo-L: No, it'll convert `byte[]` to `ByteArray` by default in vm. Therefore, if we want` byte[]` is `Buffer`, we must convert it when compiler once meet the c# `byte[]`.

Tommo-L:
```cs
// StackItem.cs
public static implicit operator StackItem(byte[] value)
{
	return (ByteArray)value;
}
```
https://github.com/neo-project/neo-vm/blob/f292fed3e9add32a2746e2c4cbb569d8dce6f25a/src/neo-vm/Types/StackItem.cs#L70-L74

Tommo-L: It will be more friendly for compiler if we can adjust the `VM.Types.ByteArray` to c# `byte[]`, which with fixed length and mutable, as `Buffer` can be created by `ByteArray`.

erikzhang: That is the vm code, not the contracts. In a contract, `byte[]` should be `Buffer`.

Tommo-L: It may need a lot of work to support this feature, it's better for us to merge this first. We'll add the types of VM in compiler later.

erikzhang: Be patient. We must do the best.

PR #238, opened 2020-04-07
Allow Multiple return

shargon:
Close https://github.com/neo-project/neo-debugger/issues/17

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
@@ -1340,6 +1340,12 @@ namespace Neo.Compiler.MSIL
                 }
             }
 
+            if (type.DeclaringType.FullName.StartsWith("System.ValueTuple`"))
+            {
+                // Multiple returns
+                return 0;
+            }
+
             Convert1by1(VM.OpCode.NOP, src, to);
             ConvertPushNumber(type.DeclaringType.Fields.Count, null, to);
             if (type.DeclaringType.IsValueType)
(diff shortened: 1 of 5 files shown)


shargon:
@lightszero could you check my code, when we return directly the multiple values it works, but the problem is if we use before this ValueTuple, it still use `PICKITEM`.

lightszero:
ValueTuple is new for me，I will study that.

lightszero:
@shargon had fixed

ProDog:
`returnType` is incorrect.

contract:
```
public static (int, int) Return(int a, int b)
 {
            return (a / b, a % b);
 }
```

amnifest.json:
```
 {
            "name":"return",
            "parameters":
            [
                {
                    "name":"a",
                    "type":"Integer"
                },
                {
                    "name":"b",
                    "type":"Integer"
                }
            ],
            "returnType":"Unknown:System.ValueTuple`2<System.Int32,System.Int32>"
        }
```

We need other modifications @lightszero

lightszero:
We can make another issue and pr for it. 
what is returnType should be?
@ProDog

ProDog:
For multiple return, the  `returnType`  is more than one, I think like this:
```
 {
            "name":"return",
            "parameters":
            [
                {
                    "name":"a",
                    "type":"Integer"
                },
                {
                    "name":"b",
                    "type":"Integer"
                }
            ],            
            "returnType":["Integer","Integer"]
}
```

lightszero:
> For multiple return, the `returnType` is more than one, I think like this:
> 
> ```
>  {
>             "name":"return",
>             "parameters":
>             [
>                 {
>                     "name":"a",
>                     "type":"Integer"
>                 },
>                 {
>                     "name":"b",
>                     "type":"Integer"
>                 }
>             ],            
>             "returnType":["Integer","Integer"]
> }
> ```

that is a option,please create a new issue for it.

erikzhang:
It should be `Struct`.

erikzhang:
We shouldn't support multiple return in public methods.

PR #251, opened 2020-04-23
Optimizer: Optimize EQUAL

shargon:


shargon:
@lightszero we need an easy way to detect if we have one JMP to one of the opcodes that we are optimizing.

Example:

```
DUP
JMP myLabel;
PUSH1
*myLabel:*
PUSH0
EQUAL
```
In this case we should not optimize it.

lightszero:
DUP
JMP myLabel;
PUSH1 // I think this opcode should be delete？
*myLabel:*
PUSH0
EQUAL

shargon:
Nice!

PR #253, opened 2020-04-24
Ensure combined rules

shargon:
- Fixed a bug in optimizer with short jumps
- Ensure combined rules


shargon:
Please @lightszero take a look to this script `2204114022fe` it fails during optimization and make the UT fail

shargon:
@lightszero done, we have a wrong address calculation when it's short jump and negative

lightszero:
got it.right,it is my bug.

PR #186, opened 2020-01-21
add trycatch support

lightszero:
working

PR #255, opened 2020-04-26
fix UT

ShawnYun:
fix UT for #186


ShawnYun:
@lightszero Please review it, thanks.

PR #259, opened 2020-04-29
add try-finally ut

Tommo-L:
- Fixed nest try-catch ut
- format, translate comments
- Fixed try-finally ut
- Add try-catch ut

@lightszero Could help merge it, I have fixed the ut.

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Converter.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Converter.cs
@@ -352,7 +352,7 @@ namespace Neo.Compiler.MSIL
                         var srcFinal = BitConverter.ToInt32(c.bytes, 4);
                         if (srcCatch == -1)
                         {
-                            var bytesCatch = new byte[0, 0, 0, 0];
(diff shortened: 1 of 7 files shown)


src/Neo.Compiler.MSIL/MSIL/Converter.cs
```diff
@@ -402,19 +403,25 @@ private int ConvertCode(ILMethod method, OpCode src, NeoMethod to)
                 {
                     if (info.addr_Try_Begin == src.addr)
                     {
-                        var bytesFinally = BitConverter.GetBytes(info.addr_Finally_Begin);
-                        if (info.catch_Infos.Count != 1)
+                        if (info.catch_Infos.Count > 1)
```
shargon, line 406: This logic was changed, it fix a bug?

shargon: @Tommo-L

Tommo-L: We need to support `try-finally` without catch body, which means `info.catch_Infos.Count` is zero.

PR #260, opened 2020-04-29
Fix Parser-DeleteDeadCode

### LINKED ISSUE #246: Modify optimizer of `Parser_DeleteDeadCode`
Now if we use abi offset, we will remove the automain method. The  optimizer  can't  find all method entry points. So we will modify it later.

ShawnYun:
close #246


shargon:
Could you add some ut for `DELETE_DEAD_CODE`?

ShawnYun:
Yes, I will add UT for it.

lightszero:
give me some time,I think need to check this code carefully.

lightszero:
I had checked this code.and I do not think this is a good idea.

we need to split INefItem[]  to several groups by methods.
parse singe group is safe,and do not need to change any code in parser.
and finally,just link back.

lightszero:
@ShawnYun  I made a framework for this
https://github.com/neo-project/neo-devpack-dotnet/pull/263/checks?check_run_id=638155201
we should close this pr and working on that.
can you add unittest for that?

PR #262, opened 2020-04-30
Create and call pointers

shargon:
Require https://github.com/neo-project/neo-vm/pull/317
Required for https://github.com/neo-project/neo/issues/284

Also was added a `DROP` optimization rule 

**Optimized tests require https://github.com/neo-project/neo-devpack-dotnet/pull/263**


shargon:
We need to wait for #263

shargon:
Ready to merge

ProDog:
I have tested it.
- The callback function runs normally without arguments.
- An error occurred while compiling the contract, when the callback function has arguments:

image

In addition, an exception is thrown when deploy contract if `CreateFuncPointer()` is public:

image

shargon:
Could you create a PR with the failing UT?

ProDog:
> Could you create a PR with the failing UT?

I have added it https://github.com/neo-project/neo-devpack-dotnet/pull/269.

PR #269, opened 2020-05-15
Add UT for callback function with arguments

ProDog:
Link https://github.com/neo-project/neo-devpack-dotnet/pull/262#issuecomment-629080574

Add UT for callback function with arguments.

### DIFF
--- a/tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Pointers.cs
+++ b/tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Pointers.cs
@@ -1,5 +1,7 @@
 using Neo.SmartContract.Framework.Services.Neo;
+using Neo.SmartContract.Framework;
 using System;
+using System.Numerics;
 
 namespace Neo.Compiler.MSIL.TestClasses
 {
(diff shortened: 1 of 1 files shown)


tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Pointers.cs
```diff
@@ -19,6 +21,23 @@ public static int CallFuncPointer()
         {
             var pointer = CreateFuncPointer();
             return pointer.Invoke();
-        }
+        }
+
+        public static Func<byte[], BigInteger> CreateFuncPointerWithArg()
+        {
+            return new Func<byte[], BigInteger>(num => MyMethodWithArg(num));
```
shargon, line 28: Could you take a look @lightszero It seems that it produce different names 
```
System.Func`2<System.Byte[],System.Numerics.BigInteger> Neo.Compiler.MSIL.TestClasses.Contract_Pointers::CreateFuncPointerWithArg()
System.Func`2<System.Byte[],System.Numerics.BigInteger> Neo.Compiler.MSIL.TestClasses.Contract_Pointers/<>c::<>9__3_0
```

If we replace it by 

```csharp
 public static Func<byte[], BigInteger> CreateFuncPointerWithArg()
        {
            return new Func<byte[], BigInteger>(MyMethodWithArg);
        }
```

It throw a different error.

ShawnYun:
Could we close it now ?

PR #274, opened 2020-05-26
Fix enumerator

### LINKED ISSUE #272: An contract deploy exception of Enumerator<int> or Enumerator<string>...
If the contract contains return Enumerator\<int> or Enumerator\<string>, an exception will be thrown when the deploy contract.

Contract like :
```C#
        public static Enumerator<int> GetEnumerator()
        {            
            var enumerator = Enumerator<int>.Create(new int[3] { 4, 6, 8 });
            return enumerator;
        }
```

The exception when deploy :

image

Enumerator\<string>, Enumerator\<BigInteger>, Enumerator\<long> have the same exception.

ShawnYun:
close #272


ShawnYun:
wait #271

ProDog:
I tested it, and it worked.🤞

PR #287, opened 2020-06-11
Fix #284

### LINKED ISSUE #284: 7 and more than 7 case branch in switch body is not supported
```
public static bool testswitch(string test)
        {
            int a = 1;
            switch (test)
            {
                case "a":
                    a++;
                    break;
                case "b":
                    a--;
                    break;
                case "c":
                    a = a * 2;
                    break;
                case "d":
                    a = a * (-1);
                    break;
                case "e":
                    a = a * a;
                    break;
                case "f":
                    a = a * 3;
                    break;
                case "g":
                    a = a + 2;
                    break;
                /* 您可以有任意数量的 case 语句 */
                default: /* 可选的 */
                    a = a / 1;
                    break;
            }
            if (a >= 2)
            {
                return true;
            }
            else { return false; }
        }

```
there are two return values for 'a' to 'g', and other values return engine faulted.
Result:
| Value |Result Stack|
|--|--|
| a | [{"type":"ByteString","value":"YQ=="},{"type":"Integer","value":"1"}] |
| b | [{"type":"ByteString","value":"Yg=="},{"type":"Integer","value":"0"}] |
| c | [{"type":"ByteString","value":"Yw=="},{"type":"Integer","value":"1"}]| 
| d | [{"type":"ByteString","value":"ZA=="},{"type":"Integer","value":"0"}]|
| f |  [{"type":"ByteString","value":"Zg=="},{"type":"Integer","value":"1"}] |
| g | [{"type":"ByteString","value":"Zw=="},{"type":"Integer","value":"1"}] | 
| aa| [] | 

If there are less than 7 case branch. such as 6 case branch and default branch.
```
public static bool testswitch(string test)
        {
            int a = 1;
            switch (test)
            {
                case "a":
                    a++;
                    break;
                case "b":
                    a--;
                    break;
                case "c":
                    a = a * 2;
                    break;
                case "d":
                    a = a * (-1);
                    break;
                case "e":
                    a = a * a;
                    break;
                case "f":
                    a = a * 3;
                    break;
                /* 您可以有任意数量的 case 语句 */
                default: /* 可选的 */
                    a = a / 1;
                    break;
            }
            if (a >= 2)
            {
                return true;
            }
            else { return false; }
        }
```

Result:

| Value |Result Stack|
|--|--|
| a | [{"type":"Integer","value":"1"}] |
| b |[{"type":"Integer","value":"0"}]|
| c |[{"type":"Integer","value":"1"}]| 
| d | [{"type":"Integer","value":"0"}]|
| f | [{"type":"Integer","value":"1"}] |
| aa| [{"type":"Integer","value":"0"}] |

lightszero:


tests/Neo.Compiler.MSIL.UnitTests/UnitTest_Switch.cs
```diff
@@ -33,6 +33,44 @@ public void Test_SwitchLong()
             Assert.AreEqual(result.Pop().GetBigInteger(), 99);
         }
 
+        [TestMethod]
+        public void Test_SwitchLongLong()
+        {
+            TestEngine testengine = new TestEngine();
+            testengine.AddEntryScript("./TestClasses/Contract_SwitchLongLong.cs", false, false);
```
shargon, line 40: @lightszero you forgot this file?

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_SwitchLongLong.cs
```diff
@@ -0,0 +1,39 @@
+namespace Neo.Compiler.MSIL.UnitTests.TestClasses
+{
+    class Contract_SwitchInvalid : SmartContract.Framework.SmartContract
```
cloud8little, line 3: if we would support this case, it should be SwitchValid?

PR #291, opened 2020-06-12
add support for switch integer

cloud8little:


lightszero:
Have not implement that yet. I will working on it.

Tommo-L:
plz, waiting for minutes to merge, @ShawnYun  will add some comments.

ShawnYun:
I think we can move `CONVERT` ahead of the for loop?

lightszero:
> I think we can move `CONVERT` ahead of the for loop?

right,that is better

Issue #290, opened 2020-06-12
Mapping between C# types and NeoVM types


erikzhang:
Supported types
----------------
**C# -> NeoVM**
null -> Null
bool -> Integer
sbyte -> Integer
byte -> Integer
short -> Integer
ushort -> Integer
int -> Integer
uint -> Integer
long -> Integer
ulong -> Integer
BigInteger -> Integer
byte[] -> Buffer
char -> Integer
string -> ByteString
Array -> Array
Enum -> Integer
class -> Array
struct -> Struct
tuple -> Struct
event -> System.Runtime.Notify()
lambda -> Pointer
Delegate -> Pointer | Callback

*Any nullable types will also be mapped to the NeoVM types corresponding to their base types. Because all NeoVM types are nullable.*

Unsupported types
------------------------
float
double
decimal

erikzhang:
Can someone check the implementation of the compiler according to this table?

shargon:
I will create a unit test in dev-pack for ensure this behaviour

shargon:
> event -> System.Runtime.Notify()

We should specify that the arguments are `Array`

erikzhang:
Why array?

erikzhang:
Maybe we should add an `eventName` to `Notify()`.

shargon:
> Maybe we should add an eventName to Notify().

Yes, I will

erikzhang:
@shargon https://github.com/neo-project/neo/pull/1701

lightszero:
fix it https://github.com/neo-project/neo-devpack-dotnet/pull/295

PR #295, opened 2020-06-13
fix mapping between C# types and NeoVM types

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/290
Wait for https://github.com/neo-project/neo/pull/1701

### DIFF
--- a/tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_Types.cs
+++ b/tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_Types.cs
@@ -68,10 +68,9 @@ namespace Neo.Compiler.MSIL.UnitTests.TestClasses
             ret.Value = "neo";
             return ret;
         }
-        //public static (string value1, string value2) checkTuple()
-        //{
-        //    var tuple = ("neo", "smart economy");
-        //    return tuple;
-        //}
+        public static (string value1, string value2) checkTuple()
+        {
+            return ("neo", "smart economy");
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


shargon:
Bool and tuple doesn't return the expected type.

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_Types.cs
```diff
@@ -0,0 +1,77 @@
...
+            ret.Value = "neo";
+            return ret;
+        }
+        public static DummyStruct checkStruct()
+        {
+            var ret = new DummyStruct();
+            ret.Value = "neo";
+            return ret;
+        }
+        //public static (string value1, string value2) checkTuple()
```
shargon, line 71: It fail

erikzhang:
@lightszero Can you add tuple support?

erikzhang:
> Bool and tuple doesn't return the expected type.

It seems that there is no `bool` type in MSIL.

shargon:
> It seems that there is no bool type in MSIL.

Yes, so we should change the issue?

`
True:  ldc.i4.1
False: ldc.i4.0
`

erikzhang:
Updated.

lightszero:
I think we support tuple => array  already.
I will check that.

shargon:
@lightszero like this it fail

```
var tuple = ("neo", "smart economy");
            return tuple ;
```

like this it works

```
            return  ("neo", "smart economy");
```

shargon:
`float` works as expected but `Decimal` and `Double` doesn't throw an `unsupported instruction`

PR #296, opened 2020-06-15
Fix operations of string "Concat" and "Substr"

### LINKED ISSUE #294: add operation on string type is not supported
Write a contract such as: 
```
        public static int teststringadd(string s1, string s2)
        {
            int a = 3;
            string c = s1 + s2;
            if(c == "student")
            {
                a = 4;
            }
            if(c == "test")
            {
                a = 5;
            }
            return a;
        }
```
Result: always return 3.
```
neo> invoke 0xe467c1f52065ea194ed6ca779169febc129ba813 testswitch4 [{"type":"String","value":"test"},{"type":"String","value":""}]
Invoking script with: '0c000c047465737412c00c0b74657374737769746368340c1413a89b12bcfe699177cad64e19ea6520f5c167e441627d5b52'
VM State: HALT
Gas Consumed: 0.0109026
Result Stack: [{"type":"Integer","value":"3"}]
Relay tx(no|yes): no
neo> invoke 0xe467c1f52065ea194ed6ca779169febc129ba813 testswitch4 [{"type":"String","value":"te"},{"type":"String","value":"st"}]
Invoking script with: '0c0273740c02746512c00c0b74657374737769746368340c1413a89b12bcfe699177cad64e19ea6520f5c167e441627d5b52'
VM State: HALT
Gas Consumed: 0.0109026
Result Stack: [{"type":"Integer","value":"3"}]
Relay tx(no|yes): no
```

ShawnYun:
close #294 
The `CAT` 、`SUBSTR`、`LEFT` and `RIGHT`  would save the result as `Buffer`，So we can convert them to `ByteString`, but that doesn't seem like a good solution.

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
@@ -513,14 +513,6 @@ namespace Neo.Compiler.MSIL
                     {
                         Convert1by1(VM.OpCode.NUMEQUAL, src, to);
                     }
-                    else if (_ref.DeclaringType.FullName == "System.String")
-                    {
-                        Insert1(VM.OpCode.CONVERT, "", to, new byte[] { (byte)VM.Types.StackItemType.ByteString });
-                        Convert1by1(VM.OpCode.SWAP, null, to);
-                        Insert1(VM.OpCode.CONVERT, "", to, new byte[] { (byte)VM.Types.StackItemType.ByteString });
-                        Convert1by1(VM.OpCode.SWAP, null, to);
-                        Convert1by1(VM.OpCode.EQUAL, src, to);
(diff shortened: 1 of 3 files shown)


erikzhang:
```c#
string s = "s1" + "s2";
```

Since `s` is a `string` type, the result returned by `+` operator should be converted to `ByteString` immediately. It should not be converted when performing equality comparisons.

ShawnYun:
@erikzhang I think we'd better to modify the `CAT` 、`SUBSTR`、`LEFT` and `RIGHT`,  If the operands is `ByteString`, we push `ByteString`, otherwise push `Buffer`.

erikzhang:
No.

ShawnYun:
Have fixed "concat" and "substr" operations and added UTs.

Issue #301, opened 2020-06-21
 Add event name to System.Runtime.Notify


Tommo-L:
Apply https://github.com/neo-project/neo/pull/1701

- Add `EventName` field in Notification entity
- Update  `System.Runtime.Notify(byte[] eventName, Array state)`

Issue #299, opened 2020-06-21
Support callback syscalls


Tommo-L:
Apply https://github.com/neo-project/neo/pull/1629

Support callback syscalls:
- _System.Callback.Create_
- _System.Callback.CreateFromMethod_
- _System.Callback.CreateFromSyscall_
- _System.Callbcak.InvokeCallback_

@shargon Would you add them?

erikzhang:
> System.Callbcak.InvokeCallback

We don't need this in devpack. We should use delegate.

shargon:
> We don't need this in devpack. We should use delegate.

Then how can we distinguish between pointers and `Callbacks`? please take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/302

erikzhang:
One way is we make `System.Callbcak.InvokeCallback` to support both callbacks and pointers.

PR #302, opened 2020-06-22
Add event name, callback support and base64 syscalls

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/301 https://github.com/neo-project/neo-devpack-dotnet/issues/299

- Updated neo nuget
- Added event name
- Added callback syscalls
- Added base64 syscalls

Require https://github.com/neo-project/neo/pull/1723


tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Callback.cs
```diff
@@ -0,0 +1,51 @@
...
+            var action = new Action<object[]>(test2);
+            return Callback.Create(action);
+        }
+
+        public static object createAndCall()
+        {
+            var action = new Action<object[]>(test2);
+            var callback = Callback.Create(action);
+
+            callback.Invoke(new object[0]);
```
shargon, line 37: @lightszero could you take a look why this pointer goes to `INITSLOT`?

```
optimization ratio = 74,22222%
op:[0055]INITSLOT
op:[0058]PUSHA
op:[005D]STLOC0
op:[005E]LDLOC0
op:[005F]PUSH0
op:[0060]SWAP
op:[0061]SYSCALL
op:[0066]STLOC1
op:[0067]LDLOC1
op:[0068]PUSH0
op:[0069]NEWARRAY
op:[006A]SWAP
op:[006B]SYSCALL
op:[000B]INITSLOT
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'index')
   at Neo.VM.EvaluationStack.Remove[T](Int32 index)
   at Neo.VM.ExecutionEngine.ExecuteInstruction(Instruction instruction)
   at Neo.VM.ExecutionEngine.ExecuteNext()
```

tests/Neo.SmartContract.Framework.UnitTests/Services/System/CallbackTest.cs
```diff
@@ -0,0 +1,113 @@
...
+            var item = result.Pop();
+            Assert.IsInstanceOfType(item, typeof(InteropInterface));
+        }
+
+        [TestMethod]
+        public void createAndCallTest()
+        {
+            _engine.Reset();
+            var result = _engine.ExecuteTestCaseStandard("createAndCall");
+            Assert.AreEqual(VMState.HALT, _engine.State);
```
shargon, line 103: @lightszero could you review this test? I think that it try to drop a non existent item

src/Neo.SmartContract.Framework/Services/Neo/Runtime.cs
```diff
@@ -46,7 +46,7 @@ public static class Runtime
         public static extern bool CheckWitness(byte[] hashOrPubkey);
 
         [Syscall("System.Runtime.Notify")]
-        public static extern void Notify(params object[] state);
+        public static extern void Notify(string eventName, params object[] state);
```
erikzhang, line 49: Do we need this?

src/Neo.SmartContract.Framework/Services/System/Callback.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        [Syscall("System.Callback.Invoke")]
+        public extern object Invoke(object[] args);
+
+        // Static
+
+        [OpCode(OpCode.PUSH1)]
+        [OpCode(OpCode.SWAP)]
+        [Syscall("System.Callback.Create")]
+        public static extern Callback Create(Func<object, object> action);
```
erikzhang, line 15: We should support any number of parameters.

src/Neo.SmartContract.Framework/Services/System/Callback.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        [Syscall("System.Callback.CreateFromMethod")]
+        public static extern Callback CreateFromMethod(byte[] hash, string method);
+
+        [Syscall("Neo.Crypto.SHA256")]
+        [OpCode(OpCode.PUSH4)]
+        [OpCode(OpCode.LEFT)]
+        [OpCode(OpCode.CONVERT, Helper.StackItemType_Integer)]
+        [Syscall("System.Callback.CreateFromSyscall")]
+        public static extern Callback CreateFromSyscall(string syscalls);
```
erikzhang, line 25:
```suggestion
        public static extern Callback CreateFromSyscall(uint syscalls);
```

shargon: It could be hard for developers if they need to specify the hash, isn't it?

src/Neo.SmartContract.Framework/Services/System/Callback.cs
shargon: object instead of `Func`?

src/Neo.SmartContract.Framework/Services/Neo/Runtime.cs
shargon: I think that it's another way, you don't need to specify the delegate and the event.

src/Neo.SmartContract.Framework/Services/System/Callback.cs
erikzhang: Computing the hashes in a contract is stupid. We can add a constant table.

src/Neo.SmartContract.Framework/Services/System/Callback.cs
erikzhang: We need the number of parameters. Maybe we should add more overloads.

```
Func<object>
Func<object, object>
Func<object, object, object>
......
Action
Action<object>
Action<object, object>
```

src/Neo.SmartContract.Framework/Services/Neo/Runtime.cs
erikzhang: Without `event`, the compiler cannot generate abi for the it.

shargon: Done

src/Neo.SmartContract.Framework/Services/System/Callback.cs
shargon: `Action` doesn't return any value, so any idea about how can we detect that it's and action in order to push a null value when we create the callback? because `Invoke` return an object...

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract2.cs
```diff
@@ -1,11 +1,19 @@
+using System;
+using System.ComponentModel;
+
 namespace Neo.Compiler.MSIL.UnitTests.TestClasses
 {
     public class Contract2 : SmartContract.Framework.SmartContract
     {
+        public delegate void mydelegate(object arg);
```
erikzhang, line 8: Use `Action<object>`?

src/Neo.SmartContract.Framework/Services/System/Callback.cs
erikzhang: If the return type is `void`, the vm will push `null`. So don't worry about it.

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract2.cs
shargon: Done

src/Neo.SmartContract.Framework/Services/System/Callback.cs
shargon: Much clearer, done! https://github.com/neo-project/neo-devpack-dotnet/pull/302/commits/e1507191a87164bbea845d7cc97ba607317a9b5c

src/Neo.SmartContract.Framework/Services/System/Callback.cs
shargon: Please take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/302/commits/37be246daa396e224d0e0d920a7d50ca8e6940db, it seems that it try to drop the return, and it fail

src/Neo.SmartContract.Framework/Services/System/SyscallCallback.cs
```diff
@@ -0,0 +1,37 @@
+namespace Neo.SmartContract.Framework.Services.System
+{
+    public enum SyscallCallback : uint
+    {
+        System_Binary_Serialize = 0x24011c3f,
```
Tommo-L, line 5: Using the following way? I think it'll be easy for us to verify.
```cs
System_Binary_Serialize = "System.Binary.Serialize".ToMethodHash();
```

shargon: You don't know all the hashes? 😂 haha you are right but `ToMethodHash` it's in `Neo.Compiler.MSIL`, not in the framework ...

```
 System_Binary_Serialize = "System.Binary.Serialize".ToMethodHash(),
        System_Binary_Deserialize = "System.Binary.Deserialize".ToMethodHash(),
        System_Binary_Base64Encode = "System.Binary.Base64Encode".ToMethodHash(),
        System_Binary_Base64Decode = "System.Binary.Base64Decode".ToMethodHash(),
        System_Blockchain_GetHeight = "System.Blockchain.GetHeight".ToMethodHash(),
        System_Blockchain_GetBlock = "System.Blockchain.GetBlock".ToMethodHash(),
        System_Blockchain_GetTransaction = "System.Blockchain.GetTransaction".ToMethodHash(),
        System_Blockchain_GetTransactionHeight = "System.Blockchain.GetTransactionHeight".ToMethodHash(),
        System_Blockchain_GetTransactionFromBlock = "System.Blockchain.GetTransactionFromBlock".ToMethodHash(),
        System_Blockchain_GetContract = "System.Blockchain.GetContract".ToMethodHash(),
        System_Contract_IsStandard = "System.Contract.IsStandard".ToMethodHash(),
        System_Contract_CreateStandardAccount = "System.Contract.CreateStandardAccount".ToMethodHash(),
        Neo_Crypto_RIPEMD160 = "Neo.Crypto.RIPEMD160".ToMethodHash(),
        Neo_Crypto_SHA256 = "Neo.Crypto.SHA256".ToMethodHash(),
        Neo_Crypto_VerifyWithECDsaSecp256r1 = "Neo.Crypto.VerifyWithECDsaSecp256r1".ToMethodHash(),
        Neo_Crypto_VerifyWithECDsaSecp256k1 = "Neo.Crypto.VerifyWithECDsaSecp256k1".ToMethodHash(),
        Neo_Crypto_CheckMultisigWithECDsaSecp256r1 = "Neo.Crypto.CheckMultisigWithECDsaSecp256r1".ToMethodHash(),
        Neo_Crypto_CheckMultisigWithECDsaSecp256k1 = "Neo.Crypto.CheckMultisigWithECDsaSecp256k1".ToMethodHash(),
        System_Json_Serialize = "System.Json.Serialize".ToMethodHash(),
        System_Json_Deserialize = "System.Json.Deserialize".ToMethodHash(),
        System_Runtime_Platform = "System.Runtime.Platform".ToMethodHash(),
        System_Runtime_GetTrigger = "System.Runtime.GetTrigger".ToMethodHash(),
        System_Runtime_GetTime = "System.Runtime.GetTime".ToMethodHash(),
        System_Runtime_GetScriptContainer = "System.Runtime.GetScriptContainer".ToMethodHash(),
        System_Runtime_GetExecutingScriptHash = "System.Runtime.GetExecutingScriptHash".ToMethodHash(),
        System_Runtime_GetCallingScriptHash = "System.Runtime.GetCallingScriptHash".ToMethodHash(),
        System_Runtime_GetEntryScriptHash = "System.Runtime.GetEntryScriptHash".ToMethodHash(),
        System_Runtime_CheckWitness = "System.Runtime.CheckWitness".ToMethodHash(),
        System_Runtime_GetInvocationCounter = "System.Runtime.GetInvocationCounter".ToMethodHash(),
        System_Runtime_GetNotifications = "System.Runtime.GetNotifications".ToMethodHash(),
        System_Runtime_GasLeft = "System.Runtime.GasLeft".ToMethodHash(),
```

If you want to check it you can do this in `TestEngine.cs` cahnge OnSyscall to log it.

```csharp
protected override void OnSysCall(uint method)
        {
            if (callmethod == null)
            {
                var sv = "";
                callmethod = new Dictionary<uint, InteropDescriptor>()
                {
                    { Native_Deploy.Hash , Native_Deploy }
                };
                foreach (var m in ApplicationEngine.Services)
                {
                    callmethod[m.Key] = m.Value;
                    if (m.Value.AllowCallback) sv += $"{m.Value.Name.Replace(".", "_")} = \"{m.Value.Name}\".ToMethodHash(),\n";
                }
            }
            if (callmethod.ContainsKey(method) == false)
            {
                throw new Exception($"Syscall not found: {method:X2} (using base call)");
            }
            else
            {
                base.OnSysCall(method);
            }
        }
```

image

src/Neo.SmartContract.Framework/Services/System/Callback.cs
erikzhang: That is because you call it in the same `ExecutionContext`, which is not allowed. In that case, they will share the stack and the vm won't push `null` to the stack.

shargon: So... We can't create a callbacks for private use? I think that we can use pointers, but how can we distinguish it?

erikzhang: A callback is a `InteropInterface`, and a pointer is a `delegate`.

shargon: But the compiler don't know if you call your own callback, it should warn you to use a pointer instead of a Callback, otherwise it will fail in runtime, should fail in compilation time, right?

ProDog:
I have tested notify, it's OK.

Contract:
```C#
        public delegate void mydelegate(params object[] arg);

        [DisplayName("event")]
        public static event mydelegate Notify;

        [DisplayName("TestEvent")]
        public static event Action<byte[], BigInteger> OnEvent;

        public static bool NotifyTest()
        {
            OnEvent(new byte[] { 0x89, 0x77, 0x20, 0xd8}, 1);

            Notify("str", 11, 12);

            Runtime.Log("test");

            return true;
        }
```
events in abi:
```
"events":
    [
        {
            "name":"event",
            "parameters":
            [
                {
                    "name":"arg",
                    "type":"Array"
                }
            ]
        },
        {
            "name":"TestEvent",
            "parameters":
            [
                {
                    "name":"arg1",
                    "type":"ByteArray"
                },
                {
                    "name":"arg2",
                    "type":"Integer"
                }
            ]
        }
    ]
```
Notify:
```
 "notifications": [
            {
                "contract": "0x90ca09ef53cb3753a3770119b31bf211d410dad5",
                "event_name": "TestEvent",
                "state": {
                    "type": "Array",
                    "value": [
                        {
                            "type": "Buffer",
                            "value": "iXcg2A=="
                        },
                        {
                            "type": "Integer",
                            "value": "1"
                        }
                    ]
                }
            },
            {
                "contract": "0x90ca09ef53cb3753a3770119b31bf211d410dad5",
                "event_name": "event",
                "state": {
                    "type": "Array",
                    "value": [
                        {
                            "type": "Array",
                            "value": [
                                {
                                    "type": "ByteString",
                                    "value": "c3Ry"
                                },
                                {
                                    "type": "Integer",
                                    "value": "11"
                                },
                                {
                                    "type": "Integer",
                                    "value": "12"
                                }
                            ]
                        }
                    ]
                }
            }
```

src/Neo.SmartContract.Framework/Services/System/SyscallCallback.cs
ProDog: We need to make it easy for contract developers to know all the hashes😦

shargon: They will have the enum, they don't need to know the hashes.

shargon: I will create a unit test for help the review, but i think that the enum it's ok like this

shargon: @ProDog @Tommo-L Please take a look to this UT https://github.com/neo-project/neo-devpack-dotnet/pull/302/commits/589eec213f0bb035e84a9746cc00eff0f94bb710#diff-22be6015aa4bd4bfa82c925e1f7e5a47R12

Tommo-L: It's good for me.

src/Neo.SmartContract.Framework/Services/System/Callback.cs
erikzhang: After some thinking, I found that I couldn't find a solution. I think that all we can do at this time is to not allow a contract to call its own callback.

shargon: > After some thinking, I found that I couldn't find a solution. I think that all we can do at this time is to not allow a contract to call its own callback.

If we only allow `Func` and not `Action`, it's not enough?

erikzhang: > If we only allow Func and not Action, it's not enough?

OK.

ProDog:
I found contract execute `FAULT` when the contract uses static variables and delegate event in this PR.
The https://github.com/neo-project/neo-devpack-dotnet/issues/309 is same problem.

My contract:
```C#
        public delegate void mydelegate(params object[] arg);

        [DisplayName("event")]
        public static event mydelegate Notify;

        private static byte[] myAddr = "NNB8GKS7mdMXXGsAwvXYyEGonkEjDbqNkG".ToScriptHash();

        public static byte[] test()
        {
            return myAddr;
        }
```

Invoke result:
```
neo> invoke 0xe7a5322430588a0eb7969100c52a5d8d7bf9642c test
Invoking script with: '10c00c04746573740c142c64f97b8d5d2ac5009196b70e8a58302432a5e741627d5b52'
VM State: FAULT
Gas Consumed: 0.0100889
Result Stack: []
Engine faulted.
Error: insufficient balance.
```

Please take a look @shargon .

shargon:
Could you take a look @lightszero ?

src/Neo.SmartContract.Framework/Services/System/Callback.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        [Syscall("System.Callback.Invoke")]
+        public extern object Invoke(object[] args);
+
+        // Static
+
+        [OpCode(OpCode.PUSH0)]
+        [OpCode(OpCode.SWAP)]
+        [Syscall("System.Callback.Create")]
+        public static extern Callback Create(Action action);
```
erikzhang, line 15: This should be removed too.

shargon: Done

ShawnYun:
We'd better merge it as soon as possible, so we can solve other problems which wait for it.

PR #303, opened 2020-06-22
Add byte.AsString and Map UTs

### LINKED ISSUE #297: Support byte[].AsString()
```cs
public static string testbytearray2(byte[] a)
{
      Map<string, string> some = new Map<string, string>();
     some[a.AsString()] = "teststring2";
     return Json.Serialize(some);
}
```
_posted at https://github.com/neo-project/neo/pull/1710#issuecomment-645947884_

In upper case, we should convert `Buffer` to `ByteString`, as c# byte[] is `NeoVM.Buffer`, c# string is `NeoVM.ByteString`. But the compiler just igore `AsString` method, and return a `Buffer`.

Three problems:
1. Compiler forget to convert `byte[]` in argument parameter list to `Buffer` 
2. `AsString` should convert `Buffer` to `ByteString`.
3.  Map<Primitive, StackItem>, the key of map can only be `Primitive`.

ShawnYun:
Close #297 .
1. (5%)Compiler forget to convert byte[] in argument parameter list to Buffer
2. (5%) AsString should convert Buffer to ByteString.
3. (90%) Map UTs


tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Iterator.cs
```diff
@@ -33,11 +33,11 @@ public static int TestConcatArray(byte[] a, byte[] b)
             return sum;
         }
 
-        public static int TestConcatMap(Map<byte, byte> a, Map<byte, byte> b)
+        public static int TestConcatMap(Map<int, int> a, Map<int, int> b)
```
shargon, line 36: It doesn't work with byte?

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -37,7 +38,7 @@ public static class Helper
         /// <summary>
         /// Converts byte[] to string. Examples: [0x68656c6c6f] -> "hello"; [] -> ""; [0x4e656f] -> "Neo"
         /// </summary>
-        [Script]
+        [OpCode(OpCode.CONVERT, StackItemType_ByteString)]
         public extern static string AsString(this byte[] source);
```
erikzhang, line 42:
```suggestion
        public extern static string ToByteString(this byte[] source);
```

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Iterator.cs
ShawnYun: Yes, if we convert byte[] in argument parameter list to Buffer, it doesn't work.

lightszero: does not work with byte？sounds like a problem

shargon: It should work with both types

lightszero: had test,it can work with byte.

ShawnYun: In neo-core, enumerator only support `Array` and `PrimitiveType`. 
`internal IEnumerator CreateEnumerator(StackItem item)
        {
            return item switch
            {
                Array array => new ArrayWrapper(array),
                PrimitiveType primitive => new ByteArrayWrapper(primitive),
                _ => throw new ArgumentException()
            };
        }`

lightszero: but byte is primitive

ShawnYun: For Map it is ok, I thought it was Enumerator.

shargon: image

It doesn't work for me

erikzhang: `byte[]` is `Buffer`, which is not supported with `IIterator`. We can modify neo to support it.

shargon: I will

shargon: https://github.com/neo-project/neo/pull/1726

Tommo-L:
Waiting for https://github.com/neo-project/neo-devpack-dotnet/pull/302

ShawnYun:
Wait #317 .

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Enumerator.cs
```diff
@@ -4,10 +4,10 @@ namespace Neo.Compiler.MSIL.TestClasses
 {
     public class Contract_Enumerator : SmartContract.Framework.SmartContract
     {
-        public static int TestNext(byte[] a)
+        public static int TestNext(int[] a)
```
shargon, line 7: Could you duplicate it and have `byte` and `int` tests?

ShawnYun: OK

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Iterator.cs
shargon: Done and merged

ShawnYun: How about `Enumerator`? It is not support `Buffer`.

ShawnYun:
Have fixed.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -122,6 +122,11 @@ private void ConvertLdArg(ILMethod method, OpCode src, NeoMethod to, int pos)
             {
                 Convert1by1(VM.OpCode.LDARG, src, to, new byte[] { (byte)pos });
             }
+
+            if (method.paramtypes[pos].type.FullName == "System.Byte[]")
+            {
+                Insert1(VM.OpCode.CONVERT, "", to, new byte[] { (byte)VM.Types.StackItemType.Buffer });
+            }
```
erikzhang, line 129: Why should we do conversion when loading argument?

ShawnYun: Because a byte[] should be convert to `Buffer`.

erikzhang: Sure. But you shouldn't convert it when loading. I think the conversion should be done when the data is pushed onto the stack.

ShawnYun: Yes, here is the opcode `LDARG`, arg will be pushed onto stack.

erikzhang: The conversion should be done when the first time that the data is pushed onto the stack. `LDARG` loads the data from the slot, and the data from the slot is from stack before.

erikzhang: Otherwise, you have to convert it everytime when load from slot.

ShawnYun: The args appears to be pushed on the stack before `Loadscript`? The compiler cannot convert it.

erikzhang: Then don't convert it. The argument should be `string`.

ShawnYun: Or we can check the type of agrs in `INITSLOT`. Now in NeoVM we only check the count, not the type.

Tommo-L: > Or we can check the type of agrs in INITSLOT.

I think it's good, it may not be efficient for compiler to check the parameter type after `INITSLOT`.

PR #316, opened 2020-06-29
Add method to convert from byte array to bigInteger

shargon:
Related to https://github.com/neo-project/neo-devpack-dotnet/pull/304/files#r446113459


src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -52,6 +52,14 @@ public static class Helper
         [OpCode(OpCode.WITHIN)]
         public extern static bool Within(this int x, BigInteger a, BigInteger b);
 
+        /// <summary>
+        /// Converts byte array to BigInteger
+        /// </summary>
+        public static BigInteger AsBigInteger(this byte[] value)
+        {
+            return value?.ToBigInteger() ?? 0;
+        }
```
erikzhang, line 61: I think `As*` means no conversion.

shargon: Changed

Tommo-L:
UT failed.

PR #329, opened 2020-07-23
Some optimizations with HashSet

shargon:


shargon:
@ShawnYun could you review it again please?

ShawnYun:
I think it's good.

PR #330, opened 2020-07-28
Unify template compilations and check by FullName

shargon:
Summary of changes:
- Unify template compilations according to https://github.com/neo-project/neo-devpack-dotnet/blob/ea65ad0fab3b01f2d0608b1c8277796055d49926/templates/Template.NEP5.CSharp/Template.NEP5.CSharp.csproj#L12 because it doesn't require `neon` in the path.
- Check types by `FullName` instead of by `Name`


templates/Template.CSharp/ProjectTemplate.csproj
```diff
@@ -11,7 +11,7 @@
   <Target Name="PostBuild" AfterTargets="PostBuildEvent">
     <Message Text="Start NeoContract converter, Source File: $(TargetPath)" Importance="high">
     </Message>
-    <Exec Command="neon -f &quot;$(TargetPath)&quot; -o" />
+    <Exec Command="dotnet $(SolutionDir)src\Neo.Compiler.MSIL\bin\$(ConfigurationName)\netcoreapp3.1\neon.dll -f &quot;$(TargetPath)&quot; -o" />
```
erikzhang, line 14: Can `netcoreapp3.1` be replaced by something?

shargon: It's possible with `$(TargetFramework)` but it takes the template value that it's `netstandard2.1` instead of `netcoreapp3.1`

PR #347, opened 2020-09-02
Add UInt160 and UInt256

### LINKED ISSUE #208: Add UInt160 and UInt256 data types
**Summary or problem description**

As UInt160, UInt256 are one of the most commonly used data types, I think we should abstract them  into basic types. Here is a [switcheo contract](https://github.com/Switcheo/switcheo-neo/blob/58bbb947d41f0bbec9563231b9f90f39f0f7bfbf/switcheo/BrokerContract.cs), as you can see, scripthash(UInt160) are widely used.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- VM
- Neo
- Compiler

ShawnYun:
Close #208

### DIFF
--- a/src/Neo.Compiler.MSIL/FuncExport.cs
+++ b/src/Neo.Compiler.MSIL/FuncExport.cs
@@ -41,7 +41,9 @@ namespace Neo.Compiler
                 case "System.Int64":
                 case "System.UInt64":
                 case "System.Numerics.BigInteger": return "Integer";
-                case "System.Byte[]": return "ByteArray";
+                case "System.Byte[]":
+                case "Neo.SmartContract.Framework.UInt160":
+                case "Neo.SmartContract.Framework.UInt256": return "ByteArray";
                 case "System.String": return "String";
                 case "IInteropInterface": return "InteropInterface";
                 case "System.Void": return "Void";
--- a/src/Neo.SmartContract.Framework/Helper.cs
+++ b/src/Neo.SmartContract.Framework/Helper.cs
@@ -242,7 +242,7 @@ namespace Neo.SmartContract.Framework
         public extern static byte[] Concat(this byte[] first, byte[] second);
 
         [NonemitWithConvert(ConvertMethod.HexToBytes)]
-        public extern static byte[] HexToBytes(this string hex);
+        public extern static byte[] HexToBytes(this string hex, bool reverse = false);
 
         [OpCode(OpCode.SUBSTR)]
         public extern static byte[] Range(this byte[] source, int index, int count);
--- a/src/Neo.SmartContract.Framework/Helper.cs
+++ b/src/Neo.SmartContract.Framework/Helper.cs
@@ -174,7 +174,7 @@ namespace Neo.SmartContract.Framework
                 var byte1 = str.Substring(0, 1).StringToByte();
                 var byte2 = str.Substring(1, 1).StringToByte();
                 var byte3 = byte1 << 4;
-                var res =  (byte)(byte3 + byte2);
+                var res = (byte)(byte3 + byte2);
                 result[i] = res;
             }
             return new UInt160(result.Reverse());
--- a/src/Neo.SmartContract.Framework/Helper.cs
+++ b/src/Neo.SmartContract.Framework/Helper.cs
@@ -160,6 +160,84 @@ namespace Neo.SmartContract.Framework
             return (byte)(source + 0);
         }
 
+        public static UInt160 HexStringToUInt160(this string value)
+        {
+            var hexString = value;
+            if (value.Length == 42 && value.Substring(0, 2) == "0x")
+                hexString = value.Substring(2, 40);
+            if (hexString.Length != 40)
+                throw new Exception();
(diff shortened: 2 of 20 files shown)


src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -160,6 +160,84 @@ public static byte ToByte(this int source)
...
+            {
+                var str = hexString.Substring(i * 2, 2);
+                var byte1 = str.Substring(0, 1).StringToByte();
+                var byte2 = str.Substring(1, 1).StringToByte();
+                var byte3 = byte1 << 4;
+                var res = (byte)(byte3 + byte2);
+                result[i] = res;
+            }
+            return new UInt160(result.Reverse());
+        }
```
erikzhang, line 181: I think this is too complicated for a contract and may cost a huge amount of GAS.

ShawnYun: Or we can add a InteropService?

ShawnYun: For static variables, we can do the conversion in the compiler. 
But for function arguments, we can't get the parameter values at compile time.

erikzhang: Can you give a scenario?

ShawnYun: For example:
```
static readonly string NeoToken = "0xde5f57d430d3dece511cf975a8d37848cb9e0525";
```
we can get the "0xde5f57d430d3dece511cf975a8d37848cb9e0525" in Compiler.
But if we have a function:
```
        public static UInt256 TestUInt256(String str)
        {
            UInt256 result = str.HexStringToUInt256();
            return result;
        }
```
the Function arguments are passed in when the contract is called.
This is after compiling，so compiler can't handle it.

erikzhang: But why do you use a `string` as the parameter? I think you should use `UInt256` directly.

ShawnYun: On the other hand, the conversion is not used too often, so the GAS consumption is acceptable.

erikzhang: > We can't use a UInt256 as a contract parameter.

Why?

ShawnYun: You mean use `public static UInt256 TestUInt256(UInt256 str)` 
and we transfer a `String` as a parameter?

devhawk: > We can't use a `UInt256` as a contract parameter.

Doesn't Neo already support using UInt160/256 as a contract parameter type? https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/ContractParameterType.cs#L12

devhawk:
What's the progress on this? Can we get this in for Preview 4 https://github.com/neo-project/neo/issues/1936?

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_UInt.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Neo.SmartContract.Framework;
+
+namespace Neo.Compiler.MSIL.TestClasses
+{
+    public class Contract_UInt : SmartContract.Framework.SmartContract
+    {
+        public static UInt160 TestUInt160(String str)
```
devhawk, line 10: We should have tests that take a UInt160/256 as an input parameter too

devhawk:
We should also add ECPoint support to smart contract framework. It behaves similarly to UInt160/256. However, we might want to do this as a separate PR

src/Neo.Compiler.MSIL/FuncExport.cs
```diff
@@ -41,7 +41,9 @@ internal static string ConvType(TypeReference t)
                 case "System.Int64":
                 case "System.UInt64":
                 case "System.Numerics.BigInteger": return "Integer";
-                case "System.Byte[]": return "ByteArray";
+                case "System.Byte[]":
+                case "Neo.SmartContract.Framework.UInt160":
```
devhawk, line 45: Why are we returning UInt160/256 as a byte array?

src/Neo.SmartContract.Framework/NonemitWithConvertAttribute.cs
```diff
@@ -7,6 +7,7 @@ public enum ConvertMethod
         HexToBytes,
         ToScriptHash,
         ToBigInteger,
+        ToUInt160,
```
shargon, line 10: ToUInt256?

devhawk (changes requested):
Obsoleted by https://github.com/neo-project/neo-devpack-dotnet/pull/362

ShawnYun:
As #362, Performing conversion through VM is expensive, so we can close it.

PR #350, opened 2020-09-09
Fix return type of bool

ShawnYun:
Add a Convert to ensure that the return type is Boolean.
According to https://github.com/neo-project/neo/issues/1912

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Converter.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Converter.cs
@@ -416,7 +416,7 @@ namespace Neo.Compiler.MSIL
                     Convert1by1(VM.OpCode.NOP, src, to);
                     break;
                 case CodeEx.Ret:
-                    // Add a Convert to ensure that the return value is of type Bool.
+                    // Add a Convert to ensure that the return type is Boolean.
                     // Because IL uses 1 and 0 for 'True' and 'False'.
                     if (method.returntype.FullName == "System.Boolean")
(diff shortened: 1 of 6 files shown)


erikzhang:
I think it's not necessary. The VM can automatically convert it to `Boolean`.

ShawnYun:
But according to neo-project/neo#1912.
Smart contract return type is irregular when return Boolean result, sometimes it's 1, sometimes it's true.

erikzhang:
> sometimes it's 1, sometimes it's true.

This will not cause any impact.

shargon:
If is not a return the 'error' still here, what it's the IL code generated?

ShawnYun:
Yes, maybe it is confusing.

erikzhang:
> what it's the IL code generated?

There is no `bool` type in IL. I don't think it is an issue.

csmuller:
> > sometimes it's 1, sometimes it's true.
> 
> This will not cause any impact.

If a dApp developer relies on the contract manifest which states that a `Boolean` is returned but gets an `Integer`, that does have an impact. I don't care if booleans are returned as `Boolean` (true, false) or `Integer` (0, 1), what's important is that it must be consistent with what the manifest says.

gsmachado:
> I don't care if booleans are returned as `Boolean` (true, false) or `Integer` (0, 1), what's important is that it must be consistent with what the manifest says.

I completely agree with this. 💯 

If we want to make Neo a developer-friendly platform, I think this argument is pretty reasonable.

PR #354, opened 2020-09-12
Change some native methods to properties

shargon:
- Changed Name, Symbol and Decimals to properties

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
@@ -893,9 +893,20 @@ namespace Neo.Compiler.MSIL
                 }
             }
 
-            var methodName = method.Name;
-            methodName = methodName[..1].ToLowerInvariant() + methodName[1..];
-            return methodName;
+            string methodName = null;
+            if (method.SemanticsAttributes.HasFlag(MethodSemanticsAttributes.Getter))
+            {
+                foreach (PropertyDefinition property in method.DeclaringType.Properties)
+                {
+                    if (property.GetMethod == method)
+                    {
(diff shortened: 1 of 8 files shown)


src/Neo.SmartContract.Framework/Services/Neo/GAS.cs
```diff
@@ -1,15 +1,28 @@
...
 {
     [Contract("0x668e0c1f9d7b70a99dd9e06eadd4c784d641afbc")]
     public class GAS
     {
-        public static extern string Name();
-        public static extern string Symbol();
-        public static extern byte Decimals();
+        public static extern string Name
+        {
+            [DisplayName("name")]
```
erikzhang, line 13: The compiler can read the property name.

shargon: The problem is that he read it as `get_Name` because properties use two methods, one for getters and other for setters.

erikzhang (approved):

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -923,6 +923,32 @@ private int ConvertCall(OpCode src, NeoMethod to)
...
+            foreach (var attr in method.CustomAttributes)
+            {
+                if (attr.AttributeType.Name == nameof(DisplayNameAttribute))
+                {
+                    return (string)attr.ConstructorArguments[0].Value;
+                }
+            }
+
+            string methodName = null;
+            if (method.SemanticsAttributes.HasFlag(MethodSemanticsAttributes.Getter))
```
shargon, line 937: Nice!

PR #362, opened 2020-09-19
Add UInt160, UInt256 and ECPoint to SmartContract Framework

### LINKED ISSUE #208: Add UInt160 and UInt256 data types

devhawk:
fixes #208, part of https://github.com/neo-project/neo/issues/1866
Obsoletes #347 

This PR introduces `Neo.UInt160`, `Neo.UInt256` and `Neo.Cryptography.ECC.ECPoint` to the Neo.SmartContract.Framework assembly. Note, these type namespaces were explicitly chosen to match their corresponding types in Neo.dll. Over the next few weeks, I'll be introducing additional PRs to further unify types in Neo and Neo.SmartContract.Framework as part of https://github.com/neo-project/neo/issues/1866.

Because these types in the two assemblies have the same fully qualified namespace, an assembly that needs to reference Neo and Neo.SmartContract.Framework needs to use a [reference alias](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/extern-alias) to disambiguate. I've added appropriate reference aliases in Neo.Compiler.MSIL and the relevant unit tests projects.

Under the hood, these types are simple byte arrays, like how strings are handled in Neo smart contracts. These types are merely there to provide a small amount of compile time type safety + familiarity to the developer. Relevant service methods (such as `Runtime.CheckWitness`) and domain properties (such as `Block.Hash`) have been updated where appropriate to be strongly typed as `UInt160`, `UInt256` or `ECPoint`.

Assorted Notes:
* SmartContract Framework types that wrap native contracts (such as `NEO` and `Oracle`) have been updated to expose their script hash as a static `Hash` property. To avoid declaring the script hash twice, NEON has been updated to look for a new `NativeContractHashAttribute` as a marker for the appropriate method to return the contract script hash.
* `ConvertNewObj` in NEON has been updated to support the default and byte array constructors of `UInt160` and `UInt256`. Default constructor pushes a byte array of appropriate size onto to stack. Byte array constructors inject VM code to validate the byte array on the evaluation stack is of the expected size
* `ECPoint` in SmartContract.Framework has neither default or byte array constructors. I wasn't sure if these types were ever constructed from byte arrays in contracts. It would be straightforward to include these constructors similar to `UInt160`/`UInt256`'s constructors.
* `UInt160`/`UInt256` each have a static Zero property that pushes a zero array of appropriate length onto the stack via `OpCodeAttribute`
* `FuncExport.ConvType` recognizes the new types and returns the appropriate `ContractParameterType` value as a string
* `ModuleConverter.IsContractCall` was updated to return contract hash out parameter as UInt160 instead of a byte array


devhawk:
@ShawnYun GitHub wouldn't let me request you as an official reviewer. Please review this PR if you have a chance.

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -0,0 +1,20 @@
...
+        public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
+
+        public UInt160()
+        {
+        }
+
+        public UInt160(byte[] _)
+        {
+        }
+
```
Tommo-L, line 16: Could we support `UInt160(string hexstring)`?

devhawk: Neo contracts don't offer run time hex string conversion to byte array. @erikzhang [commented](https://github.com/neo-project/neo-devpack-dotnet/pull/347#discussion_r481928415) on #347 that such conversion would be expensive, and he wasn't sure what the usage scenario would be.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -1448,9 +1450,46 @@ private int ConvertInitObj(OpCode src, NeoMethod to)
             return 0;
         }
 
+        private void PushCheckSize(int size, OpCode src, NeoMethod to)
+        {
+            Insert1(VM.OpCode.DUP, "dup the byte array", to);
+            Insert1(VM.OpCode.SIZE, "get byte array size", to);
```
Tommo-L, line 1456: @erikzhang Could we use `peek()` in Opcode.Size?

https://github.com/neo-project/neo-vm/blob/1b14ec0c20e3a3cddc2263e007a264b3ef90fb62/src/neo-vm/ExecutionEngine.cs#L1027-L1032
```
case OpCode.SIZE:
                    {
                        var x = Pop();
```

erikzhang: No.

src/Neo.SmartContract.Framework/UInt160.cs
shargon: We should check the length of the array, isn't it?

devhawk: > We should check the length of the array, isn't it?

We do. That code is in NEON's ConvertNewObj

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Neo
+{
+    public class UInt160
+    {
+        public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
+
+        public UInt160()
+        {
+        }
```
erikzhang, line 11: Do nothing?

devhawk: The constructors for UInt160/256 are implemented by neon in ConvertNewObj. The default constructor for these types push a byte array of appropriate length onto the stack.

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,8 @@
+namespace Neo.Cryptography.ECC
+{
+    public class ECPoint
+    {
+        [SmartContract.Framework.Script]
+        public static extern implicit operator byte[](ECPoint value);
+    }
```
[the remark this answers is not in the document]
devhawk: There's no conversion from byte[] to ECPoint in Neo.dll. What's the usage scenario for doing this conversion in a smart contract?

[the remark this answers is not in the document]
erikzhang: Some contract may save public keys to storage. When you read it from storage, it is a `ByteString`. You need to convert it to an `ECPoint`.

erikzhang:
We should add `ByteString` too.

src/Neo.SmartContract.Framework/UInt160.cs
erikzhang: `[Script(Opcode.PUSHDATA1, xxxxxx)]`?

devhawk: UInt160/256.Zero static property is implemented via `[OpCode(OpCode.PUSHDATA1, xxxx)]`. I'll remove the default constructors - I don't think we need both. 

Should the byte array constructors use `[Script]` instead of custom code in ConvertNewObj?

src/Neo.SmartContract.Framework/ECPoint.cs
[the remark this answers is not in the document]
devhawk: Good point. I'll add a byte string constructor like what uint160/256 have.

src/Neo.SmartContract.Framework/UInt160.cs
devhawk: In addition to removing the default ctor, I'm also changing the byte array ctor to be an explicit cast using OpCodeAttributes. 

I can't use ScriptAttribute because I need to use PUSHDATA1 as array length for both UInt160 & 256 is larger than 16. ScriptAttribute only appears to support OpCodes without operands.

Since I need multiple OpCode attributes, I can't use a constructor. ConvertNewObj only appears to support a single OpCodeAttribute.

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,18 @@
+using Neo.SmartContract.Framework;
+
+namespace Neo.Cryptography.ECC
+{
+    public class ECPoint
+    {
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHDATA1, "0121")] // 0x01 == 1 byte data length, 0x21 == 33 bytes expected array size
```
erikzhang, line 9:
```suggestion
        [OpCode(OpCode.PUSHINT8, "21")] // 0x21 == 33 bytes expected array size
```

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Neo
+{
+    public class UInt160
+    {
+        public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
+
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHDATA1, "0114")] // 0x01 == 1 byte data length, 0x14 == 20 bytes expected array size
```
erikzhang, line 11:
```suggestion
        [OpCode(OpCode.PUSHINT8, "14")] // 0x14 == 20 bytes expected array size
```

src/Neo.SmartContract.Framework/UInt256.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Neo
+{
+    public class UInt256
+    {
+        public static extern UInt256 Zero { [OpCode(OpCode.PUSHDATA1, "200000000000000000000000000000000000000000000000000000000000000000")] get; }
+
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHDATA1, "0120")] // 0x01 == 1 byte data length, 0x20 == 32 bytes expected array size
```
erikzhang, line 11:
```suggestion
        [OpCode(OpCode.PUSHINT8, "20")] // 0x20 == 32 bytes expected array size
```

erikzhang (changes requested):

src/Neo.SmartContract.Framework/NativeContractHashAttribute.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Neo.SmartContract.Framework
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class NativeContractHashAttribute : Attribute
```
erikzhang, line 6: Rename it to `ContractHashAttribute`? I think it can be used not only for native contracts.

src/Neo.SmartContract.Framework/Services/Neo/Crypto.cs
```diff
@@ -21,19 +21,19 @@ public static class ECDsa
             public static class Secp256r1
             {
                 [Syscall("Neo.Crypto.VerifyWithECDsaSecp256r1")]
-                public extern static bool Verify(byte[] message, byte[] pubkey, byte[] signature);
+                public extern static bool Verify(byte[] message, Cryptography.ECC.ECPoint pubkey, byte[] signature);
```
shargon, line 24: `message` can be a hash too

devhawk: UInt160/256 implicitly convert to byte[]

Tommo-L (approved):

erikzhang:
We must add `ByteString` too. Because `byte[]` in c# will be converted to NeoVM `Buffer`.

devhawk:
> We must add `ByteString` too. Because `byte[]` in c# will be converted to NeoVM `Buffer`.

For explicit byte[] to ECPoint/UInt160/UInt256 conversion, can we add `[OpCode(OpCode.CONVERT, ByteString)]`? Example:

``` csharp
[OpCode(OpCode.CONVERT, Helper.StackItemType_ByteString)]
[OpCode(OpCode.DUP)]
[OpCode(OpCode.SIZE)]
[OpCode(OpCode.PUSHINT8, "14")] // 0x14 == 20 bytes expected array size
[OpCode(OpCode.NUMEQUAL)]
[OpCode(OpCode.ASSERT)]
public static extern explicit operator UInt160(byte[] value);
```

erikzhang:
> For explicit byte[] to ECPoint/UInt160/UInt256 conversion, can we add [OpCode(OpCode.CONVERT, ByteString)]?

But we still have to add `ByteString`.

devhawk:
> Yes. But we still have to add `ByteString`.

What is the usage scenario for adding ByteString? Let's open an issue to discuss and do that work as a separate PR.

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,19 @@
+using Neo.SmartContract.Framework;
+
+namespace Neo.Cryptography.ECC
```
erikzhang, line 3: Can we use a different namespace to avoid alias?

devhawk: The point is for the smart contract framework types to match the types in neo.dll as closely as possible so developers using Neo don't have to learn two ways of doing things. Only the compiler and its tests need to reference both neo.dll and Neo.SmartContract.Framework.dll so no other project will need to use the assembly alias.

erikzhang: `Neo.SmartContract.Framework.dll` is used for smart contracts, and `Neo.dll` is used for nodes and wallets development. They have different scopes, and I don't think they should be confused. But I think it is good to add `UInt160`, `UInt256` and `ECPoint` in Neo.SmartContract.Framework. So I wish to revert the alias-related changes first, so we can merge these classes first.

In addition, we can rename `ECPoint` to `PublicKey`. Because we did not provide any method about `ECPoint`.

devhawk: I can see the value of renaming `ECPoint` to `PublicKey`, but we should make a matching change to RpcClient codebase.

I agree with your point that `Neo.dll` and `Neo.SmartContract.Framework.dll` have different scopes. I was making a similar point that the only project that needs to reference both is NEON (and its tests). Since no other project will need to reference both `Neo.dll` and `Neo.SmartContract.Framework.dll`, no other project will ever need to worry about aliasing. 

Types like `UInt160` and `ECPoint` as well as `Block` and `Transaction` have the same purpose in `Neo.dll` and `Neo.SmartContract.Framework.dll`. Giving types with the same purpose the same names to reduce developer cognitive load is one of the core points of https://github.com/neo-project/neo/issues/1866.

Tommo-L:
@devhawk could fix the conflicts? We hope we can release this feature in preview4.

Tommo-L:
@ProDog  Could have a test?

Tommo-L (dismissed):

ProDog:
Are we need change the type to `UInt160`? at: https://github.com/neo-project/neo-devpack-dotnet/blob/e3d8da36f7ff6eb4854da258f74822b941065c44/src/Neo.SmartContract.Framework/Services/Neo/NEO.cs#L14

and `pubkey` type to `ECPoint`?

ProDog:
Other changes tested OK.

devhawk:
Updated native contract method types as per @ProDog suggestion. There were a couple of places where the smart contract framework was out of sync with the native contract definitions in https://github.com/neo-project/neo

* It looks like `NeoToken.GetValidators` method has changed to `ComputeNextBlockValidators`. 
* It looks like `PolicyContract.GetBlockedAccounts` method has been removed. 

I made appropriate changes to this repo to bring them back into alignment.

Also, several of the `NeoToken` voting related methods (`GetCandidates`, `GetValidators`, `GetCommittee` and `GetNextBlockValidators`) were defined encoding `ECPoint` as `string`. I changed those as well.

src/Neo.SmartContract.Framework/Services/Neo/NEO.cs
```diff
@@ -1,26 +1,28 @@
...
-        public static extern bool Vote(byte[] account, byte[] voteTo);
-        public static extern (string, BigInteger)[] GetCandidates();
-        public static extern string[] GetValidators();
-        public static extern string[] GetCommittee();
-        public static extern string[] GetNextBlockValidators();
+        public static extern bool RegisterCandidate(ECPoint pubkey);
+        public static extern bool UnRegisterCandidate(ECPoint pubkey);
+        public static extern bool Vote(UInt160 account, ECPoint voteTo);
+        public static extern (ECPoint, BigInteger)[] GetCandidates();
+        public static extern ECPoint[] ComputeNextBlockValidators();
```
ProDog, line 24: Now `ComputeNextBlockValidators` is not `ContractMethod`:
https://github.com/neo-project/neo/blob/22f80a7ccfb00af15006569921ed87536cf3f857/src/neo/SmartContract/Native/Tokens/NeoToken.cs#L262

src/Neo.SmartContract.Framework/Services/Neo/NEO.cs
```diff
@@ -1,26 +1,28 @@
...
 
-        public static extern bool RegisterCandidate(byte[] pubkey);
-        public static extern bool UnRegisterCandidate(byte[] pubkey);
-        public static extern bool Vote(byte[] account, byte[] voteTo);
-        public static extern (string, BigInteger)[] GetCandidates();
-        public static extern string[] GetValidators();
-        public static extern string[] GetCommittee();
-        public static extern string[] GetNextBlockValidators();
+        public static extern bool RegisterCandidate(ECPoint pubkey);
+        public static extern bool UnRegisterCandidate(ECPoint pubkey);
```
ProDog, line 21:
```suggestion
        public static extern bool UnregisterCandidate(ECPoint pubkey);
```

src/Neo.SmartContract.Framework/Services/Neo/NEO.cs
Tommo-L, line 24:
```suggestion
        public static extern ECPoint[] GetNextBlockValidators();
```

devhawk: fixed

src/Neo.SmartContract.Framework/Services/Neo/NEO.cs
devhawk: I merged in latest changes

src/Neo.SmartContract.Framework/Services/Neo/NEO.cs
devhawk: merged in latest changes

Tommo-L (approved):

ProDog (approved):

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
+
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHINT8, "14")] // 0x14 == 20 bytes expected array size
+        [OpCode(OpCode.NUMEQUAL)]
+        [OpCode(OpCode.ASSERT)]
+        public static extern explicit operator UInt160(byte[] value);
```
shargon, line 15: method for `IsZero()` or `IsEmpty`? ?

templates/Template.NEP5.CSharp/NEP5.Helpers.cs
```diff
@@ -1,12 +1,11 @@
+using Neo;
 using Neo.SmartContract.Framework;
 using Neo.SmartContract.Framework.Services.Neo;
 
 namespace Template.NEP5.CSharp
 {
     public partial class NEP5 : SmartContract
     {
-        private static bool ValidateAddress(byte[] address) => address.Length == 20 && address.ToBigInteger() != 0;
```
shargon, line 8: Here we check that it's zero, now we don't

Tommo-L:
@devhawk conflicts

devhawk:
> @devhawk conflicts

Thanks @shargon

Tommo-L:
Ping

ShawnYun:
I think this is good, there is no other problem. For undiscovered problems, we can fix them later.

ShawnYun (approved):

shargon (dismissed):

shargon:
I would like to add `IsEmpty` first

shargon (dismissed):

Tommo-L (approved):

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,18 @@
...
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHINT8, "21")] // 0x21 == 33 bytes expected array size
+        [OpCode(OpCode.NUMEQUAL)]
+        [OpCode(OpCode.ASSERT)]
+        public static extern explicit operator ECPoint(byte[] value);
+
+        [Script]
+        public static extern implicit operator byte[](ECPoint value);
```
erikzhang, line 16: I think we should allow conversions between `ECPoint` and `ByteString`. Because `ECPoint` is actually `ByteString` in neo-vm, not `Buffer`.

src/Neo.Compiler.MSIL/FuncExport.cs
```diff
@@ -47,6 +50,9 @@ internal static string ConvType(TypeReference t)
                 case "IInteropInterface": return "InteropInterface";
                 case "System.Void": return "Void";
                 case "System.Object": return "Any";
+                case "Neo.UInt160": return "Hash160";
```
igormcoelho, line 53: Maybe we should stick to naming like "UInt160", and leave "Hash160" to the real operation "RIPEMD160(SHA256(VALUE))". What do you think?
I guess past standards wanted to have this have (Hash160), but I think it's bad to have some object named like this, when content is not precisely that... 
One nice thing would be to have some "Address" object, but that would require NeoVM (and compiler) to have some notion of RIPEMD and SHA256 (which we have removed from there).
So, from a generalization perspective, it's just better to describe the "content of the data", not the "semantics over the data", since we cannot guarantee that to the user.

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,18 @@
+using Neo.SmartContract.Framework;
+
+namespace Neo.Cryptography.ECC
+{
+    public class ECPoint
+    {
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHINT8, "21")] // 0x21 == 33 bytes expected array size
```
igormcoelho, line 10: Maybe we should highlight somewhere that this is a "PublicKey in Compressed Format"... since there are other equivalent formats which we are just ignoring.

src/Neo.Compiler.MSIL/FuncExport.cs
Tommo-L: > Maybe we should stick to naming like "UInt160", and leave "Hash160" to the real operation "RIPEMD160(SHA256(VALUE))". What do you think?

Then we need to update NEP14 https://github.com/neo-project/proposals/blob/d1f4e9e1a67d22a5755c45595121f80b0971ea64/nep-14.mediawiki

Maybe we can update it in another pr.

devhawk: I don't think there's much value to changing those names, but I 100% agree with @Tommo-L that it should be a separate PR

src/Neo.SmartContract.Framework/ECPoint.cs
devhawk: Lets address that in a separate PR

devhawk:
@shargon Shouldn't `IsEmpty` be called `IsZero`? UInts and ECPoints aren't really ever empty, right?

src/Neo.SmartContract.Framework/ECPoint.cs
devhawk: done

templates/Template.NEP5.CSharp/NEP5.Helpers.cs
devhawk: The original version of `ValidateAddress` only checked `address.Length` which is no longer needed, so I removed it. Will add back a `!address.isZero` check

Tommo-L (dismissed):

shargon (dismissed):

src/Neo.SmartContract.Framework/Services/Neo/Designation.cs
```diff
@@ -5,7 +5,8 @@ namespace Neo.SmartContract.Framework.Services.Neo
     [Contract("0x763afecf3ebba0a67568a2c8be06e8f068c62666")]
     public class Designation
     {
+        public static extern UInt160 Hash { [ContractHash] get; }
```
erikzhang, line 8: Is `ContractHashAttribute` necessary? We can default the `Hash` method or property in the classes marked by `ContractAttribute` to return the hash of the contract.

devhawk: It doesn't affect the developers experience so I think it's better to be explicit here.

erikzhang: Then `ContractHashAttribute` should be `internal`.

devhawk: Seems like something 3rd parties would want in building wrapper libraries for their contracts, but I'm OK with taking the more conservative path now. We can always flip this back public if need to later

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,29 @@
...
+        [OpCode(OpCode.NUMEQUAL)]
+        [OpCode(OpCode.ASSERT)]
+        public static extern explicit operator ECPoint(byte[] value);
+
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
```
erikzhang, line 15: This line should be removed.

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -0,0 +1,29 @@
...
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHINT8, "21")] // 0x21 == 33 bytes expected array size
+        [OpCode(OpCode.NUMEQUAL)]
+        [OpCode(OpCode.ASSERT)]
+        public static extern explicit operator ECPoint(ByteString value);
+
+        [Script]
+        public static extern implicit operator byte[](ECPoint value);
```
erikzhang, line 24:
```suggestion
        [OpCode(OpCode.CONVERT, StackItemType.Buffer)]
        public static extern explicit operator byte[](ECPoint value);
```

src/Neo.SmartContract.Framework/ECPoint.cs
devhawk: agreed. Will also remove from UInt160/256

src/Neo.SmartContract.Framework/ECPoint.cs
igormcoelho: I just meant a comment @devhawk 😂

src/Neo.Compiler.MSIL/FuncExport.cs
igormcoelho: Thanks for the clarification, I didn't realize NEP14 had accepted status. We can keep Hash160 with no issue in this case, most of my comments are only related to documentation aspects of the proposal. Congratulations for the amazing job.

src/Neo.SmartContract.Framework/ECPoint.cs
devhawk: If still rather not keep making changes to this PR 😁

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Neo
+{
+    public class UInt160
+    {
+        public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
+
+        public extern bool IsZero
+        {
+            [OpCode(OpCode.CONVERT, StackItemType.Integer)]
```
erikzhang, line 11: The conversion is not necessary. Because `NUMEQUAL` automatically converts the operands to `Integer`s.

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHINT8, "14")] // 0x14 == 20 bytes expected array size
+        [OpCode(OpCode.NUMEQUAL)]
+        [OpCode(OpCode.ASSERT)]
+        public static extern explicit operator UInt160(ByteString value);
+
+        [Script]
+        public static extern implicit operator byte[](UInt160 value);
```
erikzhang, line 33:
```suggestion
        [OpCode(OpCode.CONVERT, StackItemType.Buffer)]
        public static extern explicit operator byte[](UInt160 value);
```

src/Neo.SmartContract.Framework/UInt256.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Neo
+{
+    public class UInt256
+    {
+        public static extern UInt256 Zero { [OpCode(OpCode.PUSHDATA1, "200000000000000000000000000000000000000000000000000000000000000000")] get; }
+
+        public extern bool IsZero
+        {
+            [OpCode(OpCode.CONVERT, StackItemType.Integer)]
```
erikzhang, line 11: Not necessary.

src/Neo.SmartContract.Framework/UInt256.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.SIZE)]
+        [OpCode(OpCode.PUSHINT8, "20")] // 0x20 == 32 bytes expected array size
+        [OpCode(OpCode.NUMEQUAL)]
+        [OpCode(OpCode.ASSERT)]
+        public static extern explicit operator UInt256(ByteString value);
+
+        [Script]
+        public static extern implicit operator byte[](UInt256 value);
```
erikzhang, line 33:
```suggestion
        [OpCode(OpCode.CONVERT, StackItemType.Buffer)]
        public static extern explicit operator byte[](UInt256 value);
```

src/Neo.SmartContract.Framework/UInt160.cs
devhawk: Done, but this has downstream negative effects for using the storage API since put/delete methods lack bytestring overloads

erikzhang: We can add the overloads?

devhawk: Yes, but I think we should do that as a separate PR. Storage API methods have overloads allowing byte[] and string to be used as keys and byte[], string and BigInteger to be used as values. I think we should step back and consider defining storage APIs in terms of ByteString and create implicit operators for types that can be converted to ByteStrings

erikzhang (approved):

erikzhang (approved):

devhawk:
Am I supposed to squash and merge this PR now that it's approved?

ProDog:
> Am I supposed to squash and merge this PR now that it's approved?
I have tested it again, calling all methods using the contract, everything is normal, except for one place it should be `UnregisterCandidate` rather than `UnRegisterCandidate`, needs to be consistent with Neo core. 
Maybe we can fix it in another PR, we need merge this as soon as possible.

Issue #363, opened 2020-09-20
_deploy compilation


shargon:
Please, choose your preference for `_deploy`

1
```csharp
public static void Deploy(bool update) { }
``` 
2
```csharp
private static void _deploy(bool update) { }
``` 
3
```csharp
private static void deploy(bool update) { }
``` 
4
```csharp
[Deploy]
public static void MyMethod(bool update) { }
``` 

Another idea?

devhawk:
#4, but call the attribute "OnDeployment"

erikzhang:
```c#
private static void _deploy(bool update) { }
```

or

```c#
[DisplayName("_deploy")]
public static void MyMethod(bool update) { }
```

We can add a placeholder for this method in the project template.

shargon:
I think that the csharp compiler (Release) could remove the private methods if they are not used

Tommo-L:
Could we also add `OnVerification`?

PR #364, opened 2020-09-20
_deploy compilation

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/363
Require https://github.com/neo-project/neo/pull/1933


Tommo-L:
Don't forget to update the example.

https://github.com/neo-project/neo-devpack-dotnet/blob/10a75b47a69326bfcca278dd6c32ae6f16a14d9d/templates/Template.CSharp/Contract1.cs#L25-L31

shargon:
Ready to review

erikzhang:
Why do we need these attributes? I think we just need to add two methods to the template.

```c#
public bool Verify()
{
}

[DisplayName("_deploy")]
public void Deploy(bool update)
{
}
```

erikzhang (approved):

Issue #168, opened 2019-12-03
Smart contract doesn't support object array


Qiao-Jin:
It seems that neo-compiler & neo-vm don't support object array in smart contract currently, i.e.

```
using Neo.SmartContract.Framework;
using System.Numerics;

namespace TestArray
{
    public class Contract1 : SmartContract
    {
        public static object Main(string operation, object[] args)
        {
            State[] states = new State[] { new State() };
            for (var i = 0; i < 1; i++)
            {
                State state = states[i];
            }
            return true;
        }

        public struct State
        {
            public byte[] from;
            public byte[] to;
            public BigInteger amount;
        }
    }
}
```

The simple smart contract will report an error if built in release mode: 

```
IL_000D Ldelema  ---> System.Exception: unsupported instruction Ldelema
```	

After researching I found out this is due to some lack of functionality in func `neo-compiler\neon\MSIL\Converter.ConvertCode`:

```
case CodeEx.Switch:
{
    throw new Exception("need neo.VM update.");
    //var addrdata = new byte[src.tokenAddr_Switch.Length * 2 + 2];
    //var shortaddrcount = (UInt16)src.tokenAddr_Switch.Length;
    //var data = BitConverter.GetBytes(shortaddrcount);
    //addrdata[0] = data[0];
    //addrdata[1] = data[1];
    //var code = _Convert1by1(VM.OpCode.SWITCH, src, to, addrdata);
    //code.needfix = true;
    //code.srcaddrswitch = new int[shortaddrcount];
    //for (var i = 0; i < shortaddrcount; i++)
    //{
    //    code.srcaddrswitch[i] = src.tokenAddr_Switch[i];
    //}
    }
    break;
```	

This piece of code is commented out because VM.OpCode.SWITCH and corresponding logic is not yet defined. By blaming the code it seems this piece of code is inserted in 2017. I suppose this pending functionality might be forgotten?

Furthermore, even if it can be built successfully under debug mode, still an error would occur if run on chain. After debugging it seems that there is an array bound problem in func `neo-vm\src\neo-vm\ExecutionEngine.ExecuteInstruction`, case OpCode.PICKITEM.

Could anyone have a look at this problem?

shargon:
Try with
states =new State[1];
states[0] =s ;

Tommo-L:
It works well.
![image](https://user-images.githubusercontent.com/43396869/70698682-ed80c200-1d02-11ea-8998-f05effd8d91d.png)

[...]

PR #375, opened 2020-10-02
Fix struct initialization

shargon:
Test for https://github.com/neo-project/neo-devpack-dotnet/issues/168

### DIFF
--- a/tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_Struct.cs
+++ b/tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract_Struct.cs
@@ -1,0 +1,24 @@
+namespace Neo.Compiler.MSIL.UnitTests.TestClasses
+{
+    class Contract_Struct : SmartContract.Framework.SmartContract
+    {
+        public static object Test1(int a, int b, int c)
+        {
+            State[] states = new State[] { new State() };
+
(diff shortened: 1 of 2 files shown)


shargon:
@lightszero Could you take a look? it fails in `setItem`

```
op:[0000]INITSLOT
op:[0003]PUSH1
op:[0004]NEWARRAY
op:[0005]DUP
op:[0006]PUSH0
op:[0007]PICKITEM
op:[0008]STLOC0
op:[0009]LDLOC0
op:[000A]LDARG0
op:[000B]PUSH0
op:[000C]SWAP
op:[000D]SETITEM
```

shargon:
I think that the problem is that it using a ref value.

lightszero:
@shargon checked ,this is not about #168
if you use "class State" it will be fine.
but “struct State" will be crash.

becauce Array in NeoVM is  more like object[]
a struct[] do not need to initialization,it is just a  memory block,c# just fill allbytes to zero.

we do not have a samething like struct[] yet.

lightszero (changes requested):
use class State will be fine.

Qiao-Jin:
I have once post a draft PR #379 whose logic might help for struct initialization (though at first I think this is to solve some problems for class initialization before I realize that class object can be initialized via code in main function).

Issue #383, opened 2020-10-28
Dynamic byte array initialization


shargon:
Author: @csmuller

The next code doesn't compile

```
public static byte[] main(string s)
    {
        return new byte[s.Length];
    }
```

csmuller:
Other array types work fine, e.g. `int[]`. Any dynamic array initialization with `byte[]` fails compilation.
E.g. the following does also not work:
```
public static byte[] TestDynamicArrayInit(int length)
{
        var arrobj = new byte[length];
        ...
}
```

lightszero:
this is right. because bytearray is not Array in NeoVM.it must be inited as a const.

Qiao-Jin:
Similiarly there are also other cases that fail:

```
int[] array = new int[1];
byte[] b = new byte[array[0]];
```

```
List<int> list = new List<int>();
list.Add(0);
byte[] b = new byte[list[0]];
```

```
int minusOne = -1;
byte[] b = new byte[-minusOne];
```

```
int minusOne = -1;
byte[] b = new byte[Math.Abs(minusOne)];
```

```
int index = 0;
byte[] b = new byte[++index];
```

```
int index = 1;
byte[] b = new byte[--index];
```

...etc.

Maybe we should consider other opcodes other than SIZE in this scenario too.

PR #384, opened 2020-10-28
Allow dynamic array with length

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/383


lightszero (changes requested):
but byte array is not a Array in NEOVM.

shargon:
@lightszero please, could you review it again?

lightszero:
good job,I will test this.

lightszero (approved):

PR #393, opened 2020-11-19
publish neon to myget/github packages

devhawk:

PR #395, opened 2020-11-21
Revert "publish neon to myget/github packages"

shargon:
Related to https://github.com/neo-project/neo-devpack-dotnet/pull/393

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,9 +51,7 @@ jobs:
     - name: Setup NuGet.exe for use with actions
       uses: NuGet/setup-nuget@v1
     - name: Pack with dotnet
-      run: |
-        git rev-list --count HEAD | xargs printf "CI%05d" | xargs dotnet pack src/Neo.SmartContract.Framework -c Debug -o out --include-source --version-suffix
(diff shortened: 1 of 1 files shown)


erikzhang:
@devhawk The package of neon is too large. Maybe we can publish the library only (not include the dotnet tools).

devhawk:
> @devhawk The package of neon is too large. Maybe we can publish the library only (not include the dotnet tools).

The tool is the part that matters. I'm not sure what projects use the library directly.

Based on my local testing, `--include-source` is creating an additional 18mb symbols package. Is that getting uploaded to myget? I don't think we need to tool package to include symbols and source

shargon:
Shall I remove `--include-source` first?

erikzhang:
> --include-source is creating an additional 18mb symbols package.

Even without the symbols package, the nupkg file itself is 18MB in size. MyGet only has 100MB of space.

devhawk:
> MyGet only has 100MB of space.

Could we publish at least the neon tool somewhere other than myget? Azure artifacts free tier is 2gb and it's only $2 USD / month / GB above that. Or we could use [sleet](https://github.com/emgarten/Sleet) to publish a feed on AWS or Azure storage - that would be under a $1 USD / month

erikzhang (approved):

PR #391, opened 2020-11-12
Sync neo changes

shargon:
- NEP17 changes https://github.com/neo-project/neo/pull/2024
- Contract hash create changes https://github.com/neo-project/neo/pull/2044
- Require https://github.com/neo-project/neo-devpack-dotnet/pull/390 https://github.com/neo-project/neo/pull/2044
- Close https://github.com/neo-project/neo-devpack-dotnet/issues/400

PR #401, opened 2020-12-05
Safe Attribute

shargon:
Require https://github.com/neo-project/neo/pull/2120 for UT pass
Wait for https://github.com/neo-project/neo-devpack-dotnet/pull/391


shargon:
After merge #391 I think that we can merge this one and fix the UT in https://github.com/neo-project/neo-devpack-dotnet/pull/402 (update nuget) , What do you think?

superboyiii:
I test it merged with https://github.com/neo-project/neo-devpack-dotnet/pull/391 and https://github.com/neo-project/neo-devpack-dotnet/pull/402, it works well.
image
By the way, maybe we should make some method forced to be unsafe when generated such like which including storage write(mint, deploy...)

shargon:
@superboyiii safe should be used wisely :)

superboyiii:
> @superboyiii safe should be used wisely :)

For me it's OK for Preview4.

shargon:
Will be merged with #402 https://github.com/chenzhitong/neo-devpack-dotnet/pull/6

shargon:
Closed because it will come with https://github.com/neo-project/neo-devpack-dotnet/pull/402

Issue #412, opened 2020-12-24
Remove System.Contract.Call


ProDog:
Sync changes of https://github.com/neo-project/neo/pull/2166.

PR #414, opened 2020-12-24
Sync neo-core

Tommo-L:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/412

- Remove `System.Iterator.Concat`, `System.Enumerator.Concat` and `System.Callback.*`
- Add `System.Storage.FindKeys` and `OracleResponseCode`
- Update `CallFlags.cs`, `Contract.Call`


src/Neo.SmartContract.Framework/Services/Neo/Contract.cs
```diff
@@ -27,7 +27,10 @@ public class Contract
         /// </summary>
         public readonly string Manifest;
 
-        [Syscall("System.Contract.Call")]
+        [OpCode(OpCode.REVERSE3)]
+        [OpCode(OpCode.PUSHINT8, "0f")] // 0x0f == 15 CallFlags.All
+        [OpCode(OpCode.REVERSE4)]
+        [Syscall("System.Contract.CallEx")]
         public static extern object Call(UInt160 scriptHash, string method, object[] arguments);
```
shargon, line 34: Maybe we can add a default value in the other method, I think that it should work

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
```diff
@@ -219,5 +219,19 @@ public static class Storage
         [Syscall("System.Storage.GetContext")]
         [Syscall("System.Storage.Find")]
         public static extern Iterator<string, byte[]> Find(string prefix);
+
+        /// <summary>
+        /// Returns a key string enumerator for a byte[] prefix on current Storage context
+        /// </summary>
+        [Syscall("System.Storage.GetContext")]
+        [Syscall("System.Storage.FindKeys")]
+        public static extern Enumerator<byte[]> FindKeys(byte[] prefix, byte removePrefix);
```
shargon, line 228:
```suggestion
        public static extern Iterator<byte[]> FindKeys(byte[] prefix, byte removePrefix);
```

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
```diff
@@ -219,5 +219,19 @@ public static class Storage
...
+
+        /// <summary>
+        /// Returns a key string enumerator for a string prefix on current Storage context
+        /// </summary>
+        [Syscall("System.Storage.GetContext")]
+        [Syscall("System.Storage.FindKeys")]
+        public static extern Enumerator<byte[]> FindKeys(string prefix, byte removePrefix);
```
shargon, line 235:
```suggestion
        public static extern Iterator<byte[]> FindKeys(string prefix, byte removePrefix);
```

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
Tommo-L: But the format of `Iterator`  is `Iterator<TKey, TValue>`.

https://github.com/neo-project/neo/blob/095d7f7f5c5cad070f35f48e37bb189c5bb2a511/src/neo/SmartContract/ApplicationEngine.Storage.cs#L73

src/Neo.SmartContract.Framework/Services/Neo/Contract.cs
Tommo-L: Done

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
shargon: Did you tested with `Enumerator`?

Tommo-L: Have added ut. It passed.

vncoelho (dismissed):

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
erikzhang: It's not an `Iterator`.

erikzhang:
Maybe we should wait for neo-project/neo#2167?

Tommo-L:
We are updating the `Contract.Call`, need one or two more days.

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
```diff
@@ -219,5 +219,19 @@ public static class Storage
...
+        [Syscall("System.Storage.GetContext")]
+        [Syscall("System.Storage.FindKeys")]
+        public static extern Enumerator<byte[]> FindKeys(byte[] prefix, byte removePrefix);
+
+        /// <summary>
+        /// Returns a key string enumerator for a string prefix on current Storage context
+        /// </summary>
+        [Syscall("System.Storage.GetContext")]
+        [Syscall("System.Storage.FindKeys")]
+        public static extern Enumerator<string> FindKeys(string prefix, byte removePrefix);
```
shargon, line 235: I prefer `StorageContext` as argument

erikzhang: This should be removed.

src/Neo.SmartContract.Framework/Services/Neo/Contract.cs
```diff
@@ -27,11 +27,8 @@ public class Contract
         /// </summary>
         public readonly string Manifest;
 
-        [Syscall("System.Contract.Call")]
-        public static extern object Call(UInt160 scriptHash, string method, object[] arguments);
-
         [Syscall("System.Contract.CallEx")]
-        public static extern object CallEx(UInt160 scriptHash, string method, object[] arguments, CallFlags flag);
+        public static extern object Call(UInt160 scriptHash, string method, object[] arguments, CallFlags flag = CallFlags.All);
```
erikzhang, line 31: I think it is impossible to call this SYSCALL directly.

src/Neo.SmartContract.Framework/Services/Neo/Storage.cs
```diff
@@ -219,5 +219,31 @@ public static class Storage
         [Syscall("System.Storage.GetContext")]
         [Syscall("System.Storage.Find")]
         public static extern Iterator<string, byte[]> Find(string prefix);
+
+        /// <summary>
+        /// Returns a key string enumerator for a byte[] prefix on current Storage context
+        /// </summary>
+        [Syscall("System.Storage.GetContext")]
+        [Syscall("System.Storage.FindKeys")]
+        public static extern Enumerator<byte[]> FindKeys(byte[] prefix, FindOptions options);
```
erikzhang, line 228: There is no `Enumerator` in neo any more, right?

shargon:
@ShawnYun could you check why it's failing the Storage UT? `Test_FindKeys`

cloud8little:
> @ShawnYun could you check why it's failing the Storage UT? `Test_FindKeys`

The neo-core is still 3.0.0-CI01130, FindKeys is out-of-date already.

src/Neo.SmartContract.Framework/Services/Neo/Contract.cs
erikzhang: Wait for neo-project/neo#2200

erikzhang:
Wait for neo-project/neo#2211

erikzhang:
https://github.com/neo-project/neo/blob/d0a4976e9b06883a47426efc60a87c0d53fcd14f/src/neo/SmartContract/ContractState.cs#L11-L17

Now the `ContractState` class has the `Nef` field.

erikzhang:
The compiler should use `CALLT` for static calls.

shargon:
@erikzhang 

Any rule for implicit / explicit ? I think that it will be easier if we use implicit conversion from `byte[]`

```csharp
        [Script]
        public static extern implicit operator string(ByteString str);

        [Script]
        public static extern implicit operator ByteString(string str);

        public static extern explicit operator byte[](ByteString str);

        public static extern explicit operator ByteString(byte[] buffer);
```

erikzhang:
`implicit` means doing nothing. `explicit` emits some opcodes.

shargon:
If we don't use implicit, we will have 'cast' everywhere.

shargon:
> The compiler should use CALLT for static calls.

Can we do this in another PR?

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -924,7 +927,11 @@ private int ConvertCall(OpCode src, NeoMethod to)
             else if (calltype == 6)
             {
                 ConvertPushNumber(callpcount, src, to);
+                Convert1by1(VM.OpCode.PACK, null, to);
                 Convert1by1(VM.OpCode.ROLL, null, to);
+                Convert1by1(VM.OpCode.REVERSE3, null, to);
```
shargon, line 932: @Tommo-L @lightszero could you help me with this call (6) It has no comments.

erikzhang:
> Can we do this in another PR?

Sure.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
Tommo-L: It seems that `calltype == 6` means delegate call.

https://github.com/neo-project/neo-devpack-dotnet/blob/ba73a45fe9eb012cf19248acde0e6a8f0f98bae9/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs#L76-L94

shargon: @Tommo-L could you add a UT for hit this lines, I was not able to compile a valid call with Delegates

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -201,6 +201,15 @@ public static byte ToByte(this int source)
         [NonemitWithConvert(ConvertMethod.ToBigInteger)]
         public extern static BigInteger ToBigInteger(this string text);
```
erikzhang, line 202: BigInteger.Parse()?

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -201,6 +201,15 @@ public static byte ToByte(this int source)
...
 
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.ISNULL)]
+        [OpCode(OpCode.JMPIFNOT, "0x05")]
+        [OpCode(OpCode.PUSH0)]
+        [OpCode(OpCode.SWAP)]
+        [OpCode(OpCode.DROP)]
+        [OpCode(OpCode.CONVERT, StackItemType.Integer)]
+        public extern static BigInteger ToBigInteger(this ByteString text);
```
erikzhang, line 211: Can this be an explicit conversion?

shargon: Then we can't use null checks like this

```csharp
            var Data2 = Storage.CurrentContext.CreateMap("data");
            return Data2.Get(msg)?.ToBigInteger() ?? 0;
```
should be like this:

```csharp
            var Data2 = Storage.CurrentContext.CreateMap("data");
            return (BigInteger)(Data2.Get(msg) ?? (ByteString)"0");
```

More casts, I think

erikzhang:
```c#
var Data2 = Storage.CurrentContext.CreateMap("data");
return (BigInteger)Data2.Get(msg);
```

src/Neo.SmartContract.Framework/Helper.cs
Tommo-L: Done

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
Tommo-L: Since we can use `Contract.Call` direclty, can we remove `delegate` now?

https://github.com/neo-project/neo-devpack-dotnet/blob/ba73a45fe9eb012cf19248acde0e6a8f0f98bae9/src/Neo.SmartContract.Framework/Helper.cs#L188-L189

Tommo-L: > @Tommo-L could you add a UT for hit this lines, I was not able to compile a valid call with Delegates

Plz see https://github.com/neo-project/neo-devpack-dotnet/pull/414/commits/7b9a64ffb16d92829f528e18f25bbe64492d8d8f, but ut failed.

src/Neo.SmartContract.Framework/Helper.cs
erikzhang: I mean we don't need `[NonemitWithConvert(ConvertMethod.ToBigInteger)]`. We can support it in compiler. Just convert `BigInteger.Parse()` to the corresponding logic.

src/Neo.SmartContract.Framework/Helper.cs

src/Neo.SmartContract.Framework/Helper.cs

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
shargon: Thanks!

shargon:
Ready to review

src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
```diff
@@ -191,28 +191,41 @@ public static bool Parse(ILMethod from, NeoModule to)
...
+                                        calcStack.Push(new System.Numerics.BigInteger(p).ToByteArray());
+                                    }
+                                    else
+                                    {
+                                        var p = (int)calcStack.Pop();
+                                        calcStack.Push(new System.Numerics.BigInteger(p).ToByteArray());
+                                    }
                                 }
-                                else
+                                else if (m.Name == "Parse")
```
erikzhang, line 220: Should we check the different overloads of `Parse`?

shargon: Added with two arguments

src/Neo.Compiler.MSIL/MSIL/CctorSubVM.cs
```diff
@@ -219,13 +220,43 @@ public static bool Parse(ILMethod from, NeoModule to)
...
+                                    switch (calcStack.Count)
                                     {
-                                        p = System.Numerics.BigInteger.Parse(pstr);
-                                    }
+                                        case 1:
+                                            {
+                                                var p = calcStack.Pop();
+                                                if (p is string pstr)
+                                                {
+                                                    p = System.Numerics.BigInteger.Parse(pstr);
```
erikzhang, line 230: Now we have the SYSCALL `Atoi`.

shargon: Done

erikzhang:
@neo-project/ngd-shanghai Test.

cloud8little:
> @neo-project/ngd-shanghai Test.

Under testing.

cloud8little:
Deploy a contract like below, it seems when return value is void, it will throw exception.

Contract:
```cs
       public static string test2()
        {
            return "test2";
        }

        public static void _deploy(bool update)
        {
            Storage.Put(Storage.CurrentContext,"this","is");
        }

        public static void test3(int a)
        {
            a = a + 2;
        }

        public static int testInt(int a)
        {
            return 1;
        }

        public static void testvoidinputstring(string a)
        {
            a = a + "abc";
        }
```

6de75c5dd544f302d4e66bada3fdd25

shargon:
@cloud8little this doesn't happen before this PR?

cloud8little:
> @cloud8little this doesn't happen before this PR? or happen without `_deploy`? Maybe it is related to [neo-project/neo-modules#458](https://github.com/neo-project/neo-modules/issues/458) ?

at least we can deploy nep17 contract successfully which also have _deploy by default. I tried without _deploy in the contract. same error.

76ea79c6fb2ffa5adb196261a18d48d

superboyiii:
The same from my side. `_deploy` works well, but void can't retrun without error.
image
image

erikzhang:
The compiler didn't emit the opcode `INITSSLOT` or `INITSLOT`.

cloud8little:
@shargon The reason is parameters can not be changed in the function. It is not the return void error. For experiment you can change the test contract a bit as following: 

```cs
        public static void Testvoid(int a, int b)
        {
            a = a + 2;
        }
```

I guess it is a legacy issue. Let's discuss/fix in other PRs.

shargon:
@cloud8little please open a new issue, we can fix it in other PR

erikzhang (approved):

PR #425, opened 2021-01-14
sync FindOptions

Qiao-Jin:


erikzhang:
`FindOptions.All` is for check only. It's not a valid value.

PR #433, opened 2021-01-15
Fix BigInteger

### LINKED ISSUE #431: BigInteger.Parse Issue
Use BigInteger.Parse does not work.

f8fa6bce1fe692a28af893696dee24e

55244b3ab4c875c83e2ea96acae3f45

### LINKED ISSUE #432: BigInteger.Zero cannot compile successfully
```c#
public static readonly string mapName = "asset";

public static BigInteger Get(UInt160 key)
{
    var value = Storage.CurrentContext.CreateMap(mapName).Get(key);
    return value.Length > 0 ? (BigInteger)value : BigInteger.Zero;
}
```

```
严重性	代码	说明	项目	文件	行	禁止显示状态
错误		System.Exception: error:System.Numerics.BigInteger PEG.AssetStorage::Get(Neo.UInt160)::IL_002A Call System.Numerics.BigInteger System.Numerics.BigInteger::get_Zero()	PEG	C:\Users\chenzhitong\source\repos\NeoContract7\EXEC	1	

```

shargon:
Fix #432 fix #431

### DIFF
--- a/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
+++ b/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
@@ -764,6 +764,21 @@ namespace Neo.Compiler.MSIL
                     Insert1(VM.OpCode.SYSCALL, "", to, BitConverter.GetBytes(ApplicationEngine.System_Binary_Atoi));
                     return 0;
                 }
+                else if (src.tokenMethod == "System.Numerics.BigInteger System.Numerics.BigInteger::get_One()")
+                {
+                    ConvertPushNumber(1, null, to);
+                    return 0;
+                }
+                else if (src.tokenMethod == "System.Numerics.BigInteger System.Numerics.BigInteger::get_MinusOne()")
+                {
+                    ConvertPushNumber(-1, null, to);
+                    return 0;
+                }
+                else if (src.tokenMethod == "System.Numerics.BigInteger System.Numerics.BigInteger::get_Zero()")
+                {
+                    ConvertPushNumber(0, null, to);
+                    return 0;
+                }
             }
 
             if (calltype == 0)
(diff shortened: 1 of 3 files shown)


Tommo-L (approved):

chenzhitong (approved):

chenzhitong:
Fixed https://github.com/neo-project/neo-devpack-dotnet/issues/432

cloud8little:
@shargon you can add the UT in #434

erikzhang (approved):

shargon:
@cloud8little I think that it works but not for hex string, for hex string we should use `Atoi`, it's hard to convert System.String,System.Globalization.NumberStyles into our base argument
```
else if (src.tokenMethod == "System.Numerics.BigInteger System.Numerics.BigInteger::Parse(System.String,System.Globalization.NumberStyles)")
                {
                    ConvertPushNumber(10, null, to);        // Push Base
                    Convert1by1(VM.OpCode.SWAP, src, to);   // Swap arguments
                    Insert1(VM.OpCode.SYSCALL, "", to, BitConverter.GetBytes(ApplicationEngine.System_Binary_Atoi));
                    return 0;
                }
```

cloud8little:
@shargon  I know that this is implementation problem. but still it is very confusing, because we have 
1.  (BigInteger)xx
2. BigInteger.Parse()
3.  xx.ToBigInteger()

I find that ToBigInteger also work well for hex byte array.

```cs
        public static BigInteger testtoBigInteger()
        {
            return ((new byte[] { 0x00, 0x00, 0x8d, 0x49, 0xfd, 0x1a, 0x07 })).ToBigInteger();
        }
```

shargon:
@cloud8little I agree, please open a new issue in order to find a more user friendly way

PR #444, opened 2021-01-27
Update FindOptions.cs

chenzhitong:
sync neo https://github.com/neo-project/neo/pull/2231/files


chenzhitong:
https://github.com/neo-project/neo/blob/677bd9648d40b31076ba78c88c8a48bc0e919f66/src/neo/SmartContract/ApplicationEngine.Storage.cs#L62-L76

It seems that using FindOptions.All causes an ArgumentException.

PR #446, opened 2021-01-28
Sync with neo and add default values

chenzhitong:


shargon:
If we will add default values, we should review if we can add more of them

erikzhang:
BTW, the NFT contract should be updated too. (Added new method `tokens()`.)

shargon (approved):

src/Neo.SmartContract.Framework/Services/Neo/ContractManagement.cs
```diff
@@ -7,8 +7,8 @@ public class ContractManagement
     {
         public static extern UInt160 Hash { [ContractHash] get; }
         public static extern Contract GetContract(UInt160 hash);
-        public static extern Contract Deploy(ByteString nefFile, string manifest, object data);
-        public static extern void Update(ByteString nefFile, string manifest, object data);
+        public static extern Contract Deploy(ByteString nefFile, string manifest, object data = null);
+        public static extern void Update(ByteString nefFile, string manifest, object data = null);
```
erikzhang, line 11: They have overloads.

shargon: Done

erikzhang (dismissed):

shargon:
Require https://github.com/neo-project/neo/pull/2277

vncoelho (approved):

templates/Template.CSharp/Contract1.cs
```diff
@@ -3,12 +3,14 @@
...
 using System;
 using System.Numerics;
+using System.ComponentModel;
 
 namespace $safeprojectname$
 {
     [ManifestExtra("Author", "Neo")]
     [ManifestExtra("Email", "dev@neo.org")]
     [ManifestExtra("Description", "This is a contract example")]
+    [DisplayName("Name")]
```
erikzhang, line 13: Why do we have to add a `DisplayName`? We can use the class name as the contract name.

shargon: Yes, By default we can do it, I will change it

shargon: Done

shargon (approved):

src/Neo.Compiler.MSIL/FuncExport.cs
```diff
@@ -202,7 +201,7 @@ public static string GenerateManifest(JObject abi, NeoModule module)
             var name = module.attributes
```
erikzhang, line 201: I think we should use the class name because the attribute is applied on the class.

shargon: Got it. It will use now the class name if there are only one, otherwise, it will use the filename.

tests/Neo.Compiler.MSIL.UnitTests/TestClasses/Contract1.cs
```diff
@@ -1,5 +1,8 @@
+using System.ComponentModel;
+
 namespace Neo.Compiler.MSIL.UnitTests.TestClasses
-{
+{
+    [DisplayName("Name")]
```
erikzhang, line 5: Now we can revert these changes?

shargon: Done

shargon (approved):

vncoelho (approved):

src/Neo.Compiler.MSIL/FuncExport.cs
erikzhang: Why do we allow multiple contract classes in a project? If so, we should generate multiple `.nef` file.

Tommo-L:
Merge?

Issue #447, opened 2021-01-29
return type for Iterator is IInteropInterface 


bettybao1209:
![image](https://user-images.githubusercontent.com/12050350/106255499-10e33500-6255-11eb-819d-712b8fe87021.png)

Since the return type `Iterator ` in nameservice is compiled into `IInteropInterface ` rather than `InteropInterface`, so it fails to deploy the contract.

shargon:
Coud you send the manifest?

shargon:
Please check if https://github.com/neo-project/neo-devpack-dotnet/pull/448 fix this issue

bettybao1209:
> Please check if #448 fix this issue

It works for this case.

bettybao1209:
Also, it needs to sync https://github.com/neo-project/neo/pull/2271

shargon:
> Also, it needs to sync neo-project/neo#2271

We have this https://github.com/neo-project/neo-devpack-dotnet/pull/446

PR #448, opened 2021-01-29
Fix IInteropInterface

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/447

### DIFF
--- a/src/Neo.Compiler.MSIL/FuncExport.cs
+++ b/src/Neo.Compiler.MSIL/FuncExport.cs
@@ -5,7 +5,6 @@ using Neo.IO.Json;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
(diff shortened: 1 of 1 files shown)


src/Neo.Compiler.MSIL/FuncExport.cs
```diff
@@ -27,7 +26,7 @@ internal static string ConvType(TypeReference t)
                 {
                     if (i.InterfaceType.Name == nameof(IApiInterface))
```
erikzhang, line 27: What is `IApiInterface`?

shargon: image

shargon: Maybe we can use any, and remove IInteropInterface as valid return

erikzhang (approved):

PR #457, opened 2021-02-03
Adding name to HelloWorld Example

vncoelho:


templates/Template.CSharp/Contract1.cs
```diff
@@ -3,9 +3,11 @@
 using Neo.SmartContract.Framework.Services.Neo;
 using System;
 using System.Numerics;
+using System.ComponentModel;
 
 namespace $safeprojectname$
 {
+    [DisplayName("Hello World Demo")]
```
shargon, line 10: Just "Hello World"?

vncoelho: Could be, I followed an oracle example I saw from @Tommo-L.

shargon (approved):

templates/Template.CSharp/Contract1.cs
Tommo-L: We moved oracle demo there now. https://github.com/neo-project/examples/pull/34/files

Tommo-L (approved):

erikzhang:
This is not a demo. This is a **project template**. I will revert this commit.

Issue #459, opened 2021-02-04
We need a way to declare the manifest permissions


erikzhang:
All `NEP-17` contracts need the following permission.

```
{
  "contract": "*",
  "methods": "onNEP17Payment"
}
```

PR #461, opened 2021-02-04
Add ContractPermissionAttributes

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/459


src/Neo.SmartContract.Framework/WildcardContractPermissionAttribute.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Neo.SmartContract.Framework
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class WildcardContractPermissionAttribute : Attribute
```
erikzhang, line 6: We don't need this. We can use `[ContractPermission("*", methods)]`.

src/Neo.SmartContract.Framework/ContractPermissionAttribute.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+
+namespace Neo.SmartContract.Framework
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    public class ContractPermissionAttribute : Attribute
+    {
+        public string HashOrGroup { get; set; }
```
erikzhang, line 8:
```suggestion
        public string Contract { get; set; }
```

templates/Template.CSharp/Contract1.cs
```diff
@@ -9,6 +9,7 @@ namespace $safeprojectname$
     [ManifestExtra("Author", "Neo")]
     [ManifestExtra("Email", "dev@neo.org")]
     [ManifestExtra("Description", "This is a contract example")]
+    [WildcardContractPermission("onNEP17Payment")]
```
erikzhang, line 12: This is not a `NEP-17` contract. It doesn't need this permission.

src/Neo.SmartContract.Framework/WildcardContractPermissionAttribute.cs
shargon: Done

templates/Template.CSharp/Contract1.cs
shargon: Done

erikzhang (approved):

PR #474, opened 2021-03-04
Added some conversions

shargon:
Related to https://github.com/neo-project/neo-vm/issues/400#issuecomment-789502601


src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -672,6 +672,7 @@ private int ConvertCall(OpCode src, NeoMethod to, List<MethodToken> methodTokens
                 {
                     ConvertPushNumber(1, src, to);
                     Convert1by1(VM.OpCode.SUBSTR, null, to);
+                    Insert1(VM.OpCode.CONVERT, "", to, new byte[] { (byte)VM.Types.StackItemType.ByteString });
```
erikzhang, line 675: `char` should be mapped to `Integer`.

See #290

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -162,24 +162,40 @@ public static byte ToByte(this int source)
         [OpCode(OpCode.CAT)]
         public extern static byte[] Concat(this byte[] first, ByteString second);
 
+        [OpCode(OpCode.CAT)]
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
+        public extern static ByteString Concat(this ByteString first, ByteString second);
```
erikzhang, line 167: I think we only need to support `string + string`. All others should be explicitly converted by users.

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
shargon: Done

src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs
```diff
@@ -672,6 +672,7 @@ private int ConvertCall(OpCode src, NeoMethod to, List<MethodToken> methodTokens
                 {
                     ConvertPushNumber(1, src, to);
                     Convert1by1(VM.OpCode.SUBSTR, null, to);
+                    Insert1(VM.OpCode.CONVERT, "", to, new byte[] { (byte)VM.Types.StackItemType.Integer });
```
erikzhang, line 675: We can use a single opcode `PICKITEM`.

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -162,24 +162,40 @@ public static byte ToByte(this int source)
         [OpCode(OpCode.CAT)]
         public extern static byte[] Concat(this byte[] first, ByteString second);
 
+        [OpCode(OpCode.CAT)]
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
+        public extern static string Concat(this string first, string second);
```
erikzhang, line 167: We have already supported this feature in the compiler.

https://github.com/neo-project/neo-devpack-dotnet/blob/84bd5e762fa3c183c5dba377a136f266edd60faa/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs#L616-L663

shargon: Done

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -168,18 +168,30 @@ public static byte ToByte(this int source)
         [OpCode(OpCode.SUBSTR)]
         public extern static byte[] Range(this byte[] source, int index, int count);
 
+        [OpCode(OpCode.SUBSTR)]
+        [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
+        public extern static string Range(this string source, int index, int count);
```
erikzhang, line 173: Use `string.SubString()`?

src/Neo.SmartContract.Framework/Helper.cs
```diff
@@ -174,12 +174,22 @@ public static byte ToByte(this int source)
         [OpCode(OpCode.LEFT)]
         public extern static byte[] Take(this byte[] source, int count);
```
erikzhang, line 175: Why this overload doesn't need `SWAP`?

shargon: I am not sure :S if otherwise it wont work, I will check it...

shargon: In this block it's produced the discrepancy 
https://github.com/neo-project/neo-devpack-dotnet/blob/84bd5e762fa3c183c5dba377a136f266edd60faa/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs#L422-L429

Then it's added the swap
https://github.com/neo-project/neo-devpack-dotnet/blob/84bd5e762fa3c183c5dba377a136f266edd60faa/src/Neo.Compiler.MSIL/MSIL/Conv_Multi.cs#L813

shargon: Maybe we can add another attribute to define which kind of logic we want, instead of use the count of the opcodes?

erikzhang: Maybe we can use `CallingConvention`?

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.callingconvention?view=net-5.0

shargon: Please, take a look

src/Neo.SmartContract.Framework/CallingConversionAttribute.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Neo.SmartContract.Framework
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class CallingConversionAttribute : Attribute
+    {
+        public CallingConvention CallingConversion { get; private set; }
```
erikzhang, line 9:
```c#
public bool RightToLeft { get; }
```
?

src/Neo.SmartContract.Framework/CallingConversionAttribute.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Neo.SmartContract.Framework
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class CallingConversionAttribute : Attribute
```
erikzhang, line 7: `RightToLeftAttribute`?

Issue #475, opened 2021-03-11
[Suggestion] automatically add permissions?


chenzhitong:
If permission is not added, an error will occur when the contract is upgraded and destroyed. 
This is because it calls the `ContractManagement` native contract, but does not declare permissions.

Solutions:

1. The compiler automatically adds permissions when it detects code that calls a native contract.
2. Add ContractManagement to the permissions by default.
2. Do nothing, Let the developer add it manually.
3. Let the developer add it manually. And permissions supports adding the contract name of the native contract, because there is no guarantee that the hash of the native contract will not change (it has already changed n times).

Which is better?

shargon:
I think that option 2 it's good

PR #478, opened 2021-03-11
Generate manifest with static calls

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/475


superboyiii:
@shargon I'll test this later but I think it's better not be merged to RC1 since not much time left.

erikzhang:
@superboyiii It should be added to RC1.

superboyiii:
I've tested this. It works but I'm confused if it makes sense in this mechanism.
1. In previous, unnative contracts are allowed to be called by all contracts as default. And if add 1 permission, it becomes only allowed this permission method to call.
2. In this PR, unnative contracts are not allowed to be called by all contracts as default. That means if we'd like to call any method of other SC, we have to upgrade our SC again and again. In most DAPP, modules are added step by step not at the same time. In this way, I think most developers will add [ContractPermission("*", "*")] at the beginning and upgrade ContractPermission later. So is it useful we set forbidden all as the default?
@shargon @erikzhang

shargon:
We can add a compiler flag `--allowAllCalls`?

erikzhang:
```c#
[ContractPermission("*")]
```

You need to manually add this line to declare the calling permissions of all contracts.

erikzhang:
Tested?

superboyiii:
> Tested?

I will test it soon.

superboyiii:
@shargon I set `[ContractPermission("*", "*")]` but `ToDelegate()` still return fault because of authorization. Does this mean if natvie management contract is allowed in permission than all permission must add manually? "*" seems doesn't work.

```
{
    "jsonrpc": "2.0",
    "id": 3,
    "result": {
        "script": "ABQMFI5SMl22gwrQirvz2CzRnQyvhFHfDBSabQINJN1EW/uACj2hZ5UvHKCtDwwUj0sLioGSzZgeFe060WuXVzteXjgUwB8MBXRlc3QyDBTX2B1RYQz8amWTYixFhXS9jndqD0FifVtS",
        "state": "FAULT",
        "gasconsumed": "3471120",
        "exception": "An unhandled exception was thrown. No authorization.",
        "stack": []
    }
}
```

erikzhang:
> [ContractPermission("*", "*")]

I think it not `[ContractPermission("*", "*")]` but `[ContractPermission("*")]`.

superboyiii:
> > [ContractPermission("*", "*")]
> 
> I think it's not `[ContractPermission("*", "*")]` but `[ContractPermission("*")]`.
I've tried both, `[ContractPermission("*", "*")] ` and `[ContractPermission("*")]` does no difference in either nef hash or manifest json result. And I saw my permission and native permission exist both, which should be the final result?

image

erikzhang:
> which should be the final result?

The first one.

But I think we should optimize for it. @shargon

superboyiii:
Sorry, I turned witnessScopes from `CalledByEntry` to `Global`, everything works well.
So in current mechanism, default still is `allowAll` and whatever your permission is, `contrtactmanagement` can always be invoked in `getContract`, `destroy`, `update`.

erikzhang:
You still need the permissions. Otherwise it should fail.

superboyiii:
> You still need the permissions. Otherwise it should fail.

If one permission is manually added, then will be `denyAll` except `contractmanagerment`. But if you don't set any permission, it's the same effect as `[ContractPermission("*", "*")]`.

superboyiii:
Test **PASS**
It works well on mine.

superboyiii (dismissed):

erikzhang:
@shargon Can you optimize the permission list?

shargon:
> Can you optimize the permission list?

Yes, I will

superboyiii:
> > Can you optimize the permission list?
> 
> Yes, I will

If optimize it, we need do it quickly since I'll release CLI today and set up Testnet on 18th. It should be merged before Testnet.

shargon:
@superboyiii please test it

superboyiii:
> @superboyiii please test it

It works on mine.

superboyiii (dismissed):

superboyiii:
@erikzhang Need your review.

erikzhang:
@shargon Please use the `PermissionBuilder` added in da3b922.

shargon:
@superboyiii please test it

src/Neo.Compiler.MSIL/PermissionBuilder.cs
```diff
@@ -44,7 +49,7 @@ public void Add(string hash, string method)
         public JArray ToJson()
         {
             JArray permissions = new();
-            if (isWildcard)
+            if (isWildcard || normalItems.Count == 0)
```
erikzhang, line 52: Why? I think `*, *` should be explicitly declared.

shargon: Because otherwise the manifest will be invalid, isn't it?

erikzhang:
```
"permissions": []
```

It's valid.

superboyiii:
@shargon Double `[]`?
**
image
**

erikzhang (approved):

superboyiii:
Test **PASS**
Works well on mine.

erikzhang:
Please fix the UT.

superboyiii:
@shargon I just push a PR on your repo to fix this, could you review it? https://github.com/neo-project/neo-modules/pull/564

shargon:
@superboyiii merged!

superboyiii:
@erikzhang Merge?

erikzhang (approved):

erikzhang:
It's better to test it again before merging.

superboyiii:
> It's better to test it again before merging.

Give me half an hour.

superboyiii:
> It's better to test it again before merging.

Test **PASS**
Works well on mine.

superboyiii (approved):

erikzhang (approved):

PR #490, opened 2021-04-01
add try catch and map ut

cloud8little:
also added a Map<> use UInt160 as key type, think we should add support for Map<UInt160, int>, because there is request for it, for example, vote contracts need to record vote for each account.


erikzhang:
@cloud8little Did you check your code?

erikzhang:
You need to check your code before push.

erikzhang (approved):

cloud8little:
> You need to check your code before push.

Got it.

PR #491, opened 2021-04-01
Optimize StorageMap

erikzhang:


src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -1,8 +1,131 @@
...
+        [OpCode(OpCode.OVER)]
+        [OpCode(OpCode.PUSH1)]
+        [OpCode(OpCode.PICKITEM)]
+        [OpCode(OpCode.SWAP)]
+        [OpCode(OpCode.CAT)]
+        [OpCode(OpCode.SWAP)]
+        [OpCode(OpCode.PUSH0)]
+        [OpCode(OpCode.PICKITEM)]
+        [Syscall("System.Storage.Get")]
+        public extern ByteString Get(ByteString key);
```
Qiao-Jin, line 60: Duplicate with ByteString this[ByteString key].get?

erikzhang: Yes. They are the same.

cloud8little:
testing.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -1,8 +1,131 @@
...
+
+        [CallingConvention(CallingConvention.Cdecl)]
+        [OpCode(OpCode.DUP)]
+        [OpCode(OpCode.PUSH1)]
+        [OpCode(OpCode.PICKITEM)]
+        [OpCode(OpCode.SWAP)]
+        [OpCode(OpCode.PUSH0)]
+        [OpCode(OpCode.PICKITEM)]
+        [Syscall("System.Storage.Find")]
+        public extern Iterator Find(FindOptions options = FindOptions.None);
```
Qiao-Jin, line 70: Somehow strange here

erikzhang:
> testing.

@cloud8little You need to test it with #492.

cloud8little:
does it support byte[] type as the key for storageMap? I think it should work as the same with Storage?  do I miss something

cloud8little:
Create a contract use storagemap, and put a value of UInt160 type, but it return nothing after Put. the contract is as below. 

```cs
        public static void CreateBid(ulong _biddingTime, UInt160 _beneficiary)
        {
            ulong currenttime = Runtime.Time;
            byte[] time = (byte[])(ByteString)(StdLib.Itoa(currenttime));
            ByteString isended = store.Get((ByteString)prefix_isended);
            //ByteString isended = Storage.Get(sc, prefix_isended);
            if (isended == null)
            {
                store.Put((ByteString)prefix_isended, "0");
            }
            else if (isended == "1")
            {
                throw new Exception("Auction not ended yet");
            }
            else
            {
                store.Put((ByteString)prefix_isended, "0");
            }
            store.Put((ByteString)prefix_endtime, currenttime + _biddingTime);
            store.Put((ByteString)prefix_benef, _beneficiary);
        }

        public static ByteString GetBenefier()
        {
            return store.Get((ByteString)prefix_benef);
        }

        public static BigInteger GetEndTime()
        {
            return (BigInteger)store.Get((ByteString)prefix_endtime);
        }
```

06669d47d76ee4524ce5e0ae49a8d3b

cloud8little:
bidamount is uint type, it shows compile error when use store[xx] = xx, because it can not be implicitly converted to ByteString type. while BigInteger is widely used, it is better to support.

image

erikzhang:
> Support BigInteger as the value for storageMap when using index.

It is not allowed by C#.

erikzhang:
> does it support byte[] type as the key for storageMap?

Updated.

cloud8little:
> > does it support byte[] type as the key for storageMap?
> 
> Updated.

Test Passed, it is more user friendly for users. Still when comparing the way storagemap[] with storagemap.Get/Put, it is more simple and easy understood when using get/put dealing with BigInteger type value, because even there is a way for BigInteger explicit conversion, there still need to convert to ByteString again. 

01b6014715ea2c4d8f3276118999ae3

bdc71709287773004827bdac14013ab

cloud8little (approved):

PR #492, opened 2021-04-01
Support this property

erikzhang:


cloud8little:
testing.

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -2376,16 +2446,31 @@ private void ConvertElementAccessPreIncrementOrDecrementExpression(SemanticModel
...
+                AddInstruction(OpCode.DUP);
+                AddInstruction(OpCode.REVERSE4);
+                Call(model, property.SetMethod!, CallingConvention.Cdecl);
+            }
+            else
+            {
+                ConvertExpression(model, operand.Expression);
+                ConvertExpression(model, operand.ArgumentList.Arguments[0].Expression);
+                AddInstruction(OpCode.OVER);
+                AddInstruction(OpCode.OVER);
```
shargon, line 2466: Move these lines outside the condition?

erikzhang: They are not shared parts, they just happen to be the same.

cloud8little (approved):

cloud8little:
Test passed with #491

PR #497, opened 2021-04-02
Fix string add

erikzhang:


shargon (approved):

cloud8little:
@erikzhang  does it mean the following style is allowed?

```cs
ByteString a = "a";
ByteString b = "b";
return a + b;
```

erikzhang:
Yes. It should return "ab".

PR #504, opened 2021-04-03
Array support foreach

### LINKED ISSUE #501: Support array foreach
```cs
        public static int IntForeach()
        {
            int[] a = new int[] { 1, 2, 3, 4 };
            int sum = 0;
            foreach (var item in a)
            {
                sum += item;
            }
            return sum;
        }
```

erikzhang:
Close #501

### DIFF
--- a/src/Neo.SmartContract.Framework/UInt160.cs
+++ b/src/Neo.SmartContract.Framework/UInt160.cs
@@ -2,7 +2,7 @@ using Neo.SmartContract.Framework;
 
 namespace Neo
 {
-    public class UInt160
+    public class UInt160 : ByteString
     {
         public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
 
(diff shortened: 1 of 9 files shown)


src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Neo
 {
-    public class UInt160
+    public class UInt160 : ByteString
```
shargon, line 5: It won't check the size a null

erikzhang: https://github.com/neo-project/neo-devpack-dotnet/blob/281731934238bcba4e608572106b46524c8a89ab/src/Neo.Compiler.CSharp/MethodConvert.cs#L1963-L1976

It checks the size here.

cloud8little:
Test all passed with string/ByteString/struct/bytearray/BigInteger/ECPoint/UInt160/UInt256/List/objec array.

cloud8little (approved):

PR #505, opened 2021-04-03
Fix BigInteger explicit cast

erikzhang:


src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -3195,8 +3195,8 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                             PrepareArgumentsForMethod(model, symbol, arguments);
                         JumpTarget endTarget = new();
                         AddInstruction(OpCode.DUP);
-                        Push(-128);
-                        Push(128);
+                        Push(sbyte.MinValue);
+                        Push(sbyte.MaxValue);
```
erikzhang, line 3199: The values are different. We must revert this commit.

shargon (dismissed):

src/Neo.Compiler.CSharp/MethodConvert.cs
shargon: But it should be the same, isn't it?

erikzhang: https://github.com/neo-project/neo-vm/blob/59b8ac73d285aa9d607aee53eeadde82fcba9324/src/neo-vm/ExecutionEngine.cs#L958-L965

erikzhang: The `WITHIN` opcode checks the range `[a, b)`.

shargon: Please check it again

erikzhang: Good.

shargon (approved):

PR #511, opened 2021-04-04
Allow public instance methods

erikzhang:


shargon:
I think instance methods can create confusion with how ethereum works. They preserves instance states for future executions automatically (variables), and we don't (require to call storage), I don't know if allowing instance methods will cause developer confusion.

erikzhang:
We need instance methods to support Inheritance. For example:

```c#
public class TokenContract : SmartContract
{
    public abstract string Symbol { get; }
    public abstract byte Decimals { get; }
}

public sealed class Nep17Token : TokenContract
{
    public override string Symbol => "MyToken";
    public override byte Decimals => 9;
}
```

shargon:
Maybe we can deny instance variables in order to avoid confusion?

erikzhang:
How about we call the constructor automatically?

erikzhang:
> How about we call the constructor automatically?

Done.

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -140,6 +140,29 @@ public void Convert(SemanticModel model)
...
+            {
+                for (int i = fields.Length - 1; i >= 0; i--)
+                    PushDefault(fields[i].Type);
+                Push(fields.Length);
+                AddInstruction(OpCode.PACK);
+            }
+            IMethodSymbol? constructor = type.InstanceConstructors.FirstOrDefault(p => p.Parameters.Length == 0);
+            if (constructor is null)
+                throw new CompilationException(type, DiagnosticId.NoParameterlessConstructor, "The contract class requires a parameterless constructor.");
+            Call(model, constructor, true, Array.Empty<ArgumentSyntax>());
```
shargon, line 161: The constructor will be called only once if we call twice the instance method?

erikzhang: The forward method will call the constructor automatically. The real instance methods are not exposed.

shargon (approved):

PR #515, opened 2021-04-06
Convert all methods in the source code

erikzhang:


shargon (commented):
What methods were not previously exported in order to do a test?

erikzhang:
> What methods were not previously exported in order to do a test?

A private method that never be invoked.

shargon (approved):

PR #518, opened 2021-04-06
Add Oracle.verify

shargon:
It's easily to check that the sender it's the oracle response.


erikzhang:
What's the use case?

shargon:
> What's the use case?

https://github.com/neo-project/neo/pull/2426

PR #523, opened 2021-04-07
Support polymorphism

erikzhang:


src/Neo.Compiler.CSharp/Options.cs
```diff
@@ -3,6 +3,7 @@ namespace Neo.Compiler
     public class Options
     {
         public string? Output { get; set; }
+        public bool Polymorphism { get; set; }
```
shargon, line 6: Should this be optional?

erikzhang: Yes. It is default to `false`.

erikzhang: I've removed the option. Now the compiler will detect whether to enable polymorphism automatically.

erikzhang:
Ready to review.

shargon:
@erikzhang with my last commit, the virtual doesnt' work, could you check it?

https://github.com/neo-project/neo-devpack-dotnet/pull/523/commits/4e6fce5fe0c924345c083a7be0707c83203bfe71

erikzhang:
@shargon Fixed!

shargon (approved):

PR #535, opened 2021-04-09
Support string indexer

erikzhang:


src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -3988,6 +3988,13 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                         PrepareArgumentsForMethod(model, symbol, arguments);
                     AddInstruction(OpCode.EQUAL);
                     return true;
+                case "string.this[int].get":
```
shargon, line 3991: In my test I get

image

erikzhang: The ut is incorrect. This pr is to support string indexer:

```c#
string s = "abc";
char c = s[1]; // 'b'
```

shargon (approved):

PR #536, opened 2021-04-09
Fix the bug of cast

erikzhang:


shargon:
Any example to test ?

erikzhang:
I think it's a bug of Roslyn.

You can test it with https://github.com/erikzhang/neo-name-service

shargon (approved):

PR #544, opened 2021-04-10
Read source files from csproj

### LINKED ISSUE #540: nccs doesn't pick up CS file from parent folder specified in .csproj file
I'm trying to compile a project that references a .CS file from a parent folder. normal C# compilation succeeds but NCCS fails with a [CS0234 error "type or namespace does not exist"](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0234) for any types defined in the file in the parent folder

Example csproj file contents:

``` xml

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="../additional-code.cs" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="{path to local Neo.SmartContract.Framework.csproj}" />
  </ItemGroup>

</Project>
```

erikzhang:
Close #540


erikzhang:
@devhawk Please check it.

devhawk (commented):
I don't think this is the right approach. We don't want to try and replicate the logic of msbuild in nccs.

csc - the main C# compiler - takes a list of .cs files to compile on the command line. It's MSBuild's job to decide what .cs files to pass to the csc compiler. Taking a similar MSBuild based approach w/ nccs will ensure full compatibility w/ existing tools while also making nccs simpler

erikzhang:
@devhawk To use `MSBuildWorkspace`, the users need to install visual studio first. I don't think that the compiler must work with visual studio.

devhawk:
> @devhawk To use `MSBuildWorkspace`, the users need to install visual studio first. I don't think that the compiler must work with visual studio.

Sorry, I wasn't  clear. I'm suggesting *removing* the logic from nccs to parse project files. Instead, nccs should just accept a list of one or more .cs files on the command line. We can leverage existing msbuild logic to calculate the list of files, then feed that list to nccs via the existing Exec msbuild task. Eventually we could build an NccsTask  that is similar to the existing CscTask.

I believe this approach will be simpler to implement in nccs and better integrate into the existing dotnet build infrastructure.

erikzhang:
Not only the source files, we need the references too.

devhawk:
> Not only the source files, we need the references too.

MSBuild will manage the list of references too - including the inclusive set of references to references. Those are passed to csc via the /reference command line option. 

If a contract project references an assembly that references Neo.SmartContract.Framework, what will nccs do?

erikzhang:
> MSBuild will manage the list of references too

Maybe we can add this mode later.

> If a contract project references an assembly that references Neo.SmartContract.Framework, what will nccs do?

That's what #546 want to do. Read the source code of the Framework as a compilation reference.

erikzhang:
We need to merge this PR as soon as possible.

devhawk:
> > MSBuild will manage the list of references too
> 
> Maybe we can add this mode later.
> 
> > If a contract project references an assembly that references Neo.SmartContract.Framework, what will nccs do?
> 
> That's what #546 want to do. Read the source code of the Framework as a compilation reference.

I will open issues to track both of these

PR #546, opened 2021-04-10
Include source files of Neo.SmartContract.Framework in nuget package

### LINKED ISSUE #547: MSBUILD : error MSB1011: Specify which project or solution file to use because this folder contains more than one project or solution file
When the solution file and the project file are in the same folder and have different names, the following error will be reported
```
MSBUILD : error MSB1011: Specify which project or solution file to use because this folder contains more than one project or solution file
```

image

Change their file names to the same name to compile successfully.

image

erikzhang:
Fix #547


erikzhang:
Wait for #544

devhawk (commented):
Since all the source is on github, we can alternatively use SourceLink instead of including all the source in the package

https://github.com/dotnet/sourcelink/blob/main/README.md

erikzhang:
The problem is, once a csharp project has been compiled into a dll, we can't convert it again to neo opcodes. If we don't include the source code in the nuget package, we can't use the `Nep11Token` base class.

erikzhang:
This pr is not completed. I'm waiting for #544.

src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -16,63 +16,46 @@ static int Main(string[] args)
...
-            path = Path.GetFullPath(path);
+            if (path is null)
+                path = Environment.CurrentDirectory;
+            else
+                path = Path.GetFullPath(path);
             if (File.Exists(path))
             {
-                return Path.GetExtension(path).ToLowerInvariant() switch
-                {
-                    ".cs" => ProcessSources(options, Path.GetDirectoryName(path)!, path),
```
shargon, line 38: We will lose the ability of compile a single source file?

erikzhang: Because we need the source code of Neo.SmartContract.Framework. So we have to use nuget. That means we need a csproj.

shargon: Maybe we can add a new argument for the nuget version

erikzhang: Please take a look again.

shargon (approved):

src/Neo.Compiler.CSharp/Program.cs
shargon: It works :)

PR #556, opened 2021-04-13
Add notary

shargon:


erikzhang:
We've not yet decided to use this solution.

PR #569, opened 2021-04-15
Sync to Neo 3.0.0-CI01266

erikzhang:


src/Neo.SmartContract.Framework/Nep11Token.cs
```diff
@@ -145,7 +145,7 @@ private static void UpdateBalance(UInt160 owner, ByteString tokenId, int increme
                 balanceMap.Put(owner, balance);
             ByteString key = owner + tokenId;
             if (increment > 0)
-                accountMap[key] = tokenId;
+                accountMap.Put(key, 0);
```
shargon, line 148: 1?
```suggestion
                accountMap.Put(key, 1);
```

erikzhang: No difference. The value is never used.

shargon (approved):

Issue #560, opened 2021-04-14
Return help information when using nccs without any argument or option.


nicolegys:
When I enter "nccs" without any argument or option at the terminal, it will work on the current path by default, and throw exception. 
![image](https://user-images.githubusercontent.com/43694095/114665377-89737100-9d2f-11eb-8d52-732f21f4112f.png)

I think we should return help information in this case, just like `dotnet`
![image](https://user-images.githubusercontent.com/43694095/114665660-e2dba000-9d2f-11eb-81c8-f9e15eff263d.png)

erikzhang:
If no parameter is provided, it will compile the `.csproj` file in the current directory. You can use `--help` to show the help information.

cloud8little:
> If no parameter is provided, it will compile the `.csproj` file in the current directory. You can use `--help` to show the help information.

for better experience, it should show help info in this case, because lots of new-learners will just input the command nccs to know how it works. and if it accidently run the command in a parent folder which contains lots of csproj file, it is a disaster.

chenzhitong:
For new developers, after configuring the environment variable path, type `nccs` on the command line in any location to check if the configuration is correct. It is very unfriendly to report errors directly at this point.

ProDog:
In addition, some log information is needed when compiling, and this is the only output now:
```
PS D:\TestCode\Neo3Contract\Neo3Contract> dotnet D:\neo_code\neo-devpack-dotnet\src\Neo.Compiler.CSharp\bin\Debug\net5.0\nccs.dll .\Neo3Contract.csproj
  正在确定要还原的项目…
  所有项目均是最新的，无法还原。
PS D:\TestCode\Neo3Contract\Neo3Contract>
```
No way to know whether really successful compilation.

shargon:
@ProDog please check https://github.com/neo-project/neo-devpack-dotnet/pull/562/commits/cf12556d8dbc0a3e77b57fbe86d60bcd66ee342a

ProDog:
It's OK:
```
PS D:\TestCode\Neo3Contract\Neo3Contract> dotnet D:\neo_code\neo-devpack-dotnet\src\Neo.Compiler.CSharp\bin\Debug\net5.0\nccs.dll .\Neo3Contract.csproj
  正在确定要还原的项目…
  所有项目均是最新的，无法还原。
[...]

PR #578, opened 2021-04-16
Display help without args

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/560


erikzhang:
It will show help info even if it's a compilation error.

erikzhang (approved):

PR #579, opened 2021-04-17
Add generic methods for StorageMap

### LINKED ISSUE #84: Automatically call "Serialize" in Put method (add method overloading to Serialize to support objects)
Hi,
For a better experience, I think our Storage.Put(key, value) should automatically call .Serialize() if we send an object.
At least the version I have here, it only accepts byte[], meaning we have to call Serialize manually.

### LINKED ISSUE #530: TypeParameters on StorageMap
Is it possible with NCCS to have a typed storage map? Something like this

``` csharp
class MyContract : SmartContract
{
    StorageMap<UInt160, BigInteger> accountMap = new();

    StorageMap<UInt160, BigInteger> Accounts => accountMap;

    public TransferAll(UInt160 from, UInt160 to)
    {
        var total = accountMap.Get(from) + accountMap.Get(to);
        accountMap.Put(to, total);
        accountMap.Put(from, 0);
    }
}
```

Note, if we have/can add indexer support a la #529, then StorageMap could support indexer format as an alias for Get/Put

``` csharp
class MyContract : SmartContract
{
    StorageMap<UInt160, BigInteger> accountMap = new();
    StorageMap<UInt160, BigInteger> Accounts => accountMap;

    public TransferAll(UInt160 from, UInt160 to)
    {
        accountMap[to] = accountMap[from] + accountMap[to];
        accountMap[from] = 0;
    }
}
```

erikzhang:
Close #84, close #530

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -299,7 +299,7 @@ namespace Neo.Compiler
                     ["sequence-points"] = m.Instructions.Where(p => p.SourceLocation is not null).Select(p =>
                     {
                         FileLinePositionSpan span = p.SourceLocation!.GetLineSpan();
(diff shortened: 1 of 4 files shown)


shargon:
My UT doesn't work, @erikzhang please could you check it?

shargon (approved):

Issue #581, opened 2021-04-19
How to add a Trust field to a contract


RichardBelsum:
![image](https://user-images.githubusercontent.com/82148280/115199300-a50ebb00-a125-11eb-83fb-ae107b5d34dd.png)

Why are ContractPermissions added via Attributes, and ContractTrusts can't do that?

PR #582, opened 2021-04-19
Add ContractTrust

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/581


erikzhang:
`trusts` can be an array or `*`. If it is an array, the elements should be the hash of a contract, or the public key of a group.

shargon:
We can add a new attribute for each hash/pubKey, but I will update the code for an array

erikzhang:
I mean we should consider `*`, and verify the elements of the array.

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -277,7 +278,7 @@ public JObject CreateManifest()
                     }).ToArray()
                 },
                 ["permissions"] = permissions.ToJson(),
-                ["trusts"] = new JArray(),
+                ["trusts"] = trusts.OrderBy(p => p).Select(u => new JString(u)).ToArray(),
```
erikzhang, line 281: It can be an array or `*`. `*` is not allowed to be an element.

shargon: Done

erikzhang (approved):

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -356,6 +366,12 @@ private void ProcessClass(SemanticModel model, INamedTypeSymbol symbol)
                         case nameof(scfx.Neo.SmartContract.Framework.ContractPermissionAttribute):
                             permissions.Add((string)attribute.ConstructorArguments[0].Value!, attribute.ConstructorArguments[1].Values.Select(p => (string)p.Value!).ToArray());
                             break;
+                        case nameof(scfx.Neo.SmartContract.Framework.ContractTrustAttribute):
+                            string trust = (string)attribute.ConstructorArguments[0].Value!;
+                            if (!ValidateContractTrust(trust))
+                                throw new ArgumentException($"The value {trust} is not a valid one for ContractTrust");
```
RichardBelsum, line 372: Hold on, Trusts should only be used for wallet warnings and should not be mandatory.

Whether the contract can be called or not, should depend only on the Permissions and WitnessScope, not Trust.

roman-khimov: Reminds me of neo-project/neo#1664.

chenzhitong: > If a contract is trusted, the user interface will not give any warnings when called by the contract.
https://github.com/neo-project/neo/blob/7e2d435aada96e7e22c7564984bff1419377c33f/src/neo/SmartContract/Manifest/ContractManifest.cs#L51

Changed again?

erikzhang: I think it's a format check.

PR #588, opened 2021-04-26
Add System.Math

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/587

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert.cs
@@ -4109,7 +4109,7 @@ namespace Neo.Compiler
                         PrepareArgumentsForMethod(model, symbol, arguments);
                     Call(NativeContract.StdLib.Hash, "atoi", 1, true);
                     return true;
-                case "System.Math.Sign(byte)":
+                case "System.Math.Sign(sbyte)":
                 case "System.Math.Sign(short)":
                 case "System.Math.Sign(int)":
                 case "System.Math.Sign(long)":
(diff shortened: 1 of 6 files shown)


src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -4109,6 +4109,44 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                         PrepareArgumentsForMethod(model, symbol, arguments);
                     Call(NativeContract.StdLib.Hash, "atoi", 1, true);
                     return true;
+                case "System.Math.Pow(double, double)":
```
erikzhang, line 4112: We don't have `double` in contracts.

shargon: We can use a casted integer

erikzhang: We should use `BigInteger.Pow`.

shargon: Done

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -4109,6 +4109,38 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                         PrepareArgumentsForMethod(model, symbol, arguments);
                     Call(NativeContract.StdLib.Hash, "atoi", 1, true);
                     return true;
+                case "System.Math.Sign(byte)":
```
nicolegys, line 4112: shouldn't it be `sbyte`?

shargon: Fixed

nicolegys (approved):
Test passed.

erikzhang (approved):

PR #597, opened 2021-05-03
Fix path in linux

### LINKED ISSUE #596: nccs unable to locate nuget files due to filepath case sensitivity in Linux
The introduction of #546 causes the compiler to utilize some `files` included in our nuget packages directory. The filepath is concatenated through several strings. This causes issues on Linux as the filesystem is case sensitive, resulting in the runtime unable to find the correct files.

Reproduction using a Ubuntu 20.04 env in WSL:
```
git clone git@github.com:neo-project/neo-devpack-dotnet.git
cd neo-devpack-dotnet
dotnet publish -c Release src/Neo.Compiler.CSharp -o ./bin
./bin/nccs ./templates/Template.CSharp -o ./build
```

Expected:
Some compilation errors that is related to the contract compilation.

Got:
```
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.IO.DirectoryNotFoundException: Could not find a part of the path '/home/snowypowers/.nuget/packages/Neo.SmartContract.Framework/3.0.0-rc2/src/ByteString.cs'.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
...
```

Notice that the path attempts to look up `.nuget/packages/Neo.SmartContract.Framework`. However, when I checked with `ls ~/.nuget/packages`, every folder in `~/.nuget` is lowercased. The files and folders in the next level retain their casing. Thus, the correct path should be `/home/snowypowers/.nuget/packages/neo.smartcontract.framework/3.0.0-rc2/src/ByteString.cs`.

Nuget's behavior for installing packages in lowercased looks to be consistent across both Windows and Ubuntu. This bug does not occur on Windows because the filesystem is not case sensitive and will resolve to the intended path.

Fix:

A quick workaround is `name.toLower()` at this line:

https://github.com/neo-project/neo-devpack-dotnet/blob/6bea8e949a6767d444d96ac903d32b8bdaa8de84/src/Neo.Compiler.CSharp/CompilationContext.cs#L192

With this fix, I am able to use the compiler to get my intended behavior.

erikzhang:
Fix #596

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -171,6 +171,7 @@ namespace Neo.Compiler
                 switch (assets["libraries"][name]["type"].GetString())
                 {
                     case "package":
+                        string namePath = assets["libraries"][name]["path"].GetString();
                         string[] files = assets["libraries"][name]["files"].GetArray()
                             .Select(p => p.GetString())
                             .Where(p => p.StartsWith("src/"))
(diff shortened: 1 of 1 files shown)


snowypowers (approved):
I tested this in my linux env and it works.

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -182,15 +183,15 @@ public static Compilation GetCompilation(string csproj)
...
+                                string path = Path.Combine(packagesPath, namePath, file);
                                 if (!File.Exists(path)) continue;
                                 references.Add(MetadataReference.CreateFromFile(path));
                             }
                         }
                         else
                         {
-                            IEnumerable<SyntaxTree> st = files.OrderBy(p => p).Select(p => Path.Combine(packagesPath, name, p)).Select(p => CSharpSyntaxTree.ParseText(File.ReadAllText(p), path: p));
-                            CSharpCompilation cr = CSharpCompilation.Create(Path.GetDirectoryName(name), st, commonReferences, options);
+                            IEnumerable<SyntaxTree> st = files.OrderBy(p => p).Select(p => Path.Combine(packagesPath, namePath, p)).Select(p => CSharpSyntaxTree.ParseText(File.ReadAllText(p), path: p));
```
shargon, line 193: `namePath` it's relative to `packagesPath`?

erikzhang: Yes.

shargon (approved):

PR #607, opened 2021-05-05
Use AssemblyName

### LINKED ISSUE #602: Nccs needs option to specify the .nef filename
`nccs` uses the [contract name](https://github.com/neo-project/neo-devpack-dotnet/blob/master/src/Neo.Compiler.CSharp/Program.cs#L111) as the filename for the generated .nef/.manifest.json/.nefdbgnfo files. There is no way for the build system to know the contract name without parsing the contents of the contract source. 

`nccs` should have an option for specifying the base file name used for generated files explicitly.

erikzhang:
Close #602

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -40,7 +40,7 @@ namespace Neo.Compiler
 
         public bool Success => diagnostics.All(p => p.Severity != DiagnosticSeverity.Error);
         public IReadOnlyList<Diagnostic> Diagnostics => diagnostics;
-        public string ContractName { get; private set; } = "";
+        public string? ContractName { get; private set; }
         internal Options Options { get; private set; }
         internal IEnumerable<IFieldSymbol> StaticFieldSymbols => staticFields.OrderBy(p => p.Value).Select(p => p.Key);
         internal IEnumerable<(byte, ITypeSymbol)> VTables => vtables.OrderBy(p => p.Value).Select(p => (p.Value, p.Key));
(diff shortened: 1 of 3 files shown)


src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -156,6 +156,7 @@ public static Compilation GetCompilation(string csproj)
             List<MetadataReference> references = new(commonReferences);
             CSharpCompilationOptions options = new(OutputKind.DynamicallyLinkedLibrary);
             XDocument xml = XDocument.Load(csproj);
+            assemblyName = xml.Root!.Elements("PropertyGroup").Elements("AssemblyName").Select(p => p.Value).SingleOrDefault() ?? Path.GetFileNameWithoutExtension(csproj);
```
devhawk, line 159: This approach only works if a .csproj file is specified on the command line.

If we add an --assemblyName command line argument instead, it would work with  single file + directory compilation scenarios as well.

shargon: with single file it should take the class name, isn't it?

shargon (dismissed):

src/Neo.Compiler.CSharp/CompilationContext.cs
erikzhang: Agree with @shargon

superboyiii (dismissed):
It works well for me. Assembly > csproj > class name.

devhawk:
The issue that #602 raises is the need for build systems to know what the name of a tool's output will *without* having to know the contents of the inputs. 

I'm fine with `nccs`  to using the contract name as the base name of the generated .nef/.manifest.json/.nefdbgnfo files by default. #602 is proposing the addition of a mechanism so the build system can override that default. 

Adding an `--assembly-name` (or other similarly named) command line option enables the build system to override the default base name, regardless if nccs is provided a single .cs file, a .csproj file or a directory of .cs files.

Parsing the AssemblyName property out of the .csproj file is more complex and doesn't provide a mechanism to control output base name for the single .cs file or directory of .cs files solution.

devhawk (approved):

shargon (approved):

PR #612, opened 2021-05-07
add static-variables to debug info 

### LINKED ISSUE #611: add static variables to debug info emitted by nccs
In order to support static variables in the debugger, we need to add a top level `static-variables` array to the debug json emitted by the compiler. This addition is documented in [v1.2 (draft) of the debug info design note](https://github.com/ngdenterprise/design-notes/blob/master/NDX-DN11%20-%20NEO%20Debug%20Info%20Specification.md#v12-format-draft)

This is a one line change in `nccs` `CompilationContext.CreateDebugInformation` method 
``` csharp
["static-variables"] = StaticFieldSymbols.Select(p => (JString)$"{p.Name},{p.Type.GetContractParameterType()}").ToArray(),
```

devhawk:
fix #611


shargon (dismissed):

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -300,6 +300,7 @@ public JObject CreateDebugInformation()
             {
                 ["hash"] = Script.ToScriptHash().ToString(),
                 ["documents"] = compilation.SyntaxTrees.Select(p => (JString)p.FilePath).ToArray(),
+                ["static-variables"] = StaticFieldSymbols.Select(p => (JString)$"{p.Name},{p.Type.GetContractParameterType()}").ToArray(),
```
erikzhang, line 303: Do you need the slot indexes of the variables?

devhawk: No, but the variables in the array need to match the layout in the slot i.e. static variable slot 0 needs to match the 1st item in the `static-variables` debug info array. This is the same for params and local variable arrays as well.

devhawk: Is the `StaticFieldSymbols` collection in the same order as the slot fields?

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -303,6 +303,7 @@ public JObject CreateDebugInformation()
...
```
shargon, line 306:
```suggestion
                ["static-variables"] = staticFields.Select(p => (JString)$"{p.Value},{p.Key.Name},{p.Key.Type.GetContractParameterType()}").ToArray(),
```

src/Neo.Compiler.CSharp/CompilationContext.cs
devhawk: where are those tracked?

src/Neo.Compiler.CSharp/CompilationContext.cs
devhawk: Also, it looks like StaticFieldSymbols is the correct choice here, since it's ordered

src/Neo.Compiler.CSharp/CompilationContext.cs
devhawk: It appears that static slots are tracked by staticFields and vtables dictionaries - with the expectation that byte value in both staticFields and vtables dictonaries is the sslot index

src/Neo.Compiler.CSharp/CompilationContext.cs
shargon: it's for include the slot index

shargon (dismissed):

devhawk:
Thanks for adding the UTs @shargon !

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -299,10 +299,16 @@ public JObject CreateManifest()
         public JObject CreateDebugInformation()
         {
             SyntaxTree[] trees = compilation.SyntaxTrees.ToArray();
+            var staticVars = staticFields
+                .Select(p => (index: p.Value, name: p.Key.Name, type: p.Key.Type.GetContractParameterType()))
+                .Concat(vtables.Select(p => (index: p.Value, name: p.Key.Name, type: ContractParameterType.Any)))
```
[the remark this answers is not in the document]
devhawk: I'm unclear what you are suggesting @erikzhang. Can you clarify what you want me to rename here?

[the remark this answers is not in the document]
erikzhang: Currently we are using the type name as the variable name.

shargon (approved):

erikzhang:
https://github.com/neo-project/neo-devpack-dotnet/pull/612/files#diff-27038bb15e8236627041207a01a839e1d9fe48fe6c7055371fd2cffc837d4df5R321

`variables` are not arranged according to the index of the slot too. Because different variables may share the same index.

For example:

```c#
void Test(bool flag)
{
    if (flag)
    {
        int i1 = 1;
        //Do something
    }
    else
    {
        int i2 = 2;
        //Else
    }
}
```

Both the variables `i1` and `i2` use the slot index `0`.

devhawk:
> `variables` are not arranged according to the index of the slot too. Because different variables may share the same index.

I'll open another issue for this

erikzhang (approved):

shargon (approved):

PR #615, opened 2021-05-08
Add dotnet new template

### LINKED ISSUE #609: dotnet new templates should be in this repo
there are neo contract templates that work with `dotnet new` in the https://github.com/neo-ngd/Neo3.SmartContract.Templates repo. We should keep the various templates together in one repo.

Also, VS2019 16.8 and later supports `dotnet new` templates directly. We should consider eliminating the VS2019 templates in favor of `dotnet new` templates that work with both VS2019 and command line developers alike

erikzhang:
Close #609


devhawk (approved):

shargon (approved):

MissAlleyCat:
commit

PR #620, opened 2021-05-14
Add slot index to debug info

### LINKED ISSUE #619: nccs non-optimized builds need to emit separate variable slots for each local variable
@devhawk 

https://github.com/neo-project/neo-devpack-dotnet/pull/612/files#diff-27038bb15e8236627041207a01a839e1d9fe48fe6c7055371fd2cffc837d4df5R321

`variables` are not arranged according to the index of the slot too. Because different variables may share the same index.

For example:

```c#
void Test(bool flag)
{
    if (flag)
    {
        int i1 = 1;
        //Do something
    }
    else
    {
        int i2 = 2;
        //Else
    }
}
```

Both the variables `i1` and `i2` use the slot index `0`.

_Originally posted by @erikzhang in https://github.com/neo-project/neo-devpack-dotnet/issues/612#issuecomment-840324453_

erikzhang:
Close #619

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -310,7 +310,7 @@ namespace Neo.Compiler
                     ["name"] = $"{m.Symbol.ContainingType},{m.Symbol.Name}",
                     ["range"] = $"{m.Instructions[0].Offset}-{m.Instructions[^1].Offset}",
                     ["params"] = (m.Symbol.IsStatic ? Array.Empty<JString>() : new JString[] { "this,Any" })
-                        .Concat(m.Symbol.Parameters.Select(p => (JString)$"{p.Name},{p.Type.GetContractParameterType()}"))
+                        .Concat(m.Symbol.Parameters.Select((p, i) => (JString)$"{p.Name},{p.Type.GetContractParameterType()},{i}"))
                         .ToArray(),
                     ["return"] = m.Symbol.ReturnType.GetContractParameterType().ToString(),
                     ["variables"] = m.Variables.Select(p => (JString)$"{p.Symbol.Name},{p.Symbol.Type.GetContractParameterType()},{p.SlotIndex}").ToArray(),
@@ -324,7 +324,7 @@ namespace Neo.Compiler
                 {
                     ["id"] = e.Name,
                     ["name"] = $"{e.Symbol.ContainingType},{e.Symbol.Name}",
-                    ["params"] = e.Parameters.Select(p => (JString)$"{p.Name},{p.Type}").ToArray()
+                    ["params"] = e.Parameters.Select((p, i) => (JString)$"{p.Name},{p.Type},{i}").ToArray()
                 }).ToArray()
             };
         }
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -299,16 +299,11 @@ namespace Neo.Compiler
         public JObject CreateDebugInformation()
         {
             SyntaxTree[] trees = compilation.SyntaxTrees.ToArray();
-            var staticVars = staticFields
-                .Select(p => (index: p.Value, name: p.Key.Name, type: p.Key.Type.GetContractParameterType()))
-                .Concat(vtables.Select(p => (index: p.Value, name: $"<{p.Key.Name}>__vtable", type: ContractParameterType.Any)))
-                .OrderBy(p => p.index);
-
             return new JObject
             {
                 ["hash"] = Script.ToScriptHash().ToString(),
                 ["documents"] = compilation.SyntaxTrees.Select(p => (JString)p.FilePath).ToArray(),
-                ["static-variables"] = staticVars.Select(p => (JString)$"{p.name},{p.type}").ToArray(),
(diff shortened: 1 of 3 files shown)


src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -299,16 +299,11 @@ public JObject CreateManifest()
...
-                .Select(p => (index: p.Value, name: p.Key.Name, type: p.Key.Type.GetContractParameterType()))
-                .Concat(vtables.Select(p => (index: p.Value, name: $"<{p.Key.Name}>__vtable", type: ContractParameterType.Any)))
-                .OrderBy(p => p.index);
-
             return new JObject
             {
                 ["hash"] = Script.ToScriptHash().ToString(),
                 ["documents"] = compilation.SyntaxTrees.Select(p => (JString)p.FilePath).ToArray(),
-                ["static-variables"] = staticVars.Select(p => (JString)$"{p.name},{p.type}").ToArray(),
+                ["static-variables"] = staticFields.OrderBy(p => p.Value).Select(p => (JString)$"{p.Key.Name},{p.Key.Type.GetContractParameterType()},{p.Value}").ToArray(),
```
shargon, line 306: Not needed to concat with `vtables`?

erikzhang: I don't think exposing vtables to the debugger is any help.

shargon (approved):

src/Neo.Compiler.CSharp/CompilationContext.cs
devhawk: Agreed, we only needed vtables info to ensure implicit debug info variable index was correct.

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -318,7 +313,7 @@ public JObject CreateDebugInformation()
                         .Concat(m.Symbol.Parameters.Select(p => (JString)$"{p.Name},{p.Type.GetContractParameterType()}"))
                         .ToArray(),
                     ["return"] = m.Symbol.ReturnType.GetContractParameterType().ToString(),
-                    ["variables"] = m.Variables.Select(p => (JString)$"{p.Name},{p.Type.GetContractParameterType()}").ToArray(),
+                    ["variables"] = m.Variables.Select(p => (JString)$"{p.Symbol.Name},{p.Symbol.Type.GetContractParameterType()},{p.SlotIndex}").ToArray(),
```
devhawk, line 316: Should we include explicit slot index in params array too?

devhawk:
Is there going to be an other RC2 release of `nccs` for the hackathon. Just need to know the timeframe I have for updating the debugger.

superboyiii:
> Is there going to be an other RC2 release of `nccs` for the hackathon? Just need to know the timeframe I have for updating the debugger.

Hackthon will use current RC2 testnet. But for `nccs`, since it doesn't change neo-core, I think could release another special version for this PR if it's necessary. Need your opinion. @erikzhang @roman-khimov

src/Neo.Compiler.CSharp/CompilationContext.cs
erikzhang: The `params` are always arranged by index. Do you think it is necessary to add indexes to the JSON explicitly?

roman-khimov:
I think it's better for us not to change data formats for RC2/hackathon tools, it's not just C# devpack, but all other compilers currently emit this data without slots, so changing anything here and in debugger we're risking breaking something and I wouldn't want that to happen during hackathon. So to me it's something for RC3.

src/Neo.Compiler.CSharp/CompilationContext.cs
roman-khimov: It'd probably be more convenient to have the same data format everywhere.

devhawk:
> I think it's better for us not to change data formats for RC2/hackathon tools, it's not just C# devpack, but all other compilers currently emit this data without slots, so changing anything here and in debugger we're risking breaking something and I wouldn't want that to happen during hackathon. So to me it's something for RC3.

I agree with this. I updated the debug info parser so that it wouldn't throw exceptions if it encounters a type with the slot info, but the debugger doesn't look at that info yet. 

Note @roman-khimov, I wrote the slot info up as optional and when I do consume this data in the debugger, it will default to using the implicit index if the explicit index isn't set. So the neo-go contract compiler won't need to be updated if it doesn't need the ability to specify slots explicitly.

devhawk:
> I think it's better for us not to change data formats for RC2/hackathon tools, it's not just C# devpack, but all other compilers currently emit this data without slots, so changing anything here and in debugger we're risking breaking something and I wouldn't want that to happen during hackathon. So to me it's something for RC3.

I agree with this. I updated the debug info parser so that it wouldn't throw exceptions if it encounters a type with the slot info, but the debugger doesn't look at that info yet. 

Note @roman-khimov, I wrote the slot info up as optional and when I do consume this data in the debugger, it will default to using the implicit index if the explicit index isn't set. So the neo-go contract compiler  doesn't need to be updated if it doesn't need the ability to specify slots explicitly.

src/Neo.Compiler.CSharp/CompilationContext.cs
erikzhang: Done.

devhawk (approved):
I can't test this thoroughly until I update the debugger, but it looks right

PR #626, opened 2021-05-18
Fix StorageMap.Put for UInt160 and UInt256

### LINKED ISSUE #624: Poor developer experience with StorageMap.Get<T> and Put<T>
It is very easy to mismatch `StorageMap` `Get` and `Put` methods. In the code below, the call to `owners.Put` actually calls `owners.Put<UInt160>`, which automatically calls `StdLib.Serialize` on the provided `UInt160 owner` parameter. However, since the type parameter for `StorageMap.Get<T>` method can't be determined implicitly, the call to `owners.Get` below actually calls the non-generic method which in turn does NOT call `StdLib.Deserialize`. This leads to a hard-to-diagnose runtime error when the serialized `UInt160` value is cast w/o calling Deserialize.

Example code:

``` csharp
readonly StorageMap owners = new StorageMap(Storage.CurrentContext, nameof(owners));

public UInt160 GetOwner(BigInteger id)
{
    ByteString owner = owners.Get(id); // calls non-generic Get method, which doesn't call StdLib.Deserialize
    return (UInt160)owner ?? UInt160.Zero; // this cast fails at runtime if the owner is not null
}

public void PutOwner(BigInteger id, UInt160 owner)
{
    owners.Put(id, owner); // implicitly calls StorageMap.Put<UInt160> which calls StdLib.Serialize on the provided parameter
}
```

erikzhang:
Fix #624

### DIFF
--- a/src/Neo.SmartContract.Framework/Services/StorageMap.cs
+++ b/src/Neo.SmartContract.Framework/Services/StorageMap.cs
@@ -105,14 +105,14 @@ namespace Neo.SmartContract.Framework.Services
         [Syscall("System.Storage.Get")]
         public extern ByteString Get(byte[] key);
 
-        public T Get<T>(ByteString key) where T : class, new()
+        public object GetObject(ByteString key)
         {
-            return (T)StdLib.Deserialize(Get(key));
+            return StdLib.Deserialize(Get(key));
         }
 
-        public T Get<T>(byte[] key) where T : class, new()
+        public object GetObject(byte[] key)
         {
-            return (T)StdLib.Deserialize(Get(key));
+            return StdLib.Deserialize(Get(key));
         }
 
         [CallingConvention(CallingConvention.Cdecl)]
@@ -197,12 +197,12 @@ namespace Neo.SmartContract.Framework.Services
         [Syscall("System.Storage.Put")]
         public extern void Put(byte[] key, BigInteger value);
 
-        public void Put<T>(ByteString key, T value) where T : class, new()
+        public void PutObject(ByteString key, object value)
         {
             Put(key, StdLib.Serialize(value));
         }
 
-        public void Put<T>(byte[] key, T value) where T : class, new()
+        public void PutObject(byte[] key, object value)
         {
             Put(key, StdLib.Serialize(value));
         }
--- a/src/Neo.SmartContract.Framework/UInt160.cs
+++ b/src/Neo.SmartContract.Framework/UInt160.cs
@@ -2,7 +2,7 @@ using Neo.SmartContract.Framework;
 
 namespace Neo
 {
-    public class UInt160 : ByteString
+    public abstract class UInt160 : ByteString
     {
         public static extern UInt160 Zero { [OpCode(OpCode.PUSHDATA1, "140000000000000000000000000000000000000000")] get; }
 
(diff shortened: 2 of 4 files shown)


shargon (approved):
Code looks good, please @devhawk confirm that the issue was solved

devhawk:
This solves #624 for UInt160/256, but it doesn't solve the general problem. The core issue is that Put<T> is called *implicitly* while Get<T> has to be called *explicitly*. Imagine the same scenario with a user defined struct:

``` csharp

public class SaleInfo
{
    public ByteString Id;
    public UInt160 Seller;
    public UInt160 Buyer;
    public string Description;
    public BigInteger Price;
}

public class SafePurchaseContract : SmartContract
{
    readonly StorageMap salesMap = new StorageMap(Storage.CurrentContext, "saleInfo");

    public void CreateSale(ByteString saleId, BigInteger price, string description)
    {
        
        var tx = (Transaction)Runtime.ScriptContainer;
        var saleInfo = new SaleInfo()
        {
            Id = saleId,
            Seller = tx.Sender,
            Description = description,
            Price = price,
        };

        salesMap.Put(saleId, saleInfo); // Put<SaleInfo> implicitly called here
    }

    public string GetSale(ByteString saleId)
    {
        var sale = salesMap.Get<SaleInfo>(saleId); // Get<SaleInfo> must be explicitly called here
        return StdLib.JsonSerialize(sale);
    }
}
```

erikzhang:
@devhawk Do you have any idea? Rename the method to `PutSerialized`?

devhawk:
> @devhawk Do you have any idea? Rename the method to `PutSerialized`?

Yes, I think rename `Put<T>` is the best solution. I like `Serialize` a little better than `PutSerialized` because it's only one word and it's shorter to type. But `PutSerialized` would show up directly next `Put` in the Intellisense list. Frankly, I don't have strong feelings about the new name as long as it's different.

We should also change the deserializing version of `Get<T>` to match - i.e. either `Deserialize` or `GetDeserialized`.

erikzhang:
@devhawk Please review it again.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -105,14 +105,14 @@ public class StorageMap
         [Syscall("System.Storage.Get")]
         public extern ByteString Get(byte[] key);
 
-        public T Get<T>(ByteString key) where T : class, new()
+        public object GetObject(ByteString key)
```
devhawk, line 108: Do we need separate overloads for ByteString and byte[]?

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Neo
 {
-    public class UInt160 : ByteString
+    public abstract class UInt160 : ByteString
```
devhawk, line 5: Do we still need to mark these as abstract?

src/Neo.SmartContract.Framework/UInt256.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Neo
 {
-    public class UInt256 : ByteString
+    public abstract class UInt256 : ByteString
```
devhawk, line 5: DO we need to mark this as abstract?

devhawk (approved):
Minor nitpicks, otherwise looks good

PR #627, opened 2021-05-19
Fix sequence points

### LINKED ISSUE #604: nccs generating invalid sequence points
Test Contract: https://gist.github.com/devhawk/6d9ba02ae26a7f7618f83087c631fc41
Compiled with `dotnet nccs --debug --no-optimize TestContract.csproj`

In the generated debug info, we get sequence points that have an invalid document index of -1

> Note, this may not be a valid contract. As per #603, nccs fails to compile this contract when compiling the .cs file directly instead of compiling the contract via its .csproj file.

``` json
{
    "id": "Neo.SmartContract.Framework.Services.StorageMap.Put\u003CNeo.UInt160\u003E(Neo.SmartContract.Framework.ByteString, Neo.UInt160)",
    "name": "Neo.SmartContract.Framework.Services.StorageMap,Put",
    "range": "268-292",
    "params": [
        "key,ByteArray",
        "value,Hash160"
    ],
    "return": "Void",
    "variables": [],
    "sequence-points": [
        "271[-1]201:9-201:10",
        "272[-1]202:13-202:47",
        "292[-1]203:9-203:10"
    ]
}
```

erikzhang:
Fix #604

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -298,7 +298,7 @@ namespace Neo.Compiler
 
         public JObject CreateDebugInformation()
         {
-            SyntaxTree[] trees = compilation.SyntaxTrees.ToArray();
+            string[] sourceLocations = GetSourceLocations(compilation).Distinct().ToArray();
             var staticVars = staticFields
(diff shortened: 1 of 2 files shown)


shargon (dismissed):

devhawk:
Have we tested this with a contract that uses a method from SmartContract Framework that is implemented in code (such as StorageMap.Put<T>)? Assuming that works, I'm good.

devhawk (dismissed):

Issue #622, opened 2021-05-17
Multiple partial cs file Contract Compile error


cloud8little:
Due to the change in https://github.com/neo-project/neo-devpack-dotnet/pull/601 , contract with multiple cs files compiler error:

![dcfec89ca537e9e2c5ff3f5b6c427b7](https://user-images.githubusercontent.com/34291844/118456891-fd88a680-b72b-11eb-871f-08df43a5b06d.png)

because it runs into the logic twice:

```cs
      private void ProcessClass(SemanticModel model, INamedTypeSymbol symbol)
        {
            if (symbol.IsSubclassOf(nameof(Attribute))) return;
            bool isPublic = symbol.DeclaredAccessibility == Accessibility.Public;
            bool isAbstract = symbol.IsAbstract;
            bool isContractType = symbol.IsSubclassOf(nameof(scfx.Neo.SmartContract.Framework.SmartContract));
            bool isSmartContract = isPublic && !isAbstract && isContractType;
            if (isSmartContract)
            {
                if (scTypeFound) throw new CompilationException(DiagnosticId.MultiplyContracts, $"Only one smart contract is allowed.");
                scTypeFound = true;
               
            ....
           }
        }
```

[nNEO.zip](https://github.com/neo-project/neo-devpack-dotnet/files/6492302/nNEO.zip)

PR #629, opened 2021-05-20
Fix partial classes

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/622


erikzhang:
@shargon Please check #633

PR #631, opened 2021-05-21
Sync neo

shargon:
Related to https://github.com/neo-project/neo/pull/2468#issuecomment-845445461

### DIFF
--- a/src/Neo.SmartContract.Framework/Native/StdLib.cs
+++ b/src/Neo.SmartContract.Framework/Native/StdLib.cs
@@ -29,8 +29,6 @@ namespace Neo.SmartContract.Framework.Native
 
         public static extern ByteString Base58CheckDecode(string input);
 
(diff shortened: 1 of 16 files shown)


src/Neo.SmartContract.Framework/Native/StdLib.cs
```diff
@@ -25,6 +25,12 @@ public static class StdLib
 
         public static extern string Base58Encode(ByteString input);
 
+        public static extern string Base58CheckEncode(ByteString input);
+
+        public static extern ByteString Base58CheckDecode(string input);
+
+        public static extern ByteString Base58CheckDecode(ByteString input);
```
erikzhang, line 32: We don't need this. Because `ByteString` can be converted to `string` implicitly.

shargon: Done

erikzhang (approved):

PR #654, opened 2021-06-24
Change NEP11 visibility

### LINKED ISSUE #653: UpdateBalance and PostTransfer methods of Nep11Token class should be protected not private

### LINKED ISSUE #717: Rescope methods marked as "private protected" as "protected
There are several methods marked as "private protected in the `TokenContract.cs` module (e.g. - `UpdateTotalSupply`, etc.).  Please redeclare these functions as simply "protected".  Currently they can not be used outside the assembly they reside in.  The scope declarations as is makes proper integration and debugging of that contract code awkward and unfriendly.

shargon:
Close #653
Close #717


erikzhang:
Why?

shargon:
Maybe you want to create your own transfer logic, it follows the openzepellin pattern https://github.com/OpenZeppelin/openzeppelin-contracts/blob/eea376911b32d2c6392496d966e38601c1e762d5/contracts/token/ERC20/ERC20.sol#L223

superboyiii:
I think it make sense.

src/Neo.SmartContract.Framework/Nep11Token.cs
```diff
@@ -105,7 +105,7 @@ protected static void Burn(ByteString tokenId)
             PostTransfer(token.Owner, null, tokenId, null);
         }
 
-        private static void UpdateBalance(UInt160 owner, ByteString tokenId, int increment)
+        protected static void UpdateBalance(UInt160 owner, ByteString tokenId, int increment)
```
erikzhang, line 108: This method cannot be called alone, because `totalSupply` will not be updated.

shargon: This only happend during mint or burn, we can add an argument like `changeTotalSupply` inside this method or change `UpdateTotalSupply` to protected, what do you prefer?

devhawk (approved):

SkyKarma:
@shargon 

Is it possible to have this merged into master ? 
Is there anything stopping this ?

erikzhang (approved):

PR #656, opened 2021-06-24
Add Concat for ByteStrings

### LINKED ISSUE #655: How to concat byteStrings in a contract?
What is the right way to concatenate two ByteStrings into a single ByteString so that I can create a SHA256 hash from it? It doesn't appear byteStrings support the + operator directly. If you write `var combined = byteString1 + byteString2`, the resulting type is a string, since ByteString supports implicit converstion to string and string supports + operator. I know string is just a byteString under the hood - the implicit cast operator is a NOP - but it does seem odd that you can't concat byte strings directoy

shargon:
Close #655


devhawk:
I missed the existence of Neo.SmartContract.Framework.Helper.Concat. Should we add overloads of Concat that take a ByteString as the this param as well?

src/Neo.SmartContract.Framework/ByteString.cs
```diff
@@ -51,5 +51,8 @@ IEnumerator IEnumerable.GetEnumerator()
 
         [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
         public static extern explicit operator ByteString(BigInteger integer);
+
+        [OpCode(OpCode.CAT)]
+        public static extern ByteString operator +(ByteString a);
```
erikzhang, line 56: I think it should be `operator +(ByteString a, ByteString b)`

shargon: > I think it should be operator +(ByteString a, ByteString b)

The unit tests doesn't work if I use this way

erikzhang: But this is unary plus operator, not the addition operator.

See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/arithmetic-operators#unary-plus-and-minus-operators

shargon: If I try with 

```csharp
[OpCode(OpCode.CAT)]
public static extern ByteString operator +(ByteString a, ByteString y);
```
It try to do the sum

image

erikzhang (approved):

PR #660, opened 2021-07-02
Fix StorageMap.GetObject

### LINKED ISSUE #659: StorageMap.GetObject should return null if the key is missing
GetObject throws if the key is missing. It should return null instead

erikzhang:
Fix #659


src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -107,12 +107,16 @@ public class StorageMap
 
         public object GetObject(ByteString key)
```
devhawk, line 108: Code looks good, but why do we have to duplicate it? Can't byte[] implicitly convert to ByteString or vis versa?

devhawk (approved):

shargon (approved):

PR #671, opened 2021-08-20
copyright in files

Liaojinghui:


shargon (dismissed):

src/Neo.SmartContract.Template/templates/neocontract/Contract1.cs
```diff
@@ -1,3 +1,13 @@
+// Copyright (C) 2015-2021 The Neo Project.
```
erikzhang, line 1: This file is a part of the template. I think we can ignore this file.

shargon (approved):

erikzhang (approved):

PR #677, opened 2021-08-31
Add Modifiers

shargon:
Execute the Attributes code that inherits from the Modifier class.

```csharp
public class OwnerOnly : Modifier
    {
        UInt160 owner;

        public OwnerOnly(string hex)
        {
            owner = (UInt160)(byte[])StdLib.Base64Decode(hex);
        }

        public override void Validate()
        {
            if (!Runtime.CheckWitness(owner)) throw new System.Exception();
        }
    }

    public class Contract_Attribute : SmartContract
    {
        [OwnerOnly("AAAAAAAAAAAAAAAAAAAAAAAAAAA=")]
        public static bool test()
        {
            return true;
        }
    }
```


shargon:
@superboyiii could you test it?

Liaojinghui:
I vote on adding this pr.

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Attribute.cs
```diff
@@ -0,0 +1,23 @@
+using Neo.SmartContract.Framework;
+using Neo.SmartContract.Framework.Services;
+
+namespace Neo.SmartContract.Framework.UnitTests.TestClasses
+{
+    public class OnlyOwner : Modifier
+    {
+        public OnlyOwner()
```
Liaojinghui, line 8:
```suggestion
        public OwnerOnly()

```

superboyiii:
> I vote on adding this pr.

What's the difference from CheckOwner()? Why have to make it more complex by adding a new modifer class and can't add any param.

Liaojinghui:
> > I vote on adding this pr.
> 
> I've completed test but still confused. What's the difference from CheckOwner()? Why have to make it more complex by adding a new modifer class and can't add any param.

OwnerOnly() is a kind of permission control. though it works the same as `checkowner()`, OwnerOnly is more intuitive and easy to use. I thought this pr would directly support  `OwnerOnly()`. If u need another class to implement such a function, I agree, it becomes more complex

shargon:
OnlyOwner is only an example, this is almost the same as modifiers in solidity

superboyiii (approved):
I think it's OK for me since it's an addtional feature, no harm to the current.

Liaojinghui:
> OnlyOwner is only an example, this is almost the same as modifiers in solidity

I wont turn against it neither. My only argument is that can we use `OwnerOnly` instead. LOL, I don't know why, but `onlyowner` is kinda wired to me.

vncoelho (commented):
I vote for this feature as well.

@shargon, I changed to `OwnerOnly` as he suggested just for reading purpose. Fell free to change again.

erikzhang:
I think the constructor of `Modifier` should support parameters. And we can add an abstract parameterless `Validate` method to the class.

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -117,6 +117,13 @@ private SequencePointInserter InsertSequencePoint(SyntaxNodeOrToken syntax)
 
         public void Convert(SemanticModel model)
         {
+            foreach (var modifier in Symbol.GetAttributes().Where(u => u.AttributeClass?.BaseType?.Name == nameof(scfx::Neo.SmartContract.Framework.Modifier)))
+            {
+                if (modifier.AttributeConstructor == null) continue;
```
shargon, line 122: @erikzhang could you help me here to inject the new object and call the instance method?

shargon: @erikzhang any hint?

shargon:
@superboyiii could you test it?

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -117,6 +117,31 @@ private SequencePointInserter InsertSequencePoint(SyntaxNodeOrToken syntax)
...
+
+                // Send arguments
+                AddInstruction(OpCode.PUSHNULL);
+                AddInstruction(OpCode.PUSH1);
+                AddInstruction(OpCode.PACK);   // this
+                foreach (var arg in modifier.ConstructorArguments.Reverse()) Push(arg.Value);
+                Push(modifier.ConstructorArguments.Length);
+                AddInstruction(OpCode.PICK);        // DUP-this
+                // Call constructor
+                MethodConvert convert = context.ConvertMethod(model, modifier.AttributeConstructor);
```
erikzhang, line 138: It should be a static field and should be called only once.

shargon: But static can't be defined in the abstract class https://github.com/neo-project/neo-devpack-dotnet/pull/677/files#diff-1ad8aac41d6c456e6a06fb864d429a1a3baa4136e98cd553d78a77536c880460R7 

Do you want to remove the `Modifier` class and call `Validate` to all `Attributes` if it has the method?

If we use static, we can't define the arguments in the constructor like this https://github.com/neo-project/neo-devpack-dotnet/pull/677/files#diff-c20efd606cd201b0aad296807af579551e9ce54bf9cf637872c7770c01761336R23

This version emulte `new OwnerAttribute("AAAAAAAAAAAAAAAAAAAAAAAAAAA=").Validate();`

erikzhang: I mean the attribute instance should be static.

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Attribute.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public override void Validate()
+        {
+            if (!Runtime.CheckWitness(owner)) throw new System.Exception();
+        }
+    }
+
+    public class Contract_Attribute : SmartContract
+    {
+        [OwnerOnly("AAAAAAAAAAAAAAAAAAAAAAAAAAA=")]
```
erikzhang, line 23: I rethought it and found that this is not practical. Because you need to specify owner in attribute every time. In addition, owner is usually mutable, and the parameters of `Attribute` must be constant.

devhawk: I was going to make a similar argument. I thought the best practice was to store the contract owner in storage, which precludes hardcoding the value in an attribute.

``` cs
[DisplayName("_deploy")]
public void Deploy(object data, bool update)
{
    if (update) return;

    var tx = (Transaction)Runtime.ScriptContainer;
    var key = new byte[] { Prefix_ContractOwner };
    Storage.Put(Storage.CurrentContext, key, tx.Sender);
}
```

src/Neo.SmartContract.Framework/Attributes/Modifier.cs
```diff
@@ -0,0 +1,20 @@
...
+// 
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using System;
+
+namespace Neo.SmartContract.Framework.Attributes
+{
+    [AttributeUsage(AttributeTargets.Constructor | AttributeTargets.Method, AllowMultiple = true)]
+    public abstract class Modifier : Attribute
```
devhawk, line 16: should be named ModifierAttribute

devhawk:
Are there other examples of ModiferAttribute beyond OwnerOnly that we are expecting to see?

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Attribute.cs
shargon: We can use the storage key instead of the value

shargon:
> Are there other examples of ModiferAttribute beyond OwnerOnly that we are expecting to see?

OnlyRole, NoReentrancy for example, in solidity are a lot of them

erikzhang:
Merge?

erikzhang (approved):

shargon:
Good to merge for me

erikzhang:
@devhawk Do you agree to merge?

devhawk (approved):
I'm cool with the approach here, so long as we don't add the `OwnerOnlyAttribute` with the hard coded owner string that's used in the tests to the framework.

PR #694, opened 2021-11-12
fix compile error

Ashuaidehao:
Relate  https://github.com/neo-project/neo/pull/2613


erikzhang:
We are not going to use `CI*` packages anymore.

PR #704, opened 2021-11-25
Fix DisplayName

### LINKED ISSUE #674: DisplayName attribute of Class is not available
DisplayName attribute of Class is not available, by design?

```c#
namespace ContractA
{
    [DisplayName("ContractB")]
    public class ContractA : SmartContract
    {

    }
}
```

### LINKED ISSUE #696: Compiler ignores DisplayName for contract name
It ignores [DisplayName("Contract name")] attributes and make class name as contract name.

erikzhang:
Close #674, close #696


Liaojinghui:
@shargon  I would say that using `=` instead of `??=` is more proper here. And the reason that the result is same is because e `ContractName` is not set at all when it reaches
```C#
ContractName = (string)attribute.ConstructorArguments[0].Value!;
```

Liaojinghui (approved):

shargon (approved):

PR #706, opened 2021-11-30
Fix CompilationReference

erikzhang:


shargon:
Any way to test it? when it fails?

erikzhang:
@shargon I've invited you to galaxz's repo so that you can test this pr with the GCS contract.

shargon (approved):
Tested after and before and it works

PR #710, opened 2021-12-17
add ByteString.Empty

### LINKED ISSUE #709: ByteString should have static Empty value, like UInt160.Zero and string.Empty
When working in a nullable context, the C# compiler wants you to initialize values to a non-null value. but it's not clear how to initialize a ByteString in this context.

Example:

``` csharp
public class NftInfo
{
    public UInt160 TokenScriptHash = UInt160.Zero;
    public string TokenSymbol = string.Empty;
    public ByteString TokenId; // what goes here?
    public BigInteger Price;
}
```

Options (today) include:
* `default!` (i.e. disable null checking for this field)
* `string.Empty` (there's an implicit conversion from string to ByteString)
* `(ByteString)new byte[]` (there's an explicit conversion from byte array to ByteString)
  * Note, `(ByteString)Array.Empty<byte>()` doesn't work as NCCS doesn't know how to handle `Array.Empty`

Optimally, there should be a `ByteArray.Empty` static field for such scenarios

Qiao-Jin:
Close #709

### DIFF
--- a/src/Neo.SmartContract.Framework/ByteString.cs
+++ b/src/Neo.SmartContract.Framework/ByteString.cs
@@ -18,6 +18,8 @@ namespace Neo.SmartContract.Framework
 {
     public abstract class ByteString : IEnumerable<byte>
     {
+        public static extern ByteString Empty { [OpCode(OpCode.PUSHDATA1, "00")] get; }
+
         public extern byte this[int index]
         {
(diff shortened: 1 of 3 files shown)


shargon (approved):

Liaojinghui:
This is weird, `empty` is not empty.

Liaojinghui (approved):

PR #712, opened 2021-12-25
Use deterministic in CSharpCompilationOptions

shargon:


erikzhang (approved):
There is nothing wrong with this. But there is actually no difference. Because we will not use the compiled dll in any cases.

Issue #721, opened 2022-03-10
Running nccs at master gives Unhandled Exception


devhawk:
Unhandled exception: System.ArgumentException: The SetHandler call for command 'nccs' is missing an Argument or Option for the parameter at position 0. Did you mean to pass one of these?

   at System.CommandLine.Handler.GetValueForHandlerParameter[T](IValueDescriptor[] symbols, Int32& index, InvocationContext context)
   at System.CommandLine.Handler.<>c__DisplayClass4_0`4.<SetHandler>b__0(InvocationContext context)
   at System.CommandLine.Invocation.AnonymousCommandHandler.<>c__DisplayClass2_0.<.ctor>g__Handle|0(InvocationContext context)
   at System.CommandLine.Invocation.AnonymousCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass18_0.<<UseParseErrorReporting>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass13_0.<<UseHelp>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass22_0.<<UseVersionOption>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass20_0.<<UseTypoCorrections>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseSuggestDirective>b__19_0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass17_0.<<UseParseDirective>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<RegisterWithDotnetSuggest>b__6_0>d.MoveNext()
--- End of stack trace from previous location ---
[...]

PR #723, opened 2022-03-17
Improve CompilationContext.ContractName

### LINKED ISSUE #720: --contract-name option no longer working
PR #704 broke the `--contract-name` option. The point of the `--contract-name ` is to allow the build system to control the name of the output file.

erikzhang:
1. Revert #722
2. Fix #720

@devhawk @shargon

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -53,7 +53,7 @@ namespace Neo.Compiler
 
         public bool Success => diagnostics.All(p => p.Severity != DiagnosticSeverity.Error);
         public IReadOnlyList<Diagnostic> Diagnostics => diagnostics;
-        public string? ContractName => Options.ContractName ?? displayName ?? assemblyName ?? className;
+        public string? ContractName => displayName ?? assemblyName ?? className;
         private string? Source { get; set; }
         internal Options Options { get; private set; }
         internal IEnumerable<IFieldSymbol> StaticFieldSymbols => staticFields.OrderBy(p => p.Value).Select(p => p.Key);
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -35,6 +35,7 @@ namespace Neo.Compiler
         private static readonly MetadataReference[] commonReferences;
         private static readonly Dictionary<string, MetadataReference> metaReferences = new();
         private readonly Compilation compilation;
+        private string? assemblyName, displayName, className;
         private bool scTypeFound;
         private readonly List<Diagnostic> diagnostics = new();
         private readonly HashSet<string> supportedStandards = new();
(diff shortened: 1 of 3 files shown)


shargon (approved):

devhawk:
As I understand it, we can't even run the tool until #721 is fixed, right? IMO, this PR (and all PRs frankly) for devpack should wait until #721 is fixed.

devhawk (changes requested):
The more I think about this, the more I think that it's important NOT to tie the contract name and the base name for contract/manifest/debug info files

erikzhang:
> The more I think about this, the more I think that it's important NOT to tie the contract name and the base name for contract/manifest/debug info files

The logic and behavior of this PR is the same as the current code, just easier to read.

devhawk:
> > The more I think about this, the more I think that it's important NOT to tie the contract name and the base name for contract/manifest/debug info files
> 
> The logic and behavior of this PR is the same as the current code, just easier to read.

This is not accurate. `ProcessOutputs` currently uses `options.ContractName` as the base file name if it's not null and then falls back to using `context.ContractName`. That code was replaced with code that uses `context.ContractName` as the base file name.

Build tools need a mechanism to control the file name of the generated files, without depending on the contents of any source file. At the same time, I think it's important for developers to be able to manage the names of their contracts without constraints imposed by the build tools 

Maybe we should add a separate option to specify the base name of the generated files? The .NET C# compiler uses [`-out:`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/output#outputassembly) argument, but that would collide with the existing `--output` NCCS compiler argument. How about `--base-name`?

src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -126,27 +126,25 @@ private static int ProcessOutputs(Options options, string folder, CompilationCon
             if (context.Success)
             {
                 folder = options.Output ?? Path.Combine(folder, "bin", "sc");
-                var contractName = string.IsNullOrEmpty(options.ContractName)
-                    ? context.ContractName : options.ContractName;
                 Directory.CreateDirectory(folder);
-                string path = Path.Combine(folder, $"{contractName}.nef");
+                string path = Path.Combine(folder, $"{context.ContractName}.nef");
```
devhawk, line 130: Using context.ContractName here forces the contract name in the manifest to be the same as the base name of the generated files.

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -52,7 +53,7 @@ public class CompilationContext
 
         public bool Success => diagnostics.All(p => p.Severity != DiagnosticSeverity.Error);
         public IReadOnlyList<Diagnostic> Diagnostics => diagnostics;
-        public string? ContractName { get; private set; }
+        public string? ContractName => Options.ContractName ?? displayName ?? assemblyName ?? className;
```
devhawk, line 56: This line overrides the contract name in source, re-breaking #674 and #696

erikzhang:
> Maybe we should add a separate option to specify the base name of the generated files? The .NET C# compiler uses [-out:](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/output#outputassembly) argument, but that would collide with the existing --output NCCS compiler argument. How about --base-name?

I think it's okay. And `--contract-name` can be removed.

devhawk (approved):

PR #724, opened 2022-03-17
Fix paths

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/721


Liaojinghui (approved):

devhawk:
We should add a test that actually runs the tool from the command line so we don't get a bug like this again

erikzhang (approved):

erikzhang:
Tested?

Liaojinghui:
Problem gone on my machine

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

shargon:
I tested it locally, @superboyiii could you confirm that the issue was fixed?

Liaojinghui:
All tests pass on my machine as well

PR #731, opened 2022-04-21
Sync GetTransactionSigners

shargon:
Sync with https://github.com/neo-project/neo/pull/2685

### DIFF
--- a/src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
+++ b/src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
@@ -20,7 +20,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" />
-    <PackageReference Include="Neo" Version="3.2.1" />
+    <PackageReference Include="Neo" Version="3.2.1-CI01335" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta3.22114.1" />
     <PackageReference Include="System.CommandLine.NamingConventionBinder" Version="2.0.0-beta3.22114.1" />
   </ItemGroup>
--- a/src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
+++ b/src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
@@ -20,7 +20,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" />
-    <PackageReference Include="Neo" Version="3.2.1-CI01335" />
+    <PackageReference Include="Neo" Version="3.2.1" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta3.22114.1" />
     <PackageReference Include="System.CommandLine.NamingConventionBinder" Version="2.0.0-beta3.22114.1" />
   </ItemGroup>
--- a/src/Neo.SmartContract.Framework/Native/Ledger.cs
+++ b/src/Neo.SmartContract.Framework/Native/Ledger.cs
@@ -27,5 +27,6 @@ namespace Neo.SmartContract.Framework.Native
         public static extern Transaction GetTransactionFromBlock(UInt256 blockHash, int txIndex);
         public static extern Transaction GetTransactionFromBlock(uint blockHeight, int txIndex);
         public static extern int GetTransactionHeight(UInt256 hash);
+        public static extern object[] GetTransactionSigners(UInt256 hash);
     }
 }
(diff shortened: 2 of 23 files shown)


src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
```diff
@@ -20,7 +20,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.1.0" />
-    <PackageReference Include="Neo" Version="3.2.1" />
+    <PackageReference Include="Neo" Version="3.2.1-CI01335" />
```
erikzhang, line 23: We don't allow CI packages in master branch.

erikzhang (changes requested):

src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
shargon: Fixed

erikzhang:
And `GetTransactionVMState`?

src/Neo.SmartContract.Framework/Native/Ledger.cs
```diff
@@ -27,5 +27,6 @@ public class Ledger
         public static extern Transaction GetTransactionFromBlock(UInt256 blockHash, int txIndex);
         public static extern Transaction GetTransactionFromBlock(uint blockHeight, int txIndex);
         public static extern int GetTransactionHeight(UInt256 hash);
+        public static extern object[] GetTransactionSigners(UInt256 hash);
```
erikzhang, line 30: Return `Signer[]` instead of `object[]`?

shargon: Done

src/Neo.SmartContract.Framework/Native/Signer.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Neo.SmartContract.Framework.Native
+{
+    public class Signer
+    {
+        public byte[] Data;
+        public UInt160 Account;
+        public WitnessScope Scopes;
+        public UInt160[] AllowedContracts;
+        public ECPoint[] AllowedGroups;
+        public object[] Rules;
```
erikzhang, line 22: `WitnessRule`

shargon: How can we convert between different types?

erikzhang: Check `WitnessRule.Type` and then convert.

shargon: Done. Please, check it if do you want one file per class.

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_Blockchain.cs
```diff
@@ -83,6 +83,8 @@ private static object GetTxInfo(Transaction tx, string whatReturn)
             if (whatReturn == "SystemFee") return tx.SystemFee;
             if (whatReturn == "ValidUntilBlock") return tx.ValidUntilBlock;
             if (whatReturn == "Version") return tx.Version;
+            if (whatReturn == "Signers") return Ledger.GetTransactionSigners(tx.Hash);
```
Ashuaidehao, line 86: Please use these code here and add several "WitnessRule"signers in test engine tx:
```csharp
 {
                var signers = Ledger.GetTransactionSigners(tx.Hash);
                foreach (var signer in signers)
                {
                    var rule = signer.Rules;
                }
                return Ledger.GetTransactionSigners(tx.Hash);
            }
```

As we test, it will throw an exception `The given key 'Neo.VM.Types.Array' was not present in the dictionary.` in **ReferenceCounter.CheckZeroReferred()**.
This may be caused by missing "ReferenceCounter" in these lines:

https://github.com/neo-project/neo/blob/ed68274e7dec2622468d3bc67df6e0f2913d903b/src/neo/Network/P2P/Payloads/Signer.cs#L194

https://github.com/neo-project/neo/blob/ed68274e7dec2622468d3bc67df6e0f2913d903b/src/neo/Network/P2P/Payloads/Signer.cs#L195

https://github.com/neo-project/neo/blob/ed68274e7dec2622468d3bc67df6e0f2913d903b/src/neo/Network/P2P/Payloads/Signer.cs#L196

mialbu: I can confirm this. As soon as a signer is present with witness scope `WitnessRules`, I get the same exception (`The given key 'Neo.VM.Types.Array' was not present in the dictionary.`) when trying to access a signer from the list.

I've also noticed that when only one signer with witness scope `WitnessRules` is present:
- I can access only this signer from the list, but retrieving its fields (`account`, etc.) still results in the same exception.
- Trying to access other signers from the list results in the same exception.

When more than one signers with witness scope `WitnessRules` are present, I cannot access any of the signers (same exception).

erikzhang: It should be fixed by neo-project/neo#2711

src/Neo.SmartContract.Framework/Native/Signer.cs
shargon: Done

erikzhang (approved):

PR #735, opened 2022-05-05
Allow unsafe keyword making exceptions uncatchable

erikzhang:


Ashuaidehao:
I make a quick test and find only *UnsafeThrowCall_02* works as expected.
```csharp

        public static object tryUnsafeCall()
        {
            try
            {
                UnsafeThrowCall_02();
            }
            catch (System.Exception)
            {

                return -10;
            }
            return 10;
        }

        public static unsafe void UnsafeThrowCall_01()
        {
            throwcall();
        }

        public static void UnsafeThrowCall_02()
        {
            unsafe
            {
                throwcall();
            }
        }


        public static object throwcall()
        {
            throw new System.Exception();
        }
```

erikzhang:
> I make a quick test and find only UnsafeThrowCall_02 works as expected. "unsafe" keyword seems not work in other positions.

Fixed. Please try it again.

shargon:
@Ashuaidehao could you add your test as unit tests?

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -441,7 +452,6 @@ private void ConvertNoBody(AccessorDeclarationSyntax syntax)
 
         private void ConvertSource(SemanticModel model)
         {
-            if (SyntaxNode is null) return;
```
shargon, line 444: This line must be removed?

erikzhang: It has been moved to L157

src/Neo.SmartContract.Framework/Nep11Token.cs
```diff
@@ -127,7 +127,7 @@ private static void UpdateBalance(UInt160 owner, ByteString tokenId, int increme
                 accountMap.Delete(key);
         }
 
-        private static void PostTransfer(UInt160 from, UInt160 to, ByteString tokenId, object data)
+        private static unsafe void PostTransfer(UInt160 from, UInt160 to, ByteString tokenId, object data)
```
shargon, line 130: NEO y GAS will have a different behaviour

erikzhang: They are the same. Because native contracts use `ExecutionEngine.CallFromNativeContract()` to call other contracts. And non-native exceptions cannot pass through native contracts.

See https://github.com/neo-project/neo/blob/ed68274e7dec2622468d3bc67df6e0f2913d903b/src/neo/SmartContract/ApplicationEngine.cs#L266-L267

Ashuaidehao:
> @Ashuaidehao could you add your test as unit tests?

Yes, all my tests are pass now. Please check if some more should be added
https://github.com/neo-project/neo-devpack-dotnet/pull/736

shargon (approved):

erikzhang:
@devhawk What do you think?

shargon:
What about implement it in the reverse order? if you use `unsafe`, you are able to catch the exception.

erikzhang:
We have a `safe` field in ABI which is usually used on stateless methods. So it makes more sense to mark stateful methods as unsafe.

devhawk:
> @devhawk What do you think?

I don't fully understand the bigger picture of what this PR is trying to accomplish - particularly in concert with a few other exception related issues/PRs in other repos. 

However, I do not think we should be using the existing `unsafe` keyword for this. The semantics that this PR attaches to `unsafe` don't appear to relate to the semantics of `unsafe` in normal, non-Neo related C#.  I believe this will confuse existing C# developers significantly. 

If we really think automating this behavior in the compiler is important, I think we'd be better off creating a custom attribute such as `[AbortOnException]`. However, it's not clear to me that we need a compiler feature to wrap the token `PostTransfer` code in a try/finally block. Can't we just add the try/finally directly to the PostTransfer methods? Seems like a lot of compiler work to avoid adding three lines of code to two methods.

``` cs
private static void PostTransfer(UInt160 from, UInt160 to, BigInteger amount, object data)
{
    try
    {
        OnTransfer(from, to, amount);
        if (to is not null && ContractManagement.GetContract(to) is not null)
            Contract.Call(to, "onNEP17Payment", CallFlags.All, from, amount, data);
    }
    finally
    {
        ExecutionEngine.Abort();
    }
}
```

devhawk (changes requested):
As per my comment, we should not change the semantics of `unsafe` keyword

devhawk:
Based on my review of https://github.com/neo-project/neo/pull/2718, I don't think we need this PR as I believe the specific part of https://github.com/neo-project/neo/pull/2718 that relates to this PR should not be merged

erikzhang:
> Based on my review of https://github.com/neo-project/neo/pull/2718, I don't think we need this PR as I believe the specific part of https://github.com/neo-project/neo/pull/2718 that relates to this PR should not be merged

Not only https://github.com/neo-project/neo/pull/2718. Any contract you call may throw an exception. We need a flexible and convenient way to prevent exceptions from passing through.

> However, I do not think we should be using the existing unsafe keyword for this. The semantics that this PR attaches to unsafe don't appear to relate to the semantics of unsafe in normal, non-Neo related C#. I believe this will confuse existing C# developers significantly.

In Neo, we can never use C#'s unmanaged code and pointer functions. This means that the native functionality of `unsafe` is meaningless in Neo, and I just subtly borrowed the keyword. I originally thought of using attribute too, but it doesn't work like this:

```c#
public void TestMethod()
{
    //Do something first.
    unsafe
    {
        //Call some contracts that may throw exceptions.
    }
    //Do the rest.
}
```

erikzhang:
> Can't we just add the try/finally directly to the PostTransfer methods?

In fact what we need is a try/catch clause.

```c#
    try
    {
        OnTransfer(from, to, amount);
        if (to is not null && ContractManagement.GetContract(to) is not null)
            Contract.Call(to, "onNEP17Payment", CallFlags.All, from, amount, data);
    }
    catch
    {
        ExecutionEngine.Abort();
    }
```

txhsl:
I'm bound to say that the keyword `unsafe` is a bit of confusing. `CatchableException` in C# is manageable in execution, which enhances security rather than increasing risk. So IMO, it is safe rather than unsafe, `try/catch` is easier to understand. 🤔

devhawk:
> > Can't we just add the try/finally directly to the PostTransfer methods?
> 
> In fact what we need is a try/catch clause.

You're right, of course it's try/catch not try finally.

> In Neo, we can never use C#'s unmanaged code and pointer functions. This means that the native functionality of `unsafe` is meaningless in Neo, and I just subtly borrowed the keyword. 

Changing the semantics of the `unsafe` keyword is confusing, especially for developers who are experienced with C# but new to Neo - an audience we are explicitly trying to attract. 

I don't think it's worth confusing developers it to save writing a single line catch block.

> But we may need [try/catch] everywhere and make the code less readable. If we use the `unsafe` keyword, everything will be fine.

As stated, I find changing the meaning of `unsafe` is much less readable than try/catch. 

Beyond that, I don't understand how prevalent this pattern is. You say we "may need it everywhere". What are the scenarios where we need to explicitly catch and abort vs simply letting the exception remain uncaught? If the exception remains uncaught, the engine is going to fault anyway. Are we worried about code up the call stack catching the exception and trying to handle it? Have we got real-world examples of contracts that are doing this?

To summarize:
* It's not clear to me how common the "abort on exception" pattern is or should be
* If the "abort on exception" pattern is or should be common, it seems to me that try/catch is a perfectly reasonable solution
* If we really need a better solution than try/catch, I think we need to come up with a design doesn't change semantics of the existing `unsafe` keyword

shargon:
What about create a new CallFlag. `CrossContractCatcheable`, if a throw need to cross around different contracts, it will need this flag otherwise the execution will be aborted.
@neo-project/developers

erikzhang:
> What about create a new CallFlag. CrossContractCatcheable, if a throw need to cross around different contracts, it will need this flag otherwise the execution will be aborted.

Very good idea! Please check neo-project/neo#2722

PR #737, opened 2022-05-07
Fix AttributeTargets of ContractHashAttribute

erikzhang:


devhawk (commented):
Why are we changing this?

erikzhang:
Because `ContractHashAttribute` should be applied to properties. The original usage looks awkward.

devhawk:
> Because `ContractHashAttribute` should be applied to properties. The original usage looks awkward.

Not sure it matters much if the smart contract framework declarations look awkward. But as long as this PR doesn't break compilation of existing contract source, I'm fine with it.

shargon (approved):
The code looks good

PR #738, opened 2022-05-07
Add more math methods

erikzhang:


Liaojinghui (approved):

Liaojinghui:
Can we also support `unchecked` keyword?

erikzhang:
The underlying type `BigInteger` is always "unchecked" in neo-vm.

PR #742, opened 2022-05-09
Support conditional compilation

### LINKED ISSUE #664: nccs should support #if Debug conditional compilation
nccs can't detect #if Debug statement which neon/Neo.Compiler.MSIL compiler supports.

erikzhang:
Close #664

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The Neo.Compiler.CSharp is free software distributed under the MIT 
 // software license, see the accompanying file LICENSE in the main directory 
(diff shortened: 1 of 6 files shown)


erikzhang:
We can call `Runtime.Log()` in a `#if DEBUG` block. It is useful.

shargon (dismissed):
Agree, It could be useful

devhawk (approved):
The basic *idea* of conditional compilation support gets a big 👍 from me.

I'm not a fan of *how* we've implemented this in `nccs`. It feels to me that `nccs` is replicating a bunch of MSBuild logic - including conditional compilation settings - that would be better left to MSBuild. But that's a general issue w/ how `nccs` interacts with project files that is not specific to this PR

roschler:
I want to emphasize the need for `Runtime.Log()` statements as was pointed out.

This leads to a real workflow problem because you can't use those in a method marked as `[Safe]`.  You will get a SYSCALL error from the VM.  So you end up with this inconvenient workflow.  You have to remove the `[Safe]` attribute from every method that uses `Runtime.Log()` calls during dev, and then (have to remember) to restore that attribute and also comment out all the `Runtime.Log()` calls when it comes time to publish your smart contract.  What happens if you forget to restore the attribute?

Why doesn't the NVM just ignore those calls in a `[Safe]` method instead of throwing an error?

erikzhang:
> This leads to a real workflow problem because you can't use those in a method marked as [Safe].

With this pr, you can use `#if DEBUG` to disable `[Safe]` temporarily now.

roschler:
Thanks Erik!  How do I get this version into my build set?

erikzhang:
After our 3.2.2 release, you can get compiler, framework, and project template from nuget.

PR #745, opened 2022-05-09
Storage backed properties

### LINKED ISSUE #585: Generate  'get' and 'set' by compiler for contract global variable?
``` C#
        public BigInteger example 
        {
            get
            {
                var val = Storage.Get(Storage.CurrentContext, nameof(example));
                return val is null ? 0 : (BigInteger)val;
            }
            set => Storage.Put(Storage.CurrentContext, nameof(example), value);
        }
```
From now on, we have to write the code 'set' and 'set' for all variable in contract.
I would like to know can we generate this part by compiler automatically?

It will save lots of coding works.

erikzhang:
Close #585

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert.cs
@@ -517,9 +517,18 @@ namespace Neo.Compiler
             if (Symbol.MethodKind == MethodKind.PropertyGet)
             {
                 JumpTarget endTarget = new();
-                // AddInstruction(OpCode.DUP);
-                AddInstruction(OpCode.ISNULL);
-                Jump(OpCode.JMPIFNOT_L, endTarget);
+                if (Symbol.IsStatic)
+                {
+                    // AddInstruction(OpCode.DUP);
+                    AddInstruction(OpCode.ISNULL);
+                    // Ensure that no object was sent
+                    Jump(OpCode.JMPIFNOT_L, endTarget);
+                }
+                else
+                {
+                    // Check class
+                    Jump(OpCode.JMPIF_L, endTarget);
+                }
                 Push(key);
                 Call(ApplicationEngine.System_Storage_GetReadOnlyContext);
                 Call(ApplicationEngine.System_Storage_Get);
--- a/src/Neo.Compiler.CSharp/MethodConvert.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert.cs
@@ -28,6 +28,7 @@ using System.Collections.ObjectModel;
 using System.Linq;
 using System.Numerics;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography.Xml;
 using System.Text;
 
 namespace Neo.Compiler
@@ -484,10 +485,36 @@ namespace Neo.Compiler
             }
         }
 
+        private byte[] GetStorageBackedKey(IPropertySymbol property, AttributeData attribute)
+        {
+            byte[] key;
+
+            if (attribute.ConstructorArguments.Length == 0)
+            {
+                key = Utility.StrictUTF8.GetBytes(property.Name);
+            }
+            else
+            {
+                if (attribute.ConstructorArguments[0].Value is byte b)
+                {
+                    key = new byte[] { b };
+                }
+                else if (attribute.ConstructorArguments[0].Value is string s)
+                {
+                    key = Utility.StrictUTF8.GetBytes(s);
+                }
+                else
+                {
+                    throw new CompilationException(Symbol, DiagnosticId.SyntaxNotSupported, $"Unknown StorageBacked constructor: {Symbol}");
+                }
+            }
+            return key;
+        }
+
         private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeData attribute)
         {
             IFieldSymbol[] fields = property.ContainingType.GetAllMembers().OfType<IFieldSymbol>().ToArray();
-            byte[] key = Encoding.UTF8.GetBytes(property.Name).Prepend((byte)attribute.ConstructorArguments[0].Value!).ToArray();
+            byte[] key = GetStorageBackedKey(property, attribute);
             if (Symbol.MethodKind == MethodKind.PropertyGet)
             {
                 JumpTarget endTarget = new();
--- a/src/Neo.Compiler.CSharp/MethodConvert.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert.cs
@@ -396,48 +396,151 @@ namespace Neo.Compiler
 
         private void ConvertNoBody(AccessorDeclarationSyntax syntax)
         {
-            _inline = true;
             _callingConvention = CallingConvention.Cdecl;
             IPropertySymbol property = (IPropertySymbol)Symbol.AssociatedSymbol!;
-            INamedTypeSymbol type = property.ContainingType;
-            IFieldSymbol[] fields = type.GetAllMembers().OfType<IFieldSymbol>().ToArray();
+            AttributeData? attribute = property.GetAttributes().FirstOrDefault(p => p.AttributeClass!.Name == nameof(StorageBackedAttribute));
             using (InsertSequencePoint(syntax))
             {
+                _inline = attribute is null;
+                ConvertFieldBackedProperty(property);
+                if (attribute is not null)
+                    ConvertStorageBackedProperty(property, attribute);
+            }
+        }
+
+        private void ConvertFieldBackedProperty(IPropertySymbol property)
+        {
+            IFieldSymbol[] fields = property.ContainingType.GetAllMembers().OfType<IFieldSymbol>().ToArray();
+            if (Symbol.IsStatic)
+            {
+                IFieldSymbol backingField = Array.Find(fields, p => SymbolEqualityComparer.Default.Equals(p.AssociatedSymbol, property))!;
+                byte backingFieldIndex = context.AddStaticField(backingField);
+                switch (Symbol.MethodKind)
+                {
+                    case MethodKind.PropertyGet:
+                        AccessSlot(OpCode.LDSFLD, backingFieldIndex);
+                        break;
+                    case MethodKind.PropertySet:
+                        if (!_inline) AccessSlot(OpCode.LDARG, 0);
+                        AccessSlot(OpCode.STSFLD, backingFieldIndex);
+                        break;
+                    default:
+                        throw new CompilationException(Symbol, DiagnosticId.SyntaxNotSupported, $"Unsupported accessor: {Symbol}");
+                }
+            }
+            else
+            {
+                fields = fields.Where(p => !p.IsStatic).ToArray();
+                int backingFieldIndex = Array.FindIndex(fields, p => SymbolEqualityComparer.Default.Equals(p.AssociatedSymbol, property));
+                switch (Symbol.MethodKind)
+                {
+                    case MethodKind.PropertyGet:
+                        if (!_inline) AccessSlot(OpCode.LDARG, 0);
+                        Push(backingFieldIndex);
+                        AddInstruction(OpCode.PICKITEM);
+                        break;
+                    case MethodKind.PropertySet:
+                        if (_inline)
+                        {
+                            Push(backingFieldIndex);
+                            AddInstruction(OpCode.ROT);
+                        }
+                        else
+                        {
+                            AccessSlot(OpCode.LDARG, 0);
+                            Push(backingFieldIndex);
+                            AccessSlot(OpCode.LDARG, 1);
+                        }
+                        AddInstruction(OpCode.SETITEM);
+                        break;
+                    default:
+                        throw new CompilationException(Symbol, DiagnosticId.SyntaxNotSupported, $"Unsupported accessor: {Symbol}");
+                }
+            }
+        }
+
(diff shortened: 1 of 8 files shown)


devhawk (changes requested):
This doesn't seem worth it to me for simple, single value storage properties. 

I would *love* to see a declarative solution for contract storage. It would simplify contract development + the debugger could use up the storage declarations automatically for [storage key & item decoding](https://github.com/neo-project/neo-debugger/blob/master/docs/storage-schema-overview.md).

Instead of just solving the simplest storage problem (single value storage such as `TotalSupply`), can we instead work on an approach that would also work for:

* Single key storage maps (such as NEP 17 token balances)
* Multiple key storage maps
* Key index storages (such as NEP 11 account token storage)
* Storage Lists (i.e. storage maps w/ integer keys with List like semantics to Add/Remove items from the end)

src/Neo.SmartContract.Framework/Nep17Token.cs
```diff
@@ -50,7 +50,7 @@ protected static void Mint(UInt160 account, BigInteger amount)
             if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
             if (amount.IsZero) return;
             UpdateBalance(account, +amount);
-            UpdateTotalSupply(+amount);
+            TotalSupply += amount;
```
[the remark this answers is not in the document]
erikzhang: Using a compiler is inherently more expensive. This is inevitable. And the price of `System.Storage.GetContext` is only equivalent to executing `OpCode.ADD` twice, which is negligible in my opinion.

erikzhang:
>Instead of just solving the simplest storage problem (single value storage such as TotalSupply), can we instead work on an approach that would also work for:
>
>Single key storage maps (such as NEP 17 token balances)
Multiple key storage maps
Key index storages (such as NEP 11 account token storage)
Storage Lists (i.e. storage maps w/ integer keys with List like semantics to Add/Remove items from the end)

Maybe we can go step by step. I'm currently just implementing the functionality requested in #585. These complex functions you mentioned can be discussed separately by creating a new issue.

shargon (approved):

vncoelho:
Good change!

vncoelho (dismissed):

devhawk:
> Maybe we can go step by step. I'm currently just implementing the functionality requested in #585. These complex functions you mentioned can be discussed separately by creating a new issue.

Happy to file a new issue, but I don't think we should approach this "step by step". From what I've seen, the majority of contract storages are not simple single values like this. Most contract storages use `StorageMap` with one or more keys in order to store multiple values.

This PR saves the developer a couple lines of code, doesn't solve the real problem and potentially constrains our ability to deliver a complete solution in the future. Not worth it in my opinion.

src/Neo.SmartContract.Framework/Attributes/StorageBackedAttribute.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (C) 2015-2022 The Neo Project.
```
shargon:
```suggestion
// Copyright (C) 2015-2023 The Neo Project.
```

src/Neo.SmartContract.Framework/Nep11Token.cs
```diff
@@ -32,7 +32,7 @@ public abstract class Nep11Token<TokenState> : TokenContract
         protected const byte Prefix_AccountToken = 0x04;
 
         [Safe]
```
shargon, line 34: Using `Safe` in base class, we should check if we need to define it also here (in get)

shargon:
I think that this PR is an important one, we should work on it @Liaojinghui

Liaojinghui (dismissed):

Liaojinghui:
@shargon is this compatible to the previous code?

shargon:
> @shargon is this compatible to the previous code?

We need to check it 😅

Liaojinghui:
Since all unit test passes, it should be fine. I have approved it, you can merge when you are avilable.

shargon:
It seems that `.get` and `static` properties are not working.

Liaojinghui (approved):

Liaojinghui:
Was about to comment, misclicked approve.

Liaojinghui:
If we do it this way, it works:

```C#
       public static BigInteger getWithKey()
        {
            ValueB = value; return ValueB;
            
        }
```

if we remove ` ValueB = value;`, it fault,,,,,,even though this code does nothing.

Liaojinghui:
I can only dig this far:

`context_pop.EvaluationStack.Count != context_pop.RVCount` check fail as `EvaluationStack.Count` is 2 and `RVCount` is 1

Screenshot 2023-11-16 at 7 08 07 PM

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -421,48 +421,177 @@ private void ConvertExtern()
...
+                    case "String":
+                    case "ByteString":
+                    case "UInt160":
+                    case "UInt256":
+                        break;
+                    default:
+                        Call(NativeContract.StdLib.Hash, "deserialize", 1, true);
+                        break;
+                }
+                AddInstruction(OpCode.DUP);
```
Liaojinghui, line 548:
```suggestion
                // AddInstruction(OpCode.DUP);
```

shargon:
> I can only dig this far:
> 
> `context_pop.EvaluationStack.Count != context_pop.RVCount` check fail as `EvaluationStack.Count` is 2 and `RVCount` is 1
> 
> Screenshot 2023-11-16 at 7 08 07 PM

Maybe we need to remove the null before, when is static

Liaojinghui:
I will keep debug tomorrow if you haven't solved it by then. But this really makes me believe we should also add the compiler to the monorepo, i need to change reference of neo and vm to debug the issue.

vncoelho:
So move it all for now and let's see this experience.

cschuchardt88:
@Liaojinghui  Just create your own nuget store on your computers file system

https://learn.microsoft.com/en-us/nuget/hosting-packages/local-feeds

Liaojinghui:
> @Liaojinghui Just create your own nuget store on your computers file system
> 
> https://learn.microsoft.com/en-us/nuget/hosting-packages/local-feeds

But why not add it to the monorepo directly,,,,much easier

Liaojinghui:
> So move it all for now and let's see this experience.

@vncoelho I tried to remove the DUP, but still wont work, I think there are logic issues in the convert.

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_StorageBacked.cs
```diff
@@ -0,0 +1,47 @@
...
+using Neo.SmartContract.Framework;
+using Neo.SmartContract.Framework.Attributes;
+using Neo.SmartContract.Framework.Services;
+
+namespace Neo.SmartContract.Framework.UnitTests.TestClasses
+{
+    public class Contract_StorageBacked : SmartContract
+    {
+        [StorageBacked]
+        public BigInteger ValueA { [Safe] get; protected set; }
```
Liaojinghui, line 11:
```suggestion
        public static BigInteger ValueA { [Safe] get; protected set; }
```

tests/Neo.SmartContract.Framework.UnitTests/Services/BackedStorageTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public void Test_Kind(string kind)
+        {
+            Console.WriteLine("GET");
+            _engine.Reset();
+
+            var result = _engine.ExecuteTestCaseStandard("get" + kind);
+            Assert.AreEqual(VMState.HALT, _engine.State);
+            Assert.AreEqual(new BigInteger(0), result.Pop().GetInteger());
```
Liaojinghui, line 51:
```suggestion
            Assert.IsTrue(result.Pop().IsNull);
```

Liaojinghui:
Good to go now @shargon

Liaojinghui (approved):

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_StorageBacked.cs
shargon: We should test it with and without static

shargon: And using public getters and setters

Liaojinghui: > We should test it with and without static

Can you please clearify the behavior? You want non-static values behave the same as static values when marked as Storage? BTW, i suggest to rename as `Storage` instead of `StorageBacked`

shargon: > I would suggest to remove storageback support for non-static.

Why if we allow non-static methods?

shargon (commented):
@superboyiii could you test this new feature?

Liaojinghui:
@shargon getter is public by default, but do you think we should also add private getter UT tests?

shargon:
> @shargon getter is public by default, but do you think we should also add private getter UT tests?

Yes, all unit tests as possible

Liaojinghui:
My bad on the previous commit, working on fixing

Liaojinghui (approved):

Liaojinghui:
@superboyiii  please double check.

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_StorageBacked.cs
```diff
@@ -47,11 +47,11 @@ public static void putWithString(BigInteger value)
         public static BigInteger getWithString() => WithString;
 
 
-        [StorageBacked("testMe")]
+        [StorageBacked]
```
shargon, line 50: I want to test with different kind of prefixes

Liaojinghui: i think i did not change yours, its something that I added for the getter and setter tests

tests/Neo.SmartContract.Framework.UnitTests/TestClasses/Contract_StorageBacked.cs
Liaojinghui: > BTW, i suggest to rename as `Storage` or `Singleton` or `Global` instead of `StorageBacked`

@shargon will you consider this? Storageback is a weird name.

shargon: `StorageAttribute`

Liaojinghui: I vote the first one.

Liaojinghui (approved):

shargon:
Ping @superboyiii

superboyiii:
> Ping @superboyiii

I will

superboyiii:
Tested, it works for me. But seems it's not a correct exception message when someone not owner invoke `private set`.

shargon:
> Tested, it works for me. But seems it's not a correct exception message when someone not owner invoke `private set`. @shargon @Liaojinghui `Not Owner invocation` 
> 
> `Owner invocation` 

Could you open a issue in neo-modules, seems that the error when a method is not found is not well described

shargon (approved):

superboyiii:
> Could you open a issue in neo-modules, seems that the error when a method is not found is not well described

Sorry, it's becasue of the different scope accout from opened account.

superboyiii:
I tried to invoke a private set with an account differring from owner address, but can get the value without any interaction with owner address from deploying. Is this an expected behaviour? @Liaojinghui  @shargon 
```
//initial operator
        [InitialValue("NTADuu5qt2Rt2tder3dx4RDAWjsKozxdMT", ContractParameterType.Hash160)]
        private static readonly UInt160 Owner = default;
```

```
[StorageBacked]
        public static string Test
        {
            get;
            private set;
        }

        public static void TryTest2()
        {
            Test = "abc";
        }

        [Safe]
        public string getTest() => Test;
```

Liaojinghui:
how is that possible to invoke a private set? its not in the abi

superboyiii:
> how is that possible to invoke a private set? its not in the abi

I mean invoke `tryTest2` method here.

superboyiii:
> how is that possible to invoke a private set? its not in the abi...oh, you are calling a private set in an interface. Of course it is expected. if user does not add permission to the interface, ofcourse it can be called.

Why don't make the private set method check owner witness?

Liaojinghui:
Well, i am not sure, but isn't this a default behavior of all private methods in neo contract? you explicitely add checks, instead of by us.

superboyiii:
> Well, i am not sure, but isn't this a default behavior of all private methods in neo contract? you explicitely add checks, instead of by us.
I've already set that in verfy(), is this not enough?

superboyiii:
OK, I see. This private function is not a private function for owner but just can't be seen in Abi. Owner still need manually set funcB to invoke funcA and manually add checkwitness(owner) if he need execute funcA...

PR #746, opened 2022-05-11
Improve StorageMap

erikzhang:


src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -20,68 +20,59 @@ namespace Neo.SmartContract.Framework.Services
...
+        [OpCode(OpCode.NEWBUFFER)]
+        [OpCode(OpCode.TUCK)]
+        [OpCode(OpCode.PUSH0)]
+        [OpCode(OpCode.ROT)]
+        [OpCode(OpCode.SETITEM)]
+        [Syscall("System.Storage.GetContext")]
         [OpCode(OpCode.PUSH2)]
         [OpCode(OpCode.PACK)]
-        public extern StorageMap(StorageContext context, ByteString prefix);
+        public extern StorageMap(byte prefix);
```
shargon, line 75: If you want to create two `StorageMap`, it's not an improve, you will waste more gas than.

```
var context = StorageContext.Current;
var a = new StorageMap(context, 0);
var b = new StorageMap(context, 1);
```

erikzhang: The `StorageContext` overload is still there.

shargon:
Now we will only have a byte prefix?

erikzhang:
I think a byte is enough.

shargon:
> I think a byte is enough.

For me it's good like this, but if a token wants to update the contract with the latest framework, it might have problems migrating the keys.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -20,68 +20,59 @@ namespace Neo.SmartContract.Framework.Services
 {
     public class StorageMap
     {
-        private readonly StorageContext Context;
-        private readonly ByteString Prefix;
+        private readonly StorageContext context;
+        private readonly byte[] prefix;
```
shargon, line 24:
```suggestion
        private readonly byte prefix;
```

erikzhang: It will be a problem for a byte when executing `OpCode.CAT`.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -130,85 +137,78 @@ public object GetObject(byte[] key)
...
         [OpCode(OpCode.CAT)]
         [OpCode(OpCode.SWAP)]
-        [OpCode(OpCode.PUSH0)]
-        [OpCode(OpCode.PICKITEM)]
         [Syscall("System.Storage.Find")]
         public extern Iterator Find(ByteString prefix, FindOptions options = FindOptions.None);
 
+        public Iterator<T> Find<T>(byte[] prefix)
+        {
+            return (Iterator<T>)Find(prefix, FindOptions.ValuesOnly | FindOptions.DeserializeValues);
```
shargon, line 167: Set the flags as parameter?

erikzhang: If it's a parameter, how do I return a generic? They have overloads for non-generic versions that take arguments.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -20,59 +20,70 @@ namespace Neo.SmartContract.Framework.Services
...
+        private readonly StorageContext context;
+        private readonly byte[] prefix;
 
         public extern ByteString this[ByteString key]
         {
-            [OpCode(OpCode.OVER)]
-            [OpCode(OpCode.PUSH1)]
-            [OpCode(OpCode.PICKITEM)]
-            [OpCode(OpCode.SWAP)]
+            [CallingConvention(CallingConvention.Cdecl)]
```
devhawk, line 28: Isn't cdecl the default?

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -130,85 +137,78 @@ public object GetObject(byte[] key)
             return StdLib.Deserialize(value);
         }
 
+        public Iterator<T> Find<T>()
+        {
+            return (Iterator<T>)Find(FindOptions.ValuesOnly | FindOptions.DeserializeValues);
```
devhawk, line 142: It doesn't feel like this specific `Find` scenario is common enough to be the default.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -130,85 +137,78 @@ public object GetObject(byte[] key)
...
-        [OpCode(OpCode.PUSH0)]
-        [OpCode(OpCode.PICKITEM)]
+        [OpCode(OpCode.UNPACK)]
+        [OpCode(OpCode.DROP)]
         [Syscall("System.Storage.Find")]
         public extern Iterator Find(FindOptions options = FindOptions.None);
 
+        public Iterator<T> Find<T>(ByteString prefix)
+        {
+            return (Iterator<T>)Find(prefix, FindOptions.ValuesOnly | FindOptions.DeserializeValues);
```
devhawk, line 153: Same comment from above. Is `ValuesOnly | DeserializeValues` the most common scenario for `Find`?

These methods aren't saving much, maybe they aren't worth it?

src/Neo.SmartContract.Framework/Services/StorageMap.cs
devhawk: Same comment from above

src/Neo.SmartContract.Framework/Services/StorageMap.cs
erikzhang: This is the only scenario for this generic overload. Other scenarios should use the non-generic version of this method.

erikzhang:
Merge?

src/Neo.SmartContract.Framework/Services/StorageMap.cs
devhawk: What if the value shouldn't be deserialized? If you tried to loop thru balance records in a NEP17 contract using `Find<BigInteger>`, won't it throw?

devhawk:
> Merge?

I'm still not convinced on the `Find<T>` overloads. I agree with your point that `ValuesOnly` is the only scenario for a generic `Find<T>` method. However, I'm worried that a developer will try to `Find<int>` or `Find<string>` and the find method would blow up.

Suggestion: add a `bool deserializeValue = true` parameter to the generic `Find` methods so callers can disable the deserialization.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
erikzhang: I've removed the generic overloads.

src/Neo.SmartContract.Framework/Services/StorageMap.cs
erikzhang: I've removed the generic overloads.

shargon (approved):

PR #755, opened 2022-06-20
no_reentry

Liaojinghui:
Alternative solution to #754


Liaojinghui:
Hey @shargon , how about this one?

PR #756, opened 2022-06-21
No reentrant modifier

shargon:
Alternative for https://github.com/neo-project/neo-devpack-dotnet/pull/755 without change the core

### DIFF
--- a/src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
+++ b/src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
@@ -1,0 +1,41 @@
+// Copyright (C) 2015-2022 The Neo Project.
+// 
+// The Neo.SmartContract.Framework is free software distributed under the MIT 
+// software license, see the accompanying file LICENSE in the main directory 
+// of the project or http://www.opensource.org/licenses/mit-license.php 
+// for more details.
+// 
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using System;
+using Neo.SmartContract.Framework.Services;
+
+namespace Neo.SmartContract.Framework.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public abstract class NoReentrantAttribute : ModifierAttribute, IDisposable
+    {
+        private readonly StorageMap _context;
+        private const string _key = "noReentrant";
+
+        public NoReentrantAttribute(byte prefix = 0xFF)
+        {
+            _context = new StorageMap(prefix);
+        }
+
+        public override void Validate()
+        {
+            var data = _context.Get(_key);
+            ExecutionEngine.Assert(data == null, "Already entered");
+            _context.Put(_key, 1);
+        }
+
+#pragma warning disable CA1816 // Dispose methods should call SuppressFinalize
+        public void Dispose()
+#pragma warning restore CA1816 // Dispose methods should call SuppressFinalize
+        {
+            _context.Delete(_key);
+        }
+    }
+}
--- a/src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
+++ b/src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
@@ -9,6 +9,7 @@
 // modifications are permitted.
 
 using System;
+using System.Runtime.CompilerServices;
 using Neo.SmartContract.Framework.Services;
 
 namespace Neo.SmartContract.Framework.Attributes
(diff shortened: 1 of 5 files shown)


Liaojinghui:
easier for us or easier for contract developer, that is the question.

shargon:
@erikzhang all the `returns` inside a method doesn't ends in `_returnTarget` ?

src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
```diff
@@ -0,0 +1,41 @@
...
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using System;
+using Neo.SmartContract.Framework.Services;
+
+namespace Neo.SmartContract.Framework.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class NoReentrantAttribute : ModifierAttribute, IDisposable
```
erikzhang, line 17: I don't think an attribute should implement `IDisposable`. Instead, we can add `Enter` and `Exit` methods into `ModifierAttribute`.

shargon: Done

shargon:
@erikzhang merge?

src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using Neo.SmartContract.Framework.Services;
+
+namespace Neo.SmartContract.Framework.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class NoReentrantAttribute : ModifierAttribute
+    {
+        private readonly StorageMap _context;
+        private const string _key = "noReentrant";
```
erikzhang, line 20: This will cause the two methods with `NoReentrantAttribute` attached to be unable to call each other.

shargon: We can specify the key in the constructor

erikzhang: Why not generate random keys?

src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Neo.SmartContract.Framework.Attributes
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class NoReentrantAttribute : ModifierAttribute
+    {
+        private readonly StorageMap _context;
+        private readonly string _key;
+
+        public NoReentrantAttribute(byte prefix = 0xFF, string key = "noReentrant")
```
erikzhang, line 22:
```suggestion
        public NoReentrantAttribute(byte prefix = 0xFF, [CallerMemberName]string key = "noReentrant")
```

@shargon Can you test this?

src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
Liaojinghui: why dont you just use the method name and param lengh as the key? what is the point of asking user to assign their own key?

src/Neo.SmartContract.Framework/Attributes/NoReentrantAttribute.cs
shargon: Yes, but it will define only the default value? 
It's good to have the same value for `sell` and `buy` because you should not be able to reentrant between different methods by default

erikzhang: > It's good to have the same value for sell and buy because you should not be able to reentrant between different methods by default

If you provide an argument explicitly, the value you provide will be used.

My concern is that the `CallerMemberNameAttribute` doesn't work properly in the contract. Need to actually test it in the contract.

shargon: > Need to actually test it in the contract.

I tested and it looks good. If someone else can test it should be good, but seems that it works.

devhawk:
Is there an issue tracking the problem this PR is intended to address

shargon:
> Is there an issue tracking the problem this PR is intended to address

No, just for make the code cleaner and readable

devhawk:
> > Is there an issue tracking the problem this PR is intended to address
> 
> No, just for make the code cleaner and readable

Making *contract* code cleaner and readable right? If so, we should have an issue tracking the problem

superboyiii:
> @superboyiii Please, could you test it?

I will

superboyiii (approved):
Tested OK, works well.

Liaojinghui (approved):

shargon:
@erikzhang merge?

Liaojinghui:
@shargon needs udpate.

PR #757, opened 2022-06-23
add notify manner to assert

Liaojinghui:
Allow developers to send notification on faulty assert

### DIFF
--- a/src/Neo.SmartContract.Framework/ExecutionEngine.cs
+++ b/src/Neo.SmartContract.Framework/ExecutionEngine.cs
@@ -31,7 +31,7 @@ namespace Neo.SmartContract.Framework
         {
             if (condition) return;
 
-            if (!isNotify)
+            if (isNotify)
             {
                 Services.Runtime.Notify("Fault", new object[] { message });
                 Abort();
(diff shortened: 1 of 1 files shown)


shargon (commented):
I think that maybe we should use notify always, or create a new method instead of an argument, with the same behavior, Assert(false) instead of Abort

Liaojinghui:
> I think that maybe we should use notify always, or create a new method instead of an argument, with the same behavior, Assert(false) instead of Abort

@erikzhang any suggestion?

erikzhang:
I think `Log` is more reasonable than `Notify`.

Liaojinghui:
> I think `Log` is more reasonable than `Notify`.

@erikzhang so I just keep this pr and allow developers to choose? This pr won't change any existing contract anyway. Can we merge?

Liaojinghui:
> Maybe it's better to add a compilation option in order to choose the type

Any particular reason makes a compilation option better than a default parameter?

shargon:
> @shargon Any particular reason makes a compilation option better than a default parameter?

Gas saving, default parameter require to push the stack item

Liaojinghui:
> Gas saving, default parameter require to push the stack item

got that, may you do an option pr?

PR #766, opened 2022-09-06
add: Runtime.LoadScript

shargon:
Require https://github.com/neo-project/neo/pull/2756

### DIFF
--- a/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_IntegerParse.cs
+++ b/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_IntegerParse.cs
@@ -16,26 +16,22 @@ namespace Neo.Compiler.CSharp.UnitTests
 
             var result = testengine.ExecuteTestCaseStandard(methodname, "127");
             var value = result.Pop().GetInteger();
-            Assert.AreEqual(127, value);
+            Assert.AreEqual(sbyte.MaxValue, value);
 
             testengine.Reset();
             result = testengine.ExecuteTestCaseStandard(methodname, "-128");
 
             value = result.Pop().GetInteger();
-            Assert.AreEqual(-128, value);
+            Assert.AreEqual(sbyte.MinValue, value);
 
             //test backspace trip
             testengine.Reset();
             result = testengine.ExecuteTestCaseStandard(methodname, "20 ");
-
-            value = result.Pop().GetInteger();
-            Assert.AreEqual(20, value);
+            Assert.AreEqual(VMState.FAULT, testengine.State);
 
             testengine.Reset();
             result = testengine.ExecuteTestCaseStandard(methodname, " 20 ");
-
-            value = result.Pop().GetInteger();
-            Assert.AreEqual(20, value);
+            Assert.AreEqual(VMState.FAULT, testengine.State);
 
             testengine.Reset();
             result = testengine.ExecuteTestCaseStandard(methodname, "128");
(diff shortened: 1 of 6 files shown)


ProDog:
Add a return value?

tests/Neo.Compiler.CSharp.UnitTests/Utils/TestEngine.cs
```diff
@@ -162,7 +162,7 @@ public EvaluationStack ExecuteTestCaseStandard(int offset, ushort rvcount, NefFi
                 Console.WriteLine("op:[" +
                     this.CurrentContext.InstructionPointer.ToString("X04") +
                     "]" +
-                this.CurrentContext.CurrentInstruction.OpCode);
+                this.CurrentContext.CurrentInstruction?.OpCode);
```
shargon, line 165: @erikzhang the unique weird thing is that, the current instruction was null after the dynamic script

```diff
    op:[019D] INITSLOT
    op:[01A0] PUSHDATA1
    op:[01A3] CONVERT
    op:[01A5] CONVERT
    op:[01A7] STLOC0
    op:[01A8] LDARG1
    op:[01A9] LDARG0
    op:[01AA] PUSH2
    op:[01AB] PACK
    op:[01AC] PUSH15
    op:[01AD] LDLOC0
    op:[01AE] SYSCALL
    op:[0000] ADD
+    op:[0001]
    op: [01B3] JMP
    op:[01B5] RET
```

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_IntegerParse.cs
```diff
@@ -321,27 +301,22 @@ public void LongParse_Test()
 
             var result = testengine.ExecuteTestCaseStandard(methodname, " -9223372036854775808");
             var value = result.Pop().GetInteger();
-            Assert.AreEqual(-9223372036854775808, value);
+            Assert.AreEqual(long.MinValue, value);
```
shargon, line 304: @erikzhang With the last version, this UT doesn't pass and it should

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_IntegerParse.cs
```diff
@@ -16,26 +16,22 @@ public void SByteParse_Test()
...
             value = result.Pop().GetInteger();
-            Assert.AreEqual(-128, value);
+            Assert.AreEqual(sbyte.MinValue, value);
 
             //test backspace trip
             testengine.Reset();
             result = testengine.ExecuteTestCaseStandard(methodname, "20 ");
-
-            value = result.Pop().GetInteger();
-            Assert.AreEqual(20, value);
```
erikzhang, line 32: Why UTs are modified?

Liaojinghui: In the new "neo 3.4.0-CI01389", `byte.Parse` can not handle string with spaces like ` 20`

Liaojinghui: Becasue of this pr https://github.com/neo-project/neo/commit/5f4c14e3801c691ceccb5705c634e9e55dd3bba1#diff-47e4d66ed629fefce55bd5a2ea5812a8ae48a9633dce8a795f239e525f07568dR105

Liaojinghui: Fixed in https://github.com/neo-project/neo/pull/2813

tests/Neo.Compiler.CSharp.UnitTests/Utils/TestEngine.cs
Ashuaidehao: After execute "ADD", the current instruction point will auto jump to "Next" point even it is an outrange point. ([see here](https://github.com/neo-project/neo-vm/blob/f48de7e595ade7c72be6ccd5cce0ab99708f0afd/src/Neo.VM/ExecutionEngine.cs#L1461))

```CSharp
 internal bool MoveNext()
        {
            Instruction? current = CurrentInstruction;
            if (current is null) return false;
            InstructionPointer += current.Size;
            return InstructionPointer < Script.Length;
        }
```
Dynamic Script length is 1, InstructionPointer  +1, InstructionPointer  will be 1.

In the next iteration, the out-range point will get a "null" instruction.
By the way, the vm will auto change "null" to "RET" instruction, so it won't be fault in vm(https://github.com/neo-project/neo-vm/blob/f48de7e595ade7c72be6ccd5cce0ab99708f0afd/src/Neo.VM/ExecutionEngine.cs#L1450).

shargon:
> Merge?

It's ok for ne

vncoelho (approved):

PR #772, opened 2022-12-16
set the default context as current context.

Liaojinghui:
Turns out Storage does not need specifically set the context.


src/Neo.SmartContract.Framework/Services/Storage.cs
```diff
@@ -98,5 +98,18 @@ public static class Storage
         /// </summary>
         [Syscall("System.Storage.Find")]
         public static extern Iterator Find(StorageContext context, byte[] prefix, FindOptions options = FindOptions.None);
+
+#pragma Interface with default Context
+        public static ByteString Get(ByteString key) => Get(CurrentContext, key);
```
shargon, line 103: Use readonly for the read only ones?

Liaojinghui: Good suggestion.

shargon (approved):

Liaojinghui:
This one is ready to merge @shargon

Issue #780, opened 2023-03-31
Incorret NEP17 example


lock9:
Hi, 
The NEP-17 example is incorrect and will likely break (unless it was updated recently):
https://github.com/neo-project/neo-devpack-dotnet/blob/master/src/Neo.SmartContract.Framework/Nep17Token.cs

The following lines won't work:
```C#
protected static void PostTransfer(UInt160 from, UInt160 to, BigInteger amount, object data)
        {
            OnTransfer(from, to, amount);
            if (to is not null && ContractManagement.GetContract(to) is not null)
                Contract.Call(to, "onNEP17Payment", CallFlags.All, from, amount, data);
        }
```

We need to revert the hex before calling the GetContract and Contract.Call.

Is there any chance we can remove the need for reverting hex when invoking / calling other contracts? I don't even know how to document this. What's the justification? Can someone help me explain this requirement to a new developer?

Liaojinghui:
I am quite confused by your question, this has been adopted by many projects, I don't see any issue here. May I ask why would you need to revert the hex? Is that possible to provide more context?

lock9:
When calling external contracts, we had to revert the scripthash before using Contract.Call. At least this was happening in my local contract with Neo Express.  The opcode itself expects a reversed hex, but the compiler may auto revert it.
I'm testing the same contract in different programming language, maybe the compiler that I'm using is not converting it

roman-khimov:
The example is correct, it's just neo-project/neo#938 and it can't be fixed now.

Liaojinghui:
> The example is correct, it's just [neo-project/neo#938](https://github.com/neo-project/neo/issues/938) and it can't be fixed now.

I am always impressed by how many issues you can refer to, even dating back many years ago.

lock9:
> I've had dreams about a NEO4 where we deleted the `reverse()` call, removed the `0x` prefix and just called it `Hash160/256` (to nicely align with the ABI parameter names), it was amazing 😄 . This to me has always felt like NEOs equivalent of Tony Hoare's [null reference mistake](https://en.wikipedia.org/wiki/Tony_Hoare#cite_note-26).

Haha. I have similar dreams 😂 

Guys, I'm still confused. When should we use big endian or little endian?  

Example:
```
PUSHINT64 c800ad4787010000
PUSHDATA1 20f0bea450ada7b903b89749d7c9bbc160b148cd
[...]

PR #782, opened 2023-04-02
show error if given contact is abstract.

Liaojinghui:
Fix this https://github.com/neo-project/neo-devpack-dotnet/issues/780

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -149,11 +149,7 @@ namespace Neo.Compiler
 
         internal static CompilationContext Compile(IEnumerable<string> sourceFiles, IEnumerable<MetadataReference> references, Options options)
         {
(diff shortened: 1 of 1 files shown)


src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -149,6 +149,12 @@ private void Compile()
 
         internal static CompilationContext Compile(IEnumerable<string> sourceFiles, IEnumerable<MetadataReference> references, Options options)
         {
+            if (IsSingleAbstractClass(sourceFiles))
+            {
+                Console.Error.WriteLine("The source file does not contain a valid neo Smart Contract");
+                Environment.Exit(-2);
```
shargon, line 155: We don't have a list of error codes, I think that we should throw the exception

shargon (approved):

Liaojinghui:
@shargon

PR #793, opened 2023-09-15
Update Neo

shargon:
Implement https://github.com/neo-project/neo/issues/2851 and https://github.com/neo-project/neo/pull/2827

### DIFF
--- a/src/Neo.SmartContract.Framework/Native/StdLib.cs
+++ b/src/Neo.SmartContract.Framework/Native/StdLib.cs
@@ -73,6 +73,16 @@ namespace Neo.SmartContract.Framework.Native
 
         public static extern string[] StringSplit(string str, string separator, bool removeEmptyEntries);
 
+        /// <summary>
+        /// Get the string length by elements
+        /// </summary>
+        /// <param name="str"></param>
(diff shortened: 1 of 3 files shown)


Liaojinghui (approved):

src/Neo.SmartContract.Framework/Native/StdLib.cs
```diff
@@ -72,5 +72,7 @@ public static class StdLib
         public static extern string[] StringSplit(string str, string separator);
 
         public static extern string[] StringSplit(string str, string separator, bool removeEmptyEntries);
+
+        public static extern int StrLen(string str);
```
Liaojinghui, line 76:
```suggestion

        /// <summary>
        /// Get the string length by elements
        /// </summary>
        /// <param name="str"></param>
        /// <returns>Number of elements in the string</returns>
        /// <example>
        ///        string a = "A"; // return 1
        ///        string tilde = "ã"; // return 1
        ///        string duck = "🦆"; //return 1
        /// </example>
        public static extern int StrLen(string str);
```

Liaojinghui (approved):

PR #796, opened 2023-10-03
Regex: add couple of regex methods to  the devpack

Jim8y:
Replace https://github.com/neo-project/neo/pull/2927

As is discussed with  @roman-khimov and @shargon, now implement regex in the devpack using the `StdLib.MemorySearch`.


src/Neo.SmartContract.Framework/Regex.cs
```diff
@@ -0,0 +1,64 @@
+using Neo.SmartContract.Framework.Native;
+namespace Neo.SmartContract.Framework;
+
+public class Regex
+{
+    public static bool NumberOnly(ByteString byteString)
```
cschuchardt88, line 6: Should add these ones too https://github.com/neo-project/non-native-contracts/blob/8d72b92e5e5705d763232bcc24784ced0fb8fc87/src/NameService/NameService.cs#L531-L652
minus `SplitAndCheck`.

Liaojinghui: Lets wait for @shargon 's opinion

shargon: I think that `isAlpha` and `isNum` are included already

cschuchardt88: `CheckIPv4` `CheckIPv6` would be a nice addition. But maybe should be rename to something else.

shargon: Is smart contract related, not generic

cschuchardt88:
After thinking about it, Wouldn't it be better to add native methods to `neo` that calls methods like `char.IsAscii` and `char.IsNumber` so this way we can add all the `string` and `char` static methods from dotnet?

Liaojinghui:
> After thinking about it, Wouldn't it be better to add native methods to `neo` that calls methods like `char.IsAscii` and `char.IsNumber` so this way we can add all the `string` and `char` static methods from dotnet?

After we have native contact update.

Liaojinghui:
@shargon please review

src/Neo.SmartContract.Framework/ByteString.Extension.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </summary>
+    /// <param name="byteString">Array where to search.</param>
+    /// <param name="byteToFind">Array to search.</param>
+    /// <returns>True if ends with</returns>
+    public static bool EndsWith(this ByteString byteString, ByteString byteToFind)
+    {
+        return StdLib.MemorySearch(byteString, byteToFind) + byteToFind.Length == byteString.Length;
+    }
+
+    public static bool Contains(this ByteString byteString, ByteString byteToFind)
```
shargon, line 99: Summary?

src/Neo.SmartContract.Framework/ByteString.Extension.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace Neo.SmartContract.Framework;
+
+public static class ByteStringExtension
+{
+    /// <summary>
+    ///  Denotes whether provided character is a number.
+    /// </summary>
+    /// <param name="byteString">Input to check</param>
+    /// <returns>True if is number</returns>
+    public static bool IsNumber(this ByteString byteString)
```
shargon, line 11: It's also work for "string" type? (it could be in a different pr)

Jim8y: what? anything can work for string as well. But the problem is some methods string does not support some string does. 

Type system in neo is a mess. That is why i suggested to use neo type only.

Jim8y:
now all addressed.

src/Neo.SmartContract.Framework/ByteString.Extension.cs
shargon: > Type system in neo is a mess. 

We need to fix this in a future, but force users to use c# with other types is worst

shargon (approved):

src/Neo.SmartContract.Framework/ByteString.Extension.cs
Jim8y: > We need to fix this in a future, but force users to use c# with other types is worst

I dont agree, C# still is not a smart contract language, forcing it to work for smart contract without any restriction is impraactical and misleading. For me, having a determinastic type system is far more importand than having a system that sometimes work sometimes dont work and you have to teach them when it works and when it does not work.

Again, worst is having a system that sometimes it works, sometimes it does not work.

Issue #797, opened 2023-10-05
NCCS: IOException with invalid directory


cschuchardt88:
NCCS crashes on invalid directory path See the following Exception log. You need to check if the path is valid and/or catch the exception and display a message indicating the problem.

# Exception Log
```
Target PostBuild:
1>  Task "Exec"
1>    dotnet nccs "CallContract.csproj" -o "C:\Users\Chris\Projects\cschuchardt88\neo-examples-csharp\\sc\"
1>      Determining projects to restore...
1>      All projects are up-to-date for restore.
1>    Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
1>     ---> System.IO.IOException: The filename, directory name, or volume label syntax is incorrect. : 'C:\Users\Chris\Projects\cschuchardt88\neo-examples-csharp\sc"'
1>       at System.IO.FileSystem.CreateDirectory(String fullPath, Byte[] securityDescriptor)
1>       at System.IO.Directory.CreateDirectory(String path)
1>       at Neo.Compiler.Program.ProcessOutputs(Options options, String folder, CompilationContext context) in /home/runner/work/neo-devpack-dotnet/neo-devpack-dotnet/src/Neo.Compiler.CSharp/Program.cs:line 140
1>       at Neo.Compiler.Program.ProcessCsproj(Options options, String path) in /home/runner/work/neo-devpack-dotnet/neo-devpack-dotnet/src/Neo.Compiler.CSharp/Program.cs:line 119
1>       at Neo.Compiler.Program.Handle(RootCommand command, Options options, String[] paths, InvocationContext context) in /home/runner/work/neo-devpack-dotnet/neo-devpack-dotnet/src/Neo.Compiler.CSharp/Program.cs:line 75
1>       at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
1>       at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
1>       --- End of inner exception stack trace ---
1>       at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
1>       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
1>       at System.Delegate.DynamicInvokeImpl(Object[] args)
[...]

PR #807, opened 2023-10-19
Capture output error

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/797


cschuchardt88:
I think if  you just return the exception message that should be good enough. But looks good.

Liaojinghui (approved):

PR #811, opened 2023-11-06
fix char to bigiteger

### LINKED ISSUE #810: nccs 3.6 does not support char to biginteger conversion.
```csharp
var typeName = (ByteString)((BigInteger)'A' + secondType);
```

The above code compiles successfully with nccs 3.5 and earlier, but with 3.6 it reports an error

Unknown method: System.Numerics.BigInteger.implicit operator System.Numerics.BigInteger(char)

nccs 3.5.0

image

nccs 3.6.0

image

Liaojinghui:
close #810


Liaojinghui:
@chenzhitong

shargon (approved):

chenzhitong:
Compilation completed successfully. 👍

PR #818, opened 2023-11-15
split ExpressionConvert

Liaojinghui:
This is the first step of breaking the very large methodconvert file.

Not a single line of the code logic is changed.


shargon:
> Not a single line of the code logic is changed.

I keep thinking how I can verify that nothing has been modified, do you know any tool?

Liaojinghui:
Replace with #820

Issue #816, opened 2023-11-15
Break Method Convert


Liaojinghui:
As is in https://github.com/neo-project/neo-devpack-dotnet/pull/813, the methodconvert file is too large to maintain, we need to break it into smaller files.

This issue will trace the breaking prs. Breaking process plan:
- [ ] ExpressionConvert https://github.com/neo-project/neo-devpack-dotnet/pull/818
- [ ] PatternConvert
- [ ] StatementConvert
- [ ] SystemConvert
- [ ] ConstructorConvert
- [ ] Adding comments to ExpressionConvert
- [ ] Adding comments to PatternConvert
- [ ] Adding comments to StatementConvert
- [ ] Adding comments to SystemConvert
- [ ] Adding comments to ConstructorConvert
- [ ] Adding tests to ExpressionConvert
- [ ] Adding tests to PatternConvert
- [ ] Adding tests to StatementConvert
- [ ] Adding tests to SystemConvert
- [ ] Adding tests to ConstructorConvert

shargon:
Could you add unit tests at same time that is moved?

Liaojinghui:
Work all at once would only cause two situations, either very large pr again, or very surface unit tests. Unit tests for this has being missing for years, adding them back should be another task.

shargon:
You say that splitting this method into several files is necessary to maintain it better and improve the coverage, I think that if you split it up it has to be to add that coverage, otherwise, we don't improve anything. In addition, this way we can ensure that this code is correct.

Liaojinghui:
@shargon step by step. How about we focus on fixing issues? Large file is hard to follow and maintain, how about we break it into smaller files step by step, then  we work on checking C# language specification to find what is missing, then unit tests to ensure everything works fine, and then comments to make sure others can understand?

Liaojinghui:
Will directly work on the original methodconvert file

PR #820, opened 2023-11-17
break the convert with #region

Liaojinghui:
Alternative to #816, now with #region to break the convert

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert.cs
@@ -34,6 +34,9 @@ namespace Neo.Compiler
 {
     class MethodConvert
     {
+
+        #region Fields
+
         private readonly CompilationContext context;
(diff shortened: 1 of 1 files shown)


Liaojinghui:
@shargon how about this one, this one dont even touch the file itself. Just #region and #endregion

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -3368,16 +3412,11 @@ private void ConvertNotPattern(SemanticModel model, UnaryPatternSyntax pattern,
             AddInstruction(OpCode.NOT);
         }
 
-        private void ConvertTupleExpression(SemanticModel model, TupleExpressionSyntax expression)
```
Liaojinghui, line 3371: This is the only place where i changed the method order cause it is an expression, supposed to stay together with other expressions.

shargon (approved):

PR #821, opened 2023-11-19
Add comments to Compiler

Jim8y:
This pr will focus on adding comments to the method convert class of the compiler. It focuses on:

- Adding comments to existing code.
- Adding comments to the syntex that compiler does not support.
- Adding examples of syntex that compiler support.
- Update variable and method names (only a few) that are not clear not not standard style.
-  Update the exception to make it align to the diagnosticID


Liaojinghui:
@shargon I will mark this one as draft since I will keep updating to this big file until I add very detailed comments. But before I finish, you can check existing comments from now.

Fundamental rule of this pr is, not a single line code logic will be changed. But variable and method names may be updated.

PR #826, opened 2023-12-10
Fix: ut test compile error

Liaojinghui:
print out the ut test compile error


shargon (approved):

shargon:
I suffer it a lot of times 😆

Liaojinghui:
> I suffer it a lot of times 😆

we share the same pain, and thank you for merging this.

PR #829, opened 2023-12-11
Fix: fix name style

Jim8y:
This pr fix the naming style.

The name of properties and variables is a mess, use the recommended C# style instead.


Jim8y:
Close as we can make this done with editorstyle in the future

PR #832, opened 2023-12-14
Fix: fix bug in the compiler version

Liaojinghui:
This pr fix the bug that in mac platform, the compiler can get a version attached with commit number, which will make the compiler version longer than 64.


Liaojinghui:
actually casued by dotnet 8.0

Issue #834, opened 2023-12-14
Add support to Math.Pow


lock9:
Problem:
The Decimals property "can't" be used inside smart contracts to deal with the supply. 
Example: It's not possible to convert 100_000_000  to 100_000_000.00000000. 

Solution:
Add support to Math.Pow (or equivalent).
Python Example:
```python
@public(name="totalSupply", safe=True)
def total_supply() -> int:
    return 100000000 * 10 ** decimals() # 100,000,000.00000000
```

Note: I'm not sure if there is another solution for this.

Question: How do people calculate the total supply without using the decimals property? (inside the smart contract).

Liaojinghui:
BigInteger.Pow

lock9:
Would it be hard to add Math.Pow and make it work like BigInteger.Pow? 
Identical behavior but allowing users to also use Math.Pow?

cschuchardt88:
Can't you use BigInteger in python? Would be like this:
```csharp
BigInteger.Pow(100_000_000, 8); // (value, exponent)
```

Liaojinghui:
> Would it be hard to add Math.Pow and make it work like BigInteger.Pow? 
> 
> Identical behavior, just so it's easier to use

Well, not hard at all, just a few lines of code. I am working to make the method support more complete, but will take time to get merged.

lock9:
Sure, no rush. Thank you for looking into it

Jim8y:
can not add Math.Pow as it requires `double` type, which we dont support, keep using BigInteger.Pow

PR #836, opened 2023-12-15
Add: add math.pow

Jim8y:
Closes https://github.com/neo-project/neo-devpack-dotnet/issues/834


Liaojinghui:
@lock9 please check

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -4182,6 +4182,7 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                     AddInstruction(OpCode.SIGN);
                     return true;
                 case "System.Numerics.BigInteger.Pow(System.Numerics.BigInteger, int)":
+                case "System.Math.Pow(double, double)":
```
shargon, line 4185: I think that we haven't got double or float compatibility

Liaojinghui: yes and no. yes it uses double and neo does not support double. no it is c#, and it compiles and works well. truth is, we can not prevent developer from using double at all, it actually works in neo, c# convert them implecitly into int or other types, we cant prevent it, we cant detect it, we even 'support' it.

Liaojinghui: i will do some study on this, maybe ban those key words from source level.

Liaojinghui:
@shargon  now this can be implemented with https://github.com/neo-project/neo-devpack-dotnet/pull/839. the use of methods that take double or return double will be convertd to long by showing error highlight and suggestions.

Liaojinghui:
Now we can get ride of `double`

src/Neo.SmartContract.Framework/Math.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace Neo.SmartContract.Framework
+{
+    public class Math
+    {
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(BigInteger x, BigInteger y);
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(long x, long y);
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(int x, int y);
```
shargon, line 23:
```suggestion
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(long x, long y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(int x, int y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(ulong x, ulong y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(uint x, uint y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(short x, short y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(ushort x, ushort y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(byte x, byte y);
```

src/Neo.SmartContract.Framework/Math.cs
```diff
@@ -0,0 +1,91 @@
...
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(ulong x, ulong y);
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(uint x, uint y);
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(short x, short y);
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(ushort x, ushort y);
+        [OpCode(OpCode.POW)]
+        public static extern BigInteger Pow(byte x, byte y);
```
shargon, line 33:
```suggestion
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(byte x, byte y);
        [OpCode(OpCode.POW)]
        public static extern BigInteger Pow(sbyte x, sbyte y);
```

Jim8y:
@dotnet-policy-service agree

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -4513,44 +4513,16 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                         PrepareArgumentsForMethod(model, symbol, arguments);
                     Call(NativeContract.StdLib.Hash, "atoi", 1, true);
                     return true;
-                case "System.Math.Abs(sbyte)":
-                case "System.Math.Abs(short)":
-                case "System.Math.Abs(int)":
-                case "System.Math.Abs(long)":
```
shargon, line 4519: Why remove this compatibility?

Jim8y: to avoid confusion, use `Neo.Math` everywhere instead of `System.Math`. Otherwise you will have two `Math` system.

cschuchardt88: I prefer to use native `C#`. This may why developers can't program in `C#` _(I guess_). No internet search will tell them to use `Neo.math`, unless the `NC` error says to.

shargon: Agree with @cschuchardt88

Jim8y: You can't, many of them use `double`, i think you know what the issue is @shargon.
And I do not think using C# native is a good idea, this is neo C#, instead of dotnet C#, using dotnet C# causes confusing as people do not know what can be done and what can not.

Jim8y: Where do you know what method is supported by neo @cschuchardt88?  only at compile time right? how does it feel? every time you use something, you know if it works only if you compile it......

cschuchardt88: Understandable, Maybe we should phase out all native `dotnet` `APIs` and only have `Neo` `API` calls. For example, `Neo.Math.Abs`, `Neo.String` and etc

Jim8y: Exactly what i have in mind.

src/Neo.Compiler.CSharp/MethodConvert.cs
```diff
@@ -4513,44 +4513,16 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
...
                     AddInstruction(OpCode.MAX);
                     return true;
-                case "System.Math.Min(byte, byte)":
-                case "System.Math.Min(sbyte, sbyte)":
-                case "System.Math.Min(short, short)":
-                case "System.Math.Min(ushort, ushort)":
-                case "System.Math.Min(int, int)":
-                case "System.Math.Min(uint, uint)":
-                case "System.Math.Min(long, long)":
-                case "System.Math.Min(ulong, ulong)":
```
vncoelho, line 4553: I think this part should be in another PR, @Jim8y .

It is a good point to discuss

Jim8y: Sure, will move it to another pr.

Jim8y:
now this pr only adds Math methods

PR #838, opened 2023-12-16
Update action steps and fixed myget

cschuchardt88:
@shargon Sorry took so long.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,7 @@ name: .NET Core Test and Publish
 
 on:
   push:
-    branches: master
+    branches: [master]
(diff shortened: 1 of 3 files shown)


cschuchardt88:
Reason for `global.json` is because on my computer and on `github` _see below_. This will fix `dotnet` when you compile, targeting the right version of the sdk. Thats why your `tests` will fail if you have `dotnet` 8.0 installed.

image

shargon (approved):

shargon:
Same error...

cschuchardt88:
Must be your `myget` account. Something changed. You will have to contact support than.

PR #841, opened 2023-12-19
Add: make storage attribute also work

Jim8y:
This pr makes storagebacked and storage attributes works the same.


shargon (commented):
I think that is better to define only one name, `StoredAttribute`?

Liaojinghui:
> I think that is better to define only one name, `StoredAttribute`?

I just feel storagebacked is too long and unnecessary. I would support any name shorter than that.

shargon (approved):

PR #842, opened 2023-12-20
Fix: Code Reuse.

Jim8y:
This pr trys to shrink the size of the methodconvert.

This pr saves 400 lines of code in methodconvert.

1. reuse the existing code. 
3. move all integer system methods into MerhodConvert.Integer


Jim8y:
Close as we will break the convert into seperate sections

Issue #843, opened 2023-12-20
Remove `Runtime.Notify`


cschuchardt88:
You should just remove `Runtime.Notify` its useless now. Or have it add the name to contract manifest.

Get error `Inner InvalidOperationException: Event 'SayHello3' does not exist.`

### Contract
```csharp
public class EventContract : SmartContract
{
    public delegate void OnSayHelloDelegate(string message);

    [DisplayName("SayHello1")]
    public static event OnSayHelloDelegate OnSayHello;

    [DisplayName("SayHello2")]
    public static event Action<string> SayHello;

    public static void Main()
    {
        OnSayHello("Hello, alice");

        SayHello("Hello, bob");

        Runtime.Notify("SayHello3", new[] { "Hello, joe" }); // crash
    }
}
```

vncoelho:
Perhaps that is right to do

Jim8y:
@shargon how do you think? I think this is a bug maybe, i have asked Erik for his opinion, but get no response.
maybe we should vote to make a decision? @vncoelho @roman-khimov

Jim8y:
here https://github.com/neo-project/neo/issues/1879 as proposed by @roman-khimov  and completed by @erikzhang . But the problem is we allow user to use it in the syntax level, but not allowing it in the contract level. An annoying step to increase the bar of contract development.

Jim8y:
![Screenshot 2023-12-23 at 12 11 45 PM](https://github.com/neo-project/neo-devpack-dotnet/assets/10189511/224dec0a-cf0c-462d-b7a1-09f5e22eaab6)

Now updated the contract analyzer to check the `notify` name with id NC4019

cschuchardt88:
What if i did or want to do:
```csharp
public static void SayHelloTo(string eventName, string name)
{
    Runtime.Notify(eventName, new[] { $"Hello, {name}" });
}
```

Also what about the parameters? Do you check though too against manifest?

Jim8y:
can be added.

shargon:
I think that is a bug, seems good

cschuchardt88:
@Jim8y Why not have `Runtime.Notify` event name be added to the manifest if it doesn't exists in the contract? Because you have to compile the contract in order to `deploy` or `update`.

roman-khimov:
> Runtime.Notify event name be added to the manifest if it doesn't exist in the contract?

[...]

PR #839, opened 2023-12-17
Add: add neo smart contract syntax analyzer

Jim8y:
This pr adds the syntax rules of the neo smart contract.

It:

- Bans the use of decimal type
- Bans the use of double type
- Bans unsupported BigInteger, String, Char, Math methods and syntax.
- Bans out, ref keywords.
- Bans the use of LINQ.
- Bans unsupported collection types.

All baned syntax will show errors in the IDE.

Code Fix:

double => long;
float => int;
out => local variable and return value;
anonymous function => methods;

PR #845, opened 2023-12-23
Fix: ban  notify from contract developer

Jim8y:
Closes https://github.com/neo-project/neo-devpack-dotnet/issues/843

### DIFF
--- a/src/Neo.SmartContract.Framework/Services/Runtime.cs
+++ b/src/Neo.SmartContract.Framework/Services/Runtime.cs
@@ -95,6 +95,7 @@ namespace Neo.SmartContract.Framework.Services
         [Syscall("System.Runtime.Log")]
         public static extern void Log(string message);
 
+        // Events not present in the ABI were disabled in HF_Basilisk
         // [Syscall("System.Runtime.Notify")]
         // public static extern void Notify(string eventName, object[] state);
 
--- a/src/Neo.SmartContract.Framework/Services/Runtime.cs
+++ b/src/Neo.SmartContract.Framework/Services/Runtime.cs
@@ -95,8 +95,8 @@ namespace Neo.SmartContract.Framework.Services
         [Syscall("System.Runtime.Log")]
         public static extern void Log(string message);
 
-        [Syscall("System.Runtime.Notify")]
-        public static extern void Notify(string eventName, object[] state);
+        // [Syscall("System.Runtime.Notify")]
+        // public static extern void Notify(string eventName, object[] state);
 
         [Syscall("System.Runtime.BurnGas")]
         public static extern void BurnGas(long gas);
--- a/tests/Neo.SmartContract.Framework.UnitTests/SyscallTest.cs
+++ b/tests/Neo.SmartContract.Framework.UnitTests/SyscallTest.cs
@@ -81,7 +81,7 @@ namespace Neo.SmartContract.Framework.UnitTests
             HashSet<string> fwSyscalls = visitor.Visit(compilation.GlobalNamespace).ToHashSet();
 
             fwSyscalls.SymmetricExceptWith(neoSyscalls);
-            if (fwSyscalls.Count > 0)
+            if (fwSyscalls.Count > 0 && fwSyscalls.All(p => !p.Equals("System.Runtime.Notify")))
             {
                 Assert.Fail($"Unknown or unimplemented syscalls: {string.Join("\n", fwSyscalls)}");
             }


Jim8y:
@shargon @cschuchardt88 @roman-khimov @vncoelho @superboyiii please check

roman-khimov:
You can't write any meaningful contract with this PR merged. I'm not sure what's the problem really, just have a proper manifest (_how_ you create it is a separate matter) for your contract and that's it.

Jim8y:
> You can't write any meaningful contract with this PR merged. I'm not sure what's the problem really, just have a proper manifest (_how_ you create it is a separate matter) for your contract and that's it.

What problem are you specifying? user can still call Notify via event, existing contract will still work.

shargon:
> > You can't write any meaningful contract with this PR merged. I'm not sure what's the problem really, just have a proper manifest (_how_ you create it is a separate matter) for your contract and that's it.
> 
> What problem are you specifying? user can still call Notify via event, existing contract will still work. The only thing that I banned is directly calling Notify without defining an event...... cause it will fail the execution for sure anyway.
> 
> Well, this actually only affects the C# compiler,,,, the core remains the same.

But is not possible to call a dynamic event

Jim8y:
> But is not possible to call a dynamic event

what you mean dynamic event? can you please provide an example?

roman-khimov:
@Jim8y, maybe I just don't know enough about C# devpack, if `event` wrapper allows to emit notifications without direct `Runtime.Notify()` calls in the code, then it's not _that_ bad. Still somewhat limiting to me. I can have a pair of create/delete events with the same parameters and emitting them with `Runtime.Notify(nameVariable, params)` might be convenient.

Jim8y:
> I can have a pair of create/delete events with the same parameters and emitting them with `Runtime.Notify(nameVariable, params)` might be convenient.

But current `Notify` checking logic has banned from using arbitary `nameVariable`.  you must defined them as 

```C#
    public delegate void OnSayHelloDelegate(string message);

    [DisplayName("SayHello1")]
    public static event OnSayHelloDelegate OnSayHello;

    [DisplayName("SayHello2")]
    public static event Action<string> SayHello;
```

but if you already defined them this way, using `Notify` would be unecessary, yet keeping it makes it misunderstanding as it can take arbitary name without syntax error while actually arbitary name will fail the execution.

shargon:
> > But is not possible to call a dynamic event
> 
> Current core only accept events that are defined as `event`, what you mean `dynamic event`, i have never heard of this terminology in neo? can you please provide an example?

I mean that maybe you want to emit a event with a dynamic name, who knows..

I think we should not remove notify because if you want to call it by syscall you will be able to do it, so in reality you are not banning it, you are making it difficult

cschuchardt88:
@shargon 
make sense, but it's very misleading when it compiles when it doesn't check contract manifest of parameters for validation. Putting any name in the input of `Runtime.Notify` without the name in manifest will error at execution of the invoke of the contract.

shargon:
> @shargon Makes sense, but it's very misleading when it compiles when it doesn't check contract manifest or parameters for validation. Putting any name in the input of `Runtime.Notify` without the name in manifest will error at execution of the invoke of the contract.

But currently the core allow to call notify with a event that is not present in the manifest....

cschuchardt88:
And old contracts will still work fine. You'll just have to do it with `opcode` at low level. But at high level it should be removed.

shargon:
> Since we don't support events when they don't exist in manifest.

We do

cschuchardt88:
Do you at remember last update or so we remove this functionally. Because of NEP-14

`Inner InvalidOperationException: Event 'SayHello3' does not exist.`

cschuchardt88:
@shargon see https://github.com/neo-project/neo/pull/2810

Jim8y:
> I mean that maybe you want to emit a event with a dynamic name, who knows..
> 
> I think we should not remove notify because if you want to call it by syscall you will be able to do it, so in reality you are not banning it, you are making it difficult

It is not an issue of allowing user to call it or not, if i did not get you wrong, your so called `dynamic notify` is already banned by Erik in the core, just check the code. Yes, you can make it easier to use Notify, but that will not work at all, it will just fail the execution by all means. 

**`Dynamic Notify/Event` is BANNED BY THE CORE**, which is proposed by roman, and implemented by erik.

Jim8y:
> > Since we don't support events when they don't exist in manifest.
> 
> We do

We dont, new logic since the hardfork.

shargon:
> > > Since we don't support events when they don't exist in manifest.
> > 
> > 
> > We do
> 
> @shargon We dont, new logic since the hardfork. Just check the issue and code please https://github.com/neo-project/neo/blob/c78ac5adbd790b1fc60d3ea9abc87cf3eee74a6e/src/Neo/SmartContract/ApplicationEngine.Runtime.cs#L352

Yes, you are right...

src/Neo.SmartContract.Framework/Services/Runtime.cs
```diff
@@ -95,8 +95,8 @@ public static class Runtime
         [Syscall("System.Runtime.Log")]
         public static extern void Log(string message);
 
-        [Syscall("System.Runtime.Notify")]
-        public static extern void Notify(string eventName, object[] state);
+        // [Syscall("System.Runtime.Notify")]
+        // public static extern void Notify(string eventName, object[] state);
```
shargon, line 99:
```suggestion
        [Obsolete]
        [Syscall("System.Runtime.Notify")]
        public static extern void Notify(string eventName, object[] state);
```

roman-khimov:
> Dynamic Notify/Event is BANNED BY THE CORE,

Depends on your definition of "dynamic". If it's "I want to emit anything regardless of manifest" then yes. How is it related to `Runtime.Notify` though? It's just an interop call, it doesn't care. Banning an interop from devpack solves nothing to me. In most cases it shouldn't be used, but I think I still can write some code with it that will work (make event definitions for manifest, but emit via `Runtime.Notify`).

cschuchardt88:
@roman-khimov
Yes you are correct. But It isn't intended to be used in this new way. It was for dynamic events **only** _(all the logic for it in nccs)_. Also alot of contract developers use `Runtime.Notify`, and I think its do to not knowing how to program in `C#` or the outdated document on `neo.org` or even it was easier to emit events then adding an extra 3-4 lines.

Jim8y:
@roman-khimov truth is even shargon and i did not know this, are you expecting developers to use it correctly? how would you feel to use an interface that accepts arbitrary input without emiting any errors during compile but actually will fail the execution without saying why? you love that?

cschuchardt88:
>can not find assistence from anywhere?

Maybe should we should add `xml` code comments to `nccs`. I don't why there isn't any to begin with.

Jim8y:
> > can not find assistence from anywhere?
> 
> Maybe should we should add `xml` code comments to `nccs`. I don't why there isn't any to begin with.

Very general question is, how would developer know they can not do this if they don't already know they can not do this?

src/Neo.SmartContract.Framework/Services/Runtime.cs
Jim8y: Obsolete? it is not Obsoleted,,,,better to do in the analyzer as I have updated in https://github.com/neo-project/neo-devpack-dotnet/pull/839

roman-khimov:
@cschuchardt88, @Jim8y, I see it in a different way. Under the hood it's all `Runtime.Notify` anyway. There is no way to escape calling it in the resulting bytecode. It's just a lower layer of abstraction than `event` magic provided by devpack (that does _both_ `Runtime.Notify` and manifest adjustment).

Let's try to approach it from a different angle, how can I make a `Runtime.Notify` invocation in my code without this wrapper and without a proper `event`?

Jim8y:
> @cschuchardt88, @Jim8y, I see it in a different way. Under the hood it's all `Runtime.Notify` anyway. There is no way to escape calling it in the resulting bytecode. It's just a lower layer of abstraction than `event` magic provided by devpack (that does _both_ `Runtime.Notify` and manifest adjustment).
> 
> Let's try to approach it from a different angle, how can I make a `Runtime.Notify` invocation in my code without this wrapper and without a proper `event`?

Have some mercy to the neo contract developer, dont make it even harder to write neo contract. make every interface clear and easy to use please.

cschuchardt88:
>register event to the manifest with whatever names developer pass in

This won't work at `runtime`, If the following is done:
```csharp
public static void HelloWorld(string eventName, string name)
{
    Runtime.Notify(eventName, new[] { $"Hello, {name}" });
}
```

Jim8y:
> > register event to the manifest with whatever names developer pass in
> 
> This won't work at `runtime`, If the following is done:
> 
> ```cs
> public static void HelloWorld(string eventName, string name)
> {
>     Runtime.Notify(eventName, new[] { $"Hello, {name}" });
> }
> ```
> 
> Which leaves us at the same problem to begin with.

Have you tried it in that pr? This syntax will not pass the compile as it is not a constant string.

cschuchardt88:
>Have you tried it in that pr? This syntax will not pass the compile as it is not a constant string. When I say whatever the name, the name must be known at compile time.

Which PR? but i think _below_ will give you `const`
```csharp
Runtime.Notify($"{eventName}", new[] { $"Hello, {name}" });
```

Jim8y:
https://github.com/neo-project/neo-devpack-dotnet/pull/846 This one

vncoelho:
I am fine with banning it, in fact its directly use has been a burden after events.

vncoelho:
I am fine with banning it, in fact its directly use has been a burden after events.
Banning it can be like this proposal, just remove from the devpack.

src/Neo.SmartContract.Framework/Services/Runtime.cs
```diff
@@ -95,8 +95,8 @@ public static class Runtime
         [Syscall("System.Runtime.Log")]
         public static extern void Log(string message);
 
-        [Syscall("System.Runtime.Notify")]
-        public static extern void Notify(string eventName, object[] state);
+        // [Syscall("System.Runtime.Notify")]
```
shargon, line 98:
```suggestion
        // Events not present in the ABI were disabled in HF_Basilisk
        // [Syscall("System.Runtime.Notify")]
```

shargon (approved):

Issue #849, opened 2023-12-26
StatementConvert


Jim8y:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.statementsyntax?view=roslyn-dotnet-4.7.0

- [x] Microsoft.CodeAnalysis.CSharp.Syntax.BlockSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.BreakStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.CheckedStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.CommonForEachStatementSyntax
    - [x] Microsoft.CodeAnalysis.CSharp.Syntax.ForEachStatementSyntax
    - [x] Microsoft.CodeAnalysis.CSharp.Syntax.ForEachVariableStatementSyntax  
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.ContinueStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.DoStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.EmptyStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.ExpressionStatementSyntax
- [ ] Microsoft.CodeAnalysis.CSharp.Syntax.FixedStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.ForStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.GotoStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.IfStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.LabeledStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.LocalDeclarationStatementSyntax
- [ ] Microsoft.CodeAnalysis.CSharp.Syntax.LocalFunctionStatementSyntax
- [ ] Microsoft.CodeAnalysis.CSharp.Syntax.LockStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.ReturnStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.SwitchStatementSyntax
- [x] Microsoft.CodeAnalysis.CSharp.Syntax.ThrowStatementSyntax
[...]

PR #850, opened 2023-12-26
Converter: only break statement for the ease of review

Jim8y:
Closes https://github.com/neo-project/neo-devpack-dotnet/issues/849


shargon:
I'm not against the split, but any idea about how to compare that it was not changed? by IL code?

Jim8y:
> I'm not against the split, but any idea about how to compare that it was not changed? by IL code?

18 code blocks, not a single line of code has changed, you copy them and search them in the master branch, you'll see, wont take a few minutes.

shargon (approved):

Jim8y:
@shargon Thank you master shargon.

PR #851, opened 2023-12-31
Remove: remove System.Math

Jim8y:
This pr follows https://github.com/neo-project/neo-devpack-dotnet/pull/836


shargon:
I don't understand why

cschuchardt88:
why support math when you have to convert everything to a integer (`BigInteger`). You can have overflows. Try fitting a 256-bit integer into a int16.

Jim8y:
> why support math when you have to convert everything to a integer (`BigInteger`). You can have overflows. Try fitting a 256-bit integer into a int16.

improper convert will have exceptions, no overflow will happen.

Jim8y:
> I don't understand why

Cause you will have two Math in the system, one with Pow, another dont. With Neo.Math, it is precisely clear what is supported by neo, what is not, while using System.Math is a mass cause you can not only use double in the smart contract, you dont know what methods are supported, what are not.

Jim8y:
Hey shargon, I still insist removing the system Math, otherwise you will have to deal with the System.Math's double types in contract. I can ban the use of System.Math with contract analyzer, and help developers to directly use Neo.Math instead of System.Math.

PR #855, opened 2024-01-04
Standalone workflow

cschuchardt88:
# How to Run Workflow
## Publish a release (_Note: **DO NOT UPLOAD BINARIES**_)`
image

# Try It Out
### https://github.com/cschuchardt88/neo-devpack-dotnet/releases

# Workflow Actions (`github`)
image

# Release Page
image

# Test
image

# Extracted Tarball
#### _**NOTE**: That `aspnetcorev2_inprocess.dll` will be removed in next release. Because we removed websocket node in `neo-core`._
image

# File Icon will be
neo


cschuchardt88:
@Jim8y

shargon:
We can wait for https://github.com/neo-project/neo/pull/3018

shargon:
@superboyiii I see it well, if you see it well

Jim8y:
@superboyiii please review

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,144 @@
...
+    - name: .NET Publish (nccs)
+      run: |
+        dotnet publish ./src/Neo.Compiler.CSharp \
+        --framework net7.0 \
+        --configuration ${{ env.CONFIGURATION }} \
+        --runtime ${{ matrix.runtime }} \
+        --self-contained true \
+        --output ${{ env.OUTPUT_PATH }}/${{ matrix.runtime }} \
+        --verbosity normal \
+        -p:VersionPrefix=${{ github.event.inputs.version }} \
```
shargon, line 45: @cschuchardt88 It's possible to use the source code version? in order to avoid possible mistakes

cschuchardt88: It would require a lot work. But can be done with a script or with a few commands in the console. Problem is that we have to read from the file and parse it.

cschuchardt88:
Let add one thing.

shargon:
> @shargon can you look at error, something with test engine.
> 
> image

@cschuchardt88 Is this branch rebased with master? seems that this constructor don't exists here

shargon:
> > > @shargon it's the same as what's in `master` on this repo if that's what your asking.
> > 
> > 
> > neo submodule it's different
> > image
> 
> Got it. Thank you. I don't work with `submodules` much.

I don't know why it doesn't appear in file changed, or at least when we update the branch it should update the submodule...

Jim8y:
@superboyiii mind to take a look here?

.github/workflows/standalone.yml
AnnaShaleva: It is possible without a lot of work, can be done in two ways (or even both ways may be used, in NeoGo and Bane we use both):
1. If job trigger is `release`, then fetch version from the github release tag, it's available via actions API.
2. More common way that works not only for releasing jobs is in a separate job step to build the nccs binary (may be done even with Debug config), then run `nccs --version` and put the result of this step to an output variable. Then in the next step (publishing step) use the output of the previous step as version. Also, if needed, commit sha may be appended to the version of non-release builds.

So I'd say we need to change the existing approach.

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,168 @@
+name: Build & Release (nccs)
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+        description: 'Version Number (Example: 1.0.0)'
+        required: true
```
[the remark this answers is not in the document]
AnnaShaleva: Not in case when you need a specific commit. But it's just a suggestion.

.github/workflows/standalone.yml
cschuchardt88: **Fix**: _Gets from `tag_name` on `release`._

cschuchardt88:
@AnnaShaleva Can you check to see if the `workflow` meets your change requests.

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,146 @@
...
+        -p:PublishTrimmed=false \
+        -p:PublishSingleFile=true \
+        -p:PublishReadyToRun=true \
+        -p:EnableCompressionInSingleFile=true \
+        -p:DebugType=embedded \
+        -p:ServerGarbageCollection=true
+
+    # - name: Remove Xml Comment Files
+    #   working-directory: ${{ env.OUTPUT_PATH }}/${{ matrix.runtime }}
+    #   run: rm -v *.xml
```
AnnaShaleva, line 59: Do we need this commented code?

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,146 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        runtime: [linux-x64, linux-arm64, linux-musl-arm64, osx-x64, osx-arm64, win-x64, win-arm64]
+
+    steps:
+    - name: Set Application Version (Environment Variable)
+      run: |
+        APP_VERSION=$(echo '${{ github.event.release.tag_name }}' | cut -d 'v' -f 2)
+        echo "APP_VERSION=$APP_VERSION" >> $GITHUB_ENV
```
AnnaShaleva, line 25: Personally I prefer binary-based version, but yes, this way is also good. Have you tested on private repo, does this workflow works as expected?

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,146 @@
...
+      env:
+        FILENAME: nccs.${{ env.APP_VERSION }}-${{ matrix.runtime }}
+      run: sha256sum ${{ env.FILENAME }}.tar.xz > ${{ env.FILENAME }}.sha256
+
+    - name: Create Checksum File (win)
+      if: ${{ startsWith(matrix.runtime, 'win') }}
+      working-directory: ${{ env.DIST_PATH }}
+      env:
+        FILENAME: nccs.${{ env.APP_VERSION }}-${{ matrix.runtime }}
+      run: sha256sum ${{ env.FILENAME }}.zip > ${{ env.FILENAME }}.sha256
```
AnnaShaleva, line 98: A lot of unnecessary copy-past. Matrix strategy support structured variables, so in the matrix runtime enumeration add `suffix` of the desired checksum file and unify these three steps:
```
    strategy:
      matrix:
        runtime: [{os: linux-x64, checksumExt: .tar.gz}, ...]
```
See the example in https://github.com/nspcc-dev/neo-go/blob/b028c772f2c592a22796484dfab8f848ca8ed2de/.github/workflows/build.yml#L40.

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,146 @@
...
+
+    - name: Output/Distribution Directory Contents
+      run: |
+        ls -l ${{ env.DIST_PATH }}
+        ls -l ${{ env.OUTPUT_PATH }}/${{ matrix.runtime }}
+
+    - name: Upload Tarball File (linux)
```
AnnaShaleva, line 105: The same optimisation may be done with uploading steps.

.github/workflows/standalone.yml
cschuchardt88: @AnnaShaleva You can check here https://github.com/cschuchardt88/neo-devpack-dotnet/releases/tag/v3.7.0

cschuchardt88:
Closing

PR #857, opened 2024-01-15
Add: add neo as submodule

Jim8y:
This pr adds neo core as submodule, such that we can directly debug from source and apply the latest core update.


shargon (commented):
Fix ut?

Jim8y:
use https://github.com/neo-project/neo-devpack-dotnet/pull/864

Issue #858, opened 2024-01-15
Make `Runtime.ScriptContainer` return a Transaction object


lock9:
Hi,

Problem:
The `Runtime.ScriptContainer` is returning a generic `object`. This demands users to cast it into a transaction every time they want to use it:
`var tx = (Transaction)Runtime.ScriptContainer;`

Proposed Solution:
Since the ScriptContainer will always be a Transaction, the smart contract compiler can return a Transaction type instead of object. 
Result:
`var tx = Runtime.ScriptContainer;`

Jim8y:
we can add a method to get the transaction instead.

lock9:
@Jim8y I've spoken to Roman, who confirmed that it will always return a transaction. I'm not against adding a new interface, but it would be ok to change the existing one

cschuchardt88:
@lock9  You can just add the following code to your contract. It the same we are going to use. `CONVERT` opcode isn't required or generated. `System.Runtime.GetScriptContainer` in the `ApplicationEngine` converts for free for whatever the `IVerifiable` type is. In this case it's `Transaction`.

```json
[Syscall("System.Runtime.GetScriptContainer")]
public static extern Transaction GetTransaction();
```

shargon:
> @Jim8y I've spoken to Roman, who confirmed that it will always return a transaction. I'm not against adding a new interface, but it would be ok to change the existing one. I believe that GetTransaction can be misleading. The GetTransaction is already present elsewhere (probably at the ledger interface)

A consensus wallet that try to verify the block will return the block

roman-khimov:
Block verification is a simple multisig script that doesn't call `GetScriptContainer`. You can't run your code (created with devpack) in this context. Same goes for OnPersist/PostPersist, no user-created code runs there.

Jim8y:
Before i open this pr, I have searched everywhere, exactly as @roman-khimov said, container is only used for transaction.

shargon:
> Block verification is a simple multisig script that doesn't call `GetScriptContainer`. You can't run your code (created with devpack) in this context. Same goes for OnPersist/PostPersist, no user-created code runs there.

That's true, currently we only use a multisign for blocks

roman-khimov:
And even if we'd like to change it for whatever reason (_highly_ unlikely, but we technically can have anything in this script) that'd be code created by the system (CNs), not something created by devpack user.

PR #859, opened 2024-01-15
Add: add runtime.transaction

Jim8y:
Closes https://github.com/neo-project/neo-devpack-dotnet/issues/858


src/Neo.SmartContract.Framework/Services/Runtime.cs
```diff
@@ -34,6 +34,12 @@ public static class Runtime
             get;
         }
 
+        public static extern Transaction Transaction
+        {
+            [Syscall("System.Runtime.GetScriptContainer")]
+            get;
+        }
```
cschuchardt88, line 41: Why dont you just change `ScriptContainer` return type to `Transaction`. It will alway be `Transaction` and its like in `neo-go`

Jim8y: Delete in the future maybe, but I don't want just remove it.

cschuchardt88: But they can alway call `GetScriptContainer` `syscall` and there is no reason to keep it like that? mark as `obsolete` than?

Jim8y: > But they can alway call `GetScriptContainer` `syscall` and there is no reason to keep it like that? mark as `obsolete` than?

Mark it then, I just don't like the idea of change exiting interface arbitrarily.

shargon: In fact it could return a block if you deploy a consensus wallet

roman-khimov: How can you make your code executed in a block context?

cschuchardt88: @shargon What is the Trigger? Cause the only other place you can invoke something other than at the Application trigger is, Verification. Other than that contracts don't get loaded in another triggers. All code in other triggers pass in `null`

shargon: https://github.com/neo-project/neo/blob/a7db45c5364453b828445a541e4ce1d2dcaf9aec/src/Neo/Network/P2P/Payloads/Header.cs#L261

https://github.com/neo-project/neo/blob/a7db45c5364453b828445a541e4ce1d2dcaf9aec/src/Neo/SmartContract/Helper.cs#L319

Here you can see that it was used the block as scriptContainer

Jim8y: Can it be triggered in contract?

shargon (approved):

Issue #861, opened 2024-01-15
Add support to Array.Empty


lock9:
The compiler does not support the following line:

`Array.Empty<object>();`

Would it be possible to add support to it? The IDE doesn't show any errors, it's only possible to know that it doesn't work after trying to compile it:

Failed to compile:
`EXEC : error NC1002: Unknown method: System.Array.Empty<object>() `

This will work but the IDE doesn't like it (shows a warning):
`new object[] { };`

`Avoid unnecessary zero-length array allocations.  Use Array.Empty<object>() instead.[CA1825](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1825)`

cschuchardt88:
Have you tried
```csharp
new object[];
```

PR #862, opened 2024-01-16
Add: add array.empty

Jim8y:
closes https://github.com/neo-project/neo-devpack-dotnet/issues/861


src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -430,6 +430,11 @@ private void ConvertExtern()
                             break;
                     }
                 }
+                if ((bool)Symbol.ToString()?.StartsWith("Array.Empty"))
```
Jim8y, line 433:
```suggestion
                if ((bool)Symbol.ToString()?.StartsWith("System.Array.Empty"))
```

tests/Neo.Compiler.CSharp.UnitTests/TestClasses/Contract_Array.cs
```diff
@@ -89,6 +90,12 @@ public static object TestStructArray()
             return sarray[2];
         }
 
+        public static object[] TestEmptyArray()
```
shargon, line 93: It was not executed this test?

Jim8y: I forgot to commit the UT

Jim8y:
Good to go now. @shargon

superboyiii (approved):
Works well on mine.

PR #871, opened 2024-01-17
move neo-core projects into a dependency folder to make it clear.

Jim8y:
Just find it misleading to mix everything in the src folder. so move the neo-core to a dependency folder. No logic is changed


vncoelho:
Why not move to neo-core the devpack instead?

I think we should archive here as we discussed in the examples PR.

Jim8y:
@vncoelho 

1. Adding submodule is much faster than merging into monorrpo, indeed it to debug the devpack, can't wait for months.

2. Monorepo is for the ease of maintaining and debugging, now the submodule also work.

3. A problem with the monorepo is that it needs to release files, might be too big since nccs also need to release files.

4. Community members have asked to keep the devpack separate.

shargon (approved):

Issue #870, opened 2024-01-17
Automatically include "Debug" event when building in debug mode


lock9:
Problem:
- Many users like to use 'print'/'log' to debug their contracts. It's common for developers to include a 'Debug' event to mimic  console.log operations. 
- The Debug event used to work fine until we introduced breaking changes related to event declaration. This made existing smart contracts break. 

Example:
https://github.com/OnBlockIO/NEP11TemplatePy/blob/687b9116d8313126f44621f6228fa1cbc5ed38ac/contracts/NEP11/NEP11-Template.py#L111

Proposed solution:
If the contract is compiled using debug mode, include a default Debug event.

PR #872, opened 2024-01-17
add debug notify into contract

Jim8y:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/870

```C#
#if DEBUG
            Runtime.Debug("DEBUG");
#endif
```


Jim8y:
only work in debug mode, release mode will fault the execution. set the contract to debug mode by set `-d` while compile the contract.

src/Neo.SmartContract.Framework/Services/Runtime.cs
```diff
@@ -107,6 +109,14 @@ public static class Runtime
         // [Syscall("System.Runtime.Notify")]
         // public static extern void Notify(string eventName, object[] state);
 
+        [Syscall("System.Runtime.Notify")]
+        public static extern void Debug(string eventName, object[] state);
```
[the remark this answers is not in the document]
Jim8y: Yes, its a debug method, i think maybe just debug is Ok.

Jim8y:
@shargon

src/Neo.SmartContract.Framework/Services/Runtime.cs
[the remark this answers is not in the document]
shargon: I mean, the `eventName`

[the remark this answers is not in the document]
Jim8y: > I mean, the `eventName`

I know, make it dynamic would be much harder than make it fixed, cuase you will have to check the parameters, you have to prevent it to be an expression.....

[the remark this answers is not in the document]
Jim8y: would be great if we can make it hapen, use is not suppoed to call this, but i could not figure out a way to get ride of it.

src/Neo.SmartContract.Framework/Services/Runtime.cs
```diff
@@ -107,6 +107,11 @@ public static class Runtime
         // [Syscall("System.Runtime.Notify")]
         // public static extern void Notify(string eventName, object[] state);
 
+        //[Syscall("System.Runtime.Notify")]
+        [OpCode(OpCode.PUSHDATA1, "054465627567")] // 0x5 - Debug
+        [OpCode(OpCode.SYSCALL, "95016f61")] // SHA256(System.Runtime.Notify)[0..4]
+        public static extern void Debug(params object[] state);
```
shargon, line 113: @Jim8y What do you think? we can force "Debug" value like this

tests/Neo.Compiler.CSharp.TestContracts/Contract_Debug.cs
```diff
@@ -8,6 +9,7 @@ public class Contract_Debug : SmartContract.Framework.SmartContract
         public static int TestElse()
         {
 #if DEBUG
+            Runtime.Debug("DEBUG");
```
shargon, line 12: We should check that this is happening

src/Neo.SmartContract.Framework/Services/Runtime.cs
Jim8y: perfect

Jim8y: I wanted to make it work this way as well, just not sure how to construct the scirpt

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -329,6 +329,23 @@ public ContractManifest CreateManifest()
                 ["trusts"] = trusts.Contains("*") ? "*" : trusts.OrderBy(p => p.Length).ThenBy(p => p).Select(u => new JString(u)).ToArray(),
                 ["extra"] = manifestExtra
             };
+
+            // Debug mode
+            if (Options.Debug && eventsExported.All(p => p.Name != "Debug"))
+            {
+                json["abi"]!["events"] = ((JArray)json["abi"]!["events"] ?? new JArray()).Append(new JObject
```
shargon, line 336: Is this required @Jim8y ? maybe we can add it only if it's used

shargon (approved):

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -430,12 +430,15 @@ private void ConvertExtern()
...
-                var isExist = Symbol.ToString()?.StartsWith("System.Array.Empty");
-                if (isExist.HasValue && isExist.Value)
+                if (Symbol.ToString()?.StartsWith("System.Array.Empty") == true)
                 {
                     emitted = true;
                     AddInstruction(OpCode.NEWARRAY0);
                 }
+                else if (Symbol.ToString()?.Equals("Neo.SmartContract.Framework.Services.Runtime.Debug(string)") == true)
+                {
+                    context.AddEvent(new AbiEvent(Symbol, "Debug", new SmartContract.Manifest.ContractParameterDefinition() { Name = "message", Type = ContractParameterType.String }), false);
```
shargon, line 440: I think that it should be an array..

PR #876, opened 2024-01-24
Fix: Fix dependency

Jim8y:


Scripts/load_submodule.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
shargon, line 1: Move to lowercase folder?

shargon (approved):

PR #877, opened 2024-01-26
[Framework: Add] Define Standard Enum

Jim8y:
This pr defines a contract standard enumration, allowing user to use NEPStandard.NEP11 instead of "NEP-11" string when declare the supported standards.


tests/Neo.SmartContract.Framework.UnitTests/SupportedStandardsTest.cs
```diff
@@ -10,7 +10,15 @@ public void TestAttribute()
         {
             var testengine = new TestEngine.TestEngine();
             testengine.AddEntryScript(Utils.Extensions.TestContractRoot + "Contract_SupportedStandards.cs");
-            CollectionAssert.AreEqual(testengine.Manifest.SupportedStandards, new string[] { "NEP10", "NEP5" });
+            CollectionAssert.AreEqual(testengine.Manifest.SupportedStandards, new string[] { "NEP-10", "NEP-5" });
```
shargon, line 13: Shall we use the old format (NEP5 instead of NEP-5)?

cschuchardt88: No! It's not in the spec. There should be a dash (`-`).

shargon (approved):

PR #880, opened 2024-02-01
[Compiler: Comment] add comment to initial value

Jim8y:
This pr adds comments to the initialvalueattribute, make it clear what it does.

### DIFF
--- a/tests/Neo.Compiler.CSharp.TestContracts/Contract_StaticVar.cs
+++ b/tests/Neo.Compiler.CSharp.TestContracts/Contract_StaticVar.cs
@@ -37,7 +37,7 @@ namespace Neo.Compiler.CSharp.UnitTests.TestClasses
         /// <summary>
         /// A static string field initialized with the InitialValue attribute. This demonstrates initializing contract fields that cannot be directly assigned with their value at compile time.
         /// </summary>
-        [InitialValue("hello world", Neo.SmartContract.ContractParameterType.String)]
+        [InitialValue("hello world", ContractParameterType.String)]
         public static readonly string a4 = default;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


tests/Neo.Compiler.CSharp.TestContracts/Contract_StaticVar.cs
```diff
@@ -1,17 +1,49 @@
...
+        /// </summary>
+        [InitialValue("024700db2e90d9f02c4f9fc862abaca92725f95b4fddcc8d7ffa538693ecf463a9", ContractParameterType.PublicKey)]
+        private static readonly ECPoint eCPoint = default;
+
+        /// <summary>
+        /// A static field of type UInt160 initialized with the InitialValue attribute. This allows for compile-time
+        /// initialization of blockchain-specific types like addresses, represented here as Hash160.
+        /// </summary>
+        [InitialValue("NXV7ZhHiyM1aHXwpVsRZC6BwNFP2jghXAq", ContractParameterType.Hash160)]
+        private static readonly UInt160 uInt160 = default;
```
shargon, line 35: Add UT?

Jim8y: sure, will do it later, but i copied this from other test contract, should already being tested. Will add it still if you think it is necessary

Jim8y:
some weird issue happend in the unit test, 
8221706784426_ pic

It wont go into Push when I debug step by step, have no idea why.

shargon:
> some weird issue happend in the unit test, 8221706784426_ pic
> 
> It wont go into Push when I debug step by step, have no idea why.

```
        static readonly BigInteger a2 = BigInteger.Parse("120");
        static readonly BigInteger a3 = BigInteger.Parse("3");
```

Require storage because those calls internally call `atoi`

shargon (approved):

Jim8y:
> @Jim8y Require storage because those calls internally call `atoi`

Thank you shargon, will take me weeks without you.

PR #881, opened 2024-02-01
[Compiler: Add]verify the NEP11 and NEP17 standards while compile them.

Jim8y:
This pr adds verification logic to standard 11 and standard 17.

checking logic is borrowed from the neo-express written by @cschuchardt88  at https://github.com/neo-project/neo-express/blob/9cdfc5bb5a938da2c88257782f634b453cff6776/src/neoxp/Extensions/ContractManifestExtensions.cs, so i just keep the copyright as it is.

### DIFF
--- a/src/Neo.Compiler.CSharp/ContractManifestExtension.cs
+++ b/src/Neo.Compiler.CSharp/ContractManifestExtension.cs
@@ -161,7 +161,7 @@ namespace Neo.Compiler
             }
         }
 
-        internal static ContractManifest VerifyStandards(this ContractManifest manifest)
+        internal static ContractManifest CheckStandards(this ContractManifest manifest)
         {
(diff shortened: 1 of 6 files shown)


src/Neo.Compiler.CSharp/ContractManifestExtension.cs
```diff
@@ -0,0 +1,183 @@
...
+        {
+            if (manifest.SupportedStandards.Contains(NEPStandard.NEP11.ToStandard()))
+            {
+                return manifest.IsNep11Compliant();
+            }
+
+            if (manifest.SupportedStandards.Contains(NEPStandard.NEP17.ToStandard()))
+            {
+                return manifest.IsNep17Compliant();
+            }
```
shargon, line 178: We can have both standards

Jim8y: done

shargon (approved):

PR #883, opened 2024-02-07
[Framwork] remove all string parameters

Jim8y:
This pr focus on remove all strings parameters that are actually standard.


src/Neo.SmartContract.Framework/NEP.cs
```diff
@@ -0,0 +1,20 @@
+namespace Neo.SmartContract.Framework
+{
+    public static class NEP
+    {
+        public const string NEP11 = "NEP-11";
+        public const string NEP17 = "NEP-17";
+
+        public static string[] All => new string[] { NEP11, NEP17 };
```
shargon, line 8: Why use string instead of an enum?

Jim8y: cause enum can not have static method, must have extensions. While this works exactly the same as the previous one.

Jim8y: Well, actually i made this change becaue i want to add extension method to return onNEP17Payment, but i then realized i can just define them as field of a class

src/Neo.SmartContract.Framework/ByteString.cs
```diff
@@ -18,6 +18,11 @@ namespace Neo.SmartContract.Framework
 {
     public abstract class ByteString : IEnumerable<byte>
     {
+        /// <summary>
+        ///  The wildcard string for contract permission settings.
+        /// </summary>
+        public const string Wildcard = "*";
```
shargon, line 24: Use `ContractPermission.Wildcard` ?

src/Neo.SmartContract.Framework/NEP.cs
```diff
@@ -0,0 +1,20 @@
+namespace Neo.SmartContract.Framework
+{
+    public static class NEP
```
shargon, line 3: Use NEP17, and NEP11?

Jim8y: We can add other NEP if necessary, but i am not sure if they still active. So this is general one, not specific to NEP11 and NEP17

src/Neo.SmartContract.Framework/ByteString.cs
Jim8y: I thought about it, but WildCard exists in many situations, trust for instance, so if we have a specific one, we will have many duplicate defination.

src/Neo.SmartContract.Framework/NEP.cs
shargon: I mean, in different classes, then you will see only the methods for your NEP11, or your NEP17, it prevent mix the things

Jim8y: Having them in one place is easier to deal with. I dont think this can mix anything, already simply enough, without this one, they deal with strings directly, they dont mix, how come this can mix.

shargon: You will have all the "reserved words" together, it could be a mess

Jim8y: I dont see how this can be a problem for any  developers, They use NEP17 they use NEP.NEP17, how could they mess with NEP11? they are developers, just not familier with Neo, not becaus they can not program. 

Come on, lets focus on functionalities.

PR #879, opened 2024-01-30
[Compiler] Optimizer to remove unused instructions

Hecate2:
- I used an individual namespace Neo.Optimizer. This is to prevent conflict between Neo.VM.Instruction and Neo.Compiler.Instruction, and to indicate that Neo.Optimizer optimizes contracts written by all languages, not only C#.
- I passed all tests, but I am not really sure whether my optimizations had been applied...
- I actually used neo-fairy-test to test my codes against [NFTLoan](https://github.com/Hecate2/NFTLoan) (https://github.com/Hecate2/neo-fairy-client/blob/master/test_nftloan.py) & [RelationalDB](https://github.com/Hecate2/RelationalDB/)
- My own tests did not verify rarely-seen Instructions like ENDFINALLY and ASSERTMSG. They may not be handled correctly.

PR #887, opened 2024-02-09
Core optimize optimizer

shargon:
Clean code #879


shargon:
I merge it because https://github.com/neo-project/neo-devpack-dotnet/pull/879 was merged before this clean code

PR #888, opened 2024-02-09
Reduce notifications

shargon:
Reduce notifications during compilation

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -163,8 +163,10 @@ namespace Neo.Compiler
 
         public static CompilationContext CompileSources(string[] sourceFiles, Options options)
         {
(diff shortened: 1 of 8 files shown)


Jim8y:
Why? I still could not understand recent prs, do they solve real problems at all? more like minor code optimization. If you do this, we can push many many similar prs every day.

Jim8y (approved):

Jim8y:
Up to you to decide if you realy want to do prs like this. I wont block it.

shargon:
> Why? I still could not understand recent prs, do they solve real problems at all? more like minor code optimization. If you do this, we can push many many similar prs every day.

This is the review for the previous one

cschuchardt88:
> Up to you to decide if you realy want to do prs like this. I wont block it.

I will.

You need to revert it. Before that; copy to other branch and make changes and than put in new PR to merge that branch after the fixes and code is right.

Issue #894, opened 2024-02-13
StoredAttribute default value when integer is not working 


shargon:
During the test on nept17 I found that the totalSupply is not 0 when is not set, we should review it, also, it seems that ECPoint is not included in this switch

https://github.com/neo-project/neo-devpack-dotnet/blob/e8d08a5d8460c97c579d20dd49bb505b8bc3b9a6/src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs#L586

Currently the Nep17 template is not working

PR #895, opened 2024-02-13
bug: Fix default value when Stored

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/894

**Note:**

If a contract is updated from an older version, it could result in an incorrect reading of the storage if it used the attribute `Stored` after this pull request


src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -574,21 +574,32 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
...
+                    case "SByte":
+                    case "Int16":
                     case "Int32":
                     case "Int64":
-                    case "Byte":
                     case "UInt16":
                     case "UInt32":
                     case "UInt64":
                     case "BigInteger":
-                        ChangeType(VM.Types.StackItemType.Integer);
```
shargon, line 586: This is the bug 
https://github.com/neo-project/neo/blob/4a5a7f58567c1bf270d1c5515d90486c803b4e11/src/Neo.VM/Types/Null.cs#L30

vncoelho (commented):
Will mark as a fork?

shargon:
> Will mark as a fork?

it's the compiler, it fix the new code

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -574,21 +574,32 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
                 Call(ApplicationEngine.System_Storage_Get);
                 switch (property.Type.Name)
                 {
-                    case "SByte":
                     case "Short":
+                    case "Byte":
+                    case "SByte":
```
Jim8y, line 579: Byte then short?

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -574,21 +574,32 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
...
+                    case "Int16":
                     case "Int32":
```
shargon, line 581:
```suggestion
                    case "Byte":
                    case "SByte":
                    case "Short":
                    case "Int16":
                    case "Int32":
```

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -620,11 +631,12 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
                     AccessSlot(OpCode.LDARG, 1);
                 switch (property.Type.Name)
                 {
-                    case "SByte":
                     case "Short":
+                    case "Byte":
+                    case "SByte":
+                    case "Int16":
```
shargon, line 637:
```suggestion
                    case "Byte":
                    case "SByte":
                    case "Short":
                    case "Int16":
```

shargon:
@Jim8y @cschuchardt88 merge?

shargon:
Because it's a bug and if was tested, if no one has any review I will merge it @neo-project/core

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -620,19 +641,31 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
...
+                    case "long":
+                    case "ulong":
+                    case "Int64":
                     case "UInt64":
                     case "BigInteger":
                     case "String":
                     case "ByteString":
                     case "UInt160":
                     case "UInt256":
+                    case "ECPoint":
```
roman-khimov, line 668: This change can break existing contracts on upgrade IIUC.

roman-khimov (commented):
Makes sense for numerics, `[Stored(...)]` things should have proper default values.

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
shargon: Yes, but it's wrong, I think that we don't have any release with Stored attribute

Jim8y:
Why a workflow in this pr?

shargon:
> Why a workflow in this pr?

Agree. we can do it in a different one

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -631,14 +641,24 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
...
+
+                    case "int":
+                    case "uint":
+                    case "Int32":
                     case "UInt32":
+
+                    case "long":
+                    case "ulong":
+                    case "Int64":
                     case "UInt64":
```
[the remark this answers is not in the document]
shargon: the unique way to set a different value in a property is because an storage collision (same prefix) and manual write to the same storage

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
roman-khimov: Ah, yes, it was added in #841 and there were no releases since November, so it should be OK changing it now.

roman-khimov (approved):

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
Jim8y: > This change can break existing contracts on upgrade IIUC.

Stored attribute is still not released. Won't cause any trouble.

PR #896, opened 2024-02-13
fix catch in optimizer

Hecate2:
https://github.com/neo-project/neo-devpack-dotnet/pull/879#issuecomment-1939737052

(And, for unit tests, I think we do not verify the correctness, even if each line of codes are covered...)

### DIFF
--- a/src/Neo.Compiler.CSharp/Program.cs
+++ b/src/Neo.Compiler.CSharp/Program.cs
@@ -155,7 +155,14 @@ namespace Neo.Compiler
                 JToken debugInfo = context.CreateDebugInformation(folder);
                 if (!options.NoOptimize)
                 {
-                    (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo);
+                    try
+                    {
+                        (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo);
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.Error.WriteLine($"Failed to optimize: {ex}");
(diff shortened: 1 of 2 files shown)


src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -155,7 +155,14 @@ private static int ProcessOutputs(Options options, string folder, CompilationCon
...
                 if (!options.NoOptimize)
                 {
-                    (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo);
+                    try
+                    {
+                        (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo);
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.Error.WriteLine($"Failed to optimize: {ex}");
```
shargon, line 164: Good!

shargon (approved):

Hecate2:
This may still have problem on try finally (without catch).
And I think it's better to cover all kinds of syntax (instead of all lines of codes) for a reachability analyzer

shargon:
> This may still have problem on try finally (without catch). And I think it's better to cover all kinds of syntax (instead of all lines of codes) for a reachability analyzer

we have these tests

```
public static object tryFinally()
{
    int v = 0;
    try
    {
        v = 2;
    }
    finally
    {
        v++;
    }
    return v;
}

public static object tryFinallyAndRethrow()
{
    int v = 0;
    try
    {
        v = 2;
        throwcall();
    }
    finally
    {
        v++;
    }
    return v;
}
```

PR #897, opened 2024-02-13
Fixed `PackageReference` for `neo`

cschuchardt88:
image

There was a `PackageReference` problem that is now fix. @shargon may or may **not fix** your problem. I don't know how anyone didn't catch this.


cschuchardt88:
@shargon

shargon:
I did this change in my pr and it didn't solve the issue, but we can merge this

shargon (approved):

PR #899, opened 2024-02-14
Added `publish` to `github packages` and fix submodules

shargon:
Copy of https://github.com/neo-project/neo/pull/3072
- Fix submodules during release


shargon:
Long time without nuggets, I will merge in two hours if no one review it in order to recover the packages, it has been tested in Neo and it works, it is only ported to the development environment

PR #900, opened 2024-02-14
Modify changelog

shargon:
Reference: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes


Jim8y (approved):

roman-khimov (commented):
To be fair, I don't believe in autogenerated changelogs. But we can try. Maybe it'd be better than current ones anyway.

Issue #905, opened 2024-02-15
Release workflow doesn't work


shargon:
https://github.com/neo-project/neo-devpack-dotnet/blob/120d87683f3e851c0f3562485b3e7aac7b66b005/.github/workflows/release.yml#L1

It seems that is not working, workflow log always fails

cschuchardt88:
What are trying to accomplish with this `yml`?

cschuchardt88:
@shargon wrong directory.

should be `.github/release.yml`

shargon:
Could someone move it? If don't I will do tomorrow

PR #910, opened 2024-02-18
Fix workflow issues

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/905

- Update pkgs-delete.yml. We should preserve the release packages
- Fix changelog path error


Jim8y (approved):

Jim8y:
If think it is ok

PR #911, opened 2024-02-19
More fixes on optimizer

Hecate2:
It should pass most tests, except that...
1. Unable to support PUSHA without executing the instruction at the pushed address (Typically seen in `Contract_Polymorphism.cs` )

2. There may be error in the DebugInfo of `Contract_TryCatch.cs`:

```
{
    "id": "Neo.Compiler.CSharp.UnitTests.TestClasses.Contract_TryCatch.throwcall()",
    "name": "Neo.Compiler.CSharp.UnitTests.TestClasses.Contract_TryCatch,throwcall",
    "range": "519-531",
    "params": [],
    "return": "Any",
    "variables": [],
    "sequence-points": ["519[3]387:13-387:42", "531[3]388:9-388:10"]
},
{
    "id": "Neo.Compiler.CSharp.UnitTests.TestClasses.Contract_TryCatch.tryNest()",
    "name": "Neo.Compiler.CSharp.UnitTests.TestClasses.Contract_TryCatch,tryNest",
    "range": "532-676",
    "params": [],
    "return": "Any",
    "variables": ["v,Integer,0"],
    "sequence-points": ["535[3]76:17-76:22", "543[3]81:21-81:27", "547[3]82:21-82:33", "552[3]84:17-84:22", "553[3]86:21-86:27", "557[3]87:21-87:33", "562[3]91:21-91:33", "565[3]92:21-92:25", "619[3]95:13-95:18", "620[3]97:17-97:21", "673[3]99:13-99:22", "676[3]100:9-100:10"]
}
```

​	The `range` of instruction addresses above does not match with the assembly codes.

3. Maybe we need to test THROW and ABORT in `catch` blocks. By definition in the VM, THROW in `catch` blocks leads to `finally`, and ABORT terminates the whole execution.
4. The nef and manifest are modified in the process of optimization, and it is hard to revert the modifications. The final output may be polluted if there is exception in the optimizer.

### DIFF
--- a/src/Neo.Compiler.CSharp/Program.cs
+++ b/src/Neo.Compiler.CSharp/Program.cs
@@ -157,7 +157,7 @@ namespace Neo.Compiler
                 {
                     try
                     {
-                        (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo);
+                        (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo.Clone());
                     }
                     catch (Exception ex)
                     {
--- a/src/Neo.Compiler.CSharp/Program.cs
+++ b/src/Neo.Compiler.CSharp/Program.cs
@@ -272,9 +272,16 @@ namespace Neo.Compiler
                     path = Path.Combine(outputFolder, $"{baseName}.asm");
                     File.WriteAllText(path, context.CreateAssembly());
                     Console.WriteLine($"Created {path}");
-                    path = Path.Combine(outputFolder, $"{baseName}.nef.txt");
-                    File.WriteAllText(path, DumpNef.GenerateDumpNef(nef, debugInfo));
-                    Console.WriteLine($"Created {path}");
+                    try
+                    {
+                        path = Path.Combine(outputFolder, $"{baseName}.nef.txt");
+                        File.WriteAllText(path, DumpNef.GenerateDumpNef(nef, debugInfo));
+                        Console.WriteLine($"Created {path}");
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.Error.WriteLine($"Failed to dumpnef: {ex}");
+                    }
                 }
                 Console.WriteLine("Compilation completed successfully.");
                 return 0;
--- a/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
@@ -183,7 +183,7 @@ namespace Neo.Optimizer
                 addressToInstruction.Add(a, i);
             Dictionary<int, string> methodStartAddrToName = new();
             Dictionary<int, string> methodEndAddrToName = new();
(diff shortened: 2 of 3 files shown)


src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -272,9 +272,16 @@ private static int ProcessOutputs(Options options, string folder, CompilationCon
...
-                    File.WriteAllText(path, DumpNef.GenerateDumpNef(nef, debugInfo));
-                    Console.WriteLine($"Created {path}");
+                    try
+                    {
+                        path = Path.Combine(outputFolder, $"{baseName}.nef.txt");
+                        File.WriteAllText(path, DumpNef.GenerateDumpNef(nef, debugInfo));
+                        Console.WriteLine($"Created {path}");
+                    }
+                    catch (Exception ex)
+                    {
```
shargon, line 282: Does we need this try/catch, maybe we should return 1 instead of 0 because maybe you want to generate the dumpnef

Hecate2: There may be unknown exceptions in dumpnef. For example, I met a case where the same block of assembly code about RET works for two pieces of source codes:
```
for ...
   return Something
return null
```
And the dumpnef codes ran into exception, because I assumed that a single block of assembly codes works for only one piece of source code.

shargon:
> The nef and manifest are modified in the process of optimization, and it is hard to revert the modifications. The final output may be polluted if there is exception in the optimizer.

It fails, isn't it? the user should use `--no-optimize`

shargon:
> Unable to support PUSHA without executing the instruction at the pushed address (Typically seen in Contract_Polymorphism.cs )

Could you detect this opcode before and avoid all the process?

Hecate2:
> > Unable to support PUSHA without executing the instruction at the pushed address (Typically seen in Contract_Polymorphism.cs )
> 
> Could you detect this opcode before and avoid all the process?

Yes I can. Yet this forbids too many contracts from optimization.

shargon:
> > > Unable to support PUSHA without executing the instruction at the pushed address (Typically seen in Contract_Polymorphism.cs )
> > 
> > 
> > Could you detect this opcode before and avoid all the process?
> 
> Yes I can. Yet this forbids too many contracts from optimization.

But otherwise they don't work, isn't it?

Hecate2:
> But otherwise they don't work, isn't it?

PUSHA + the address may be actually executed -> It works, and I will not apply any optimization.  

PUSHA + the address is never executed -> I shall delete the target address, but I can neither delete PUSHA, nor determine the target address of PUSHA.  

I can also keep the instructions at the target address, but I cannot determine how many following instructions to keep. This may require read-write analysis, constant propagation analysis and basic block analysis...

shargon (approved):

vncoelho (commented):
Looks good to me, but isn't there any tests that can be added to enforce this improvement?

Jim8y:
> Looks good to me, but isn't there any tests that can be added to enforce this improvement?

@Hecate2 this is a good suggestion, please consider adding some tests in another pr.

PR #917, opened 2024-02-20
Refactor await native

shargon:
I have some problems with the netstandard change and the current package, trying to find how to solve it


shargon:
I think it won't be required after https://github.com/neo-project/neo-devpack-dotnet/pull/926

shargon:
Not required, package was fixed

PR #918, opened 2024-02-20
Clean myget

shargon:


shargon:
Merged because none of the code is changed (only workflow), we need to solve the package problem, and it cannot be delayed any longer, if anyone has a review it can be done in another PR.

PR #919, opened 2024-02-20
Update pkgs-delete.yml

shargon:


shargon:
same as https://github.com/neo-project/neo-devpack-dotnet/pull/918#issuecomment-1954078028

PR #922, opened 2024-02-20
Update pkgs-delete.yml

shargon:
It seems that it works 🥇


shargon:
Merge because I need to release myGet space

PR #934, opened 2024-02-21
TestEngine: Some fixes in TestEngine

shargon:
- ```m.Setup(m => m.MyProperty).Returns(123);``` doesn't work before this because the method is `get_MyProperty`
- Simplify class conversion


shargon:
@Jim8y merge? I want to continue with source code coverage in a different PR

Jim8y (approved):

PR #937, opened 2024-02-22
TestEngine: Source code coverage

shargon:
Part of the https://github.com/neo-project/neo-devpack-dotnet/issues/889

image

image

PR #938, opened 2024-02-22
Fix `if` statement sequence points and `SequencePointInserter`

shargon:
During the codding the https://github.com/neo-project/neo-devpack-dotnet/pull/937 I saw that there was very few sequence points, this fix one of the issue. The other issue is that we use few times `SequencePointInserter` so we can't detect very well the  source coverage, we need to call more `SequencePointInserter` during the compilation


Jim8y:
I am not sure about the logic here, are you sure using the same syntax is right here?

Jim8y (dismissed):

shargon:
> Member

We call `InsertSequencePoint` like this

```
               using (InsertSequencePoint(syntaxNode))
               {
                   preInitialize?.Invoke();
                   ConvertExpression(model, initializer.Value);
                   postInitialize?.Invoke();
               }
```

And it want to add the source code to all th instructions when it's disposed, before this pr, it only add the source code reference to the last one

vncoelho (commented):
This is a bug just related to coverage, right?

shargon:
> This is a bug just related to coverage, right?

Yes, I think that debugInfo is only used for this, and for allow to Debug the source code in neo express o vs extension

src/Neo.Compiler.CSharp/SequencePointInserter.cs
```diff
@@ -17,19 +17,29 @@ namespace Neo.Compiler
...
             this.position = instructions.Count;
         }
+
         void IDisposable.Dispose()
         {
-            if (position < instructions.Count)
-                instructions[position].SourceLocation = syntax.GetLocation();
+            for (int x = position; x < instructions.Count; x++)
+            {
+                if (instructions[position].SourceLocation is null)
```
vncoelho, line 34: instructions[x] or  position?

shargon: Nice catch

shargon, line 34:
```suggestion
                if (instructions[x].SourceLocation is null)
```

vncoelho: Isn't possible to do some simple UT's? Because a simple loop would detect this error.

vncoelho (approved):

shargon:
I will try to detect the fix

shargon:
image

Now if branches work

image

src/Neo.Compiler.CSharp/MethodConvert/Statement/IfStatement.cs
```diff
@@ -47,21 +47,31 @@ partial class MethodConvert
...
                 ConvertExpression(model, syntax.Condition);
-            Jump(OpCode.JMPIFNOT_L, elseTarget);
-            ConvertStatement(model, syntax.Statement);
-            if (syntax.Else is null)
-            {
-                elseTarget.Instruction = AddInstruction(OpCode.NOP);
-            }
-            else
+
+            using (InsertSequencePoint(syntax.Statement))
```
shargon, line 54: @Jim8y if we don't add this `InsertSequencePoint` or it will be impossible to check well the source coverage. If was fixed.

src/Neo.Compiler.CSharp/MethodConvert/SourceConvert.cs
```diff
@@ -33,19 +33,16 @@ private void ConvertSource(SemanticModel model)
...
                     ConvertStatement(model, syntax.Body);
                 else if (syntax.ExpressionBody is not null)
-                    using (InsertSequencePoint(syntax.ExpressionBody.Expression))
-                        ConvertExpression(model, syntax.ExpressionBody.Expression);
+                    ConvertExpression(model, syntax.ExpressionBody.Expression);
                 else
                     ConvertNoBody(syntax);
                 break;
             case ArrowExpressionClauseSyntax syntax:
-                using (InsertSequencePoint(syntax))
```
shargon, line 42: Here the expression in sequence point was different, another bug

vncoelho (commented):
very good @shargon, now it is working again

shargon:
Now we have 100% branch coverage in nep17 template

image

vncoelho (approved):

shargon:
@vncoelho that was last change :)

vncoelho:
what is this license parameter?

shargon:
> what is this license parameter?

The package that we use allow to specify a license, it provide more charts, and information.

vncoelho (approved):

Issue #751, opened 2022-06-07
nccs 3.3 emitting hidden warnings in code gen file


devhawk:
When I compile the registrar sample with nccs 3.3, there's a bunch of extraneous warnings about using directives in the generated `.AssemblyInfo.cs` file. It would be great if we didn't emit these warnings on the console

``` shell
$ dotnet nccs  Registrar.cs obj\Debug\net6.0\.NETCoreApp,Version=v6.0.AssemblyAttributes.cs obj\Debug\net6.0\registrar.AssemblyInfo.cs --base-name registrar --debug --no-optimize --no-inline
C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\obj\Debug\net6.0\.NETCoreApp,Version=v6.0.AssemblyAttributes.cs(2,1): hidden CS8019: Unnecessary using directive.
C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\obj\Debug\net6.0\.NETCoreApp,Version=v6.0.AssemblyAttributes.cs(3,1): hidden CS8019: Unnecessary using directive.
C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\obj\Debug\net6.0\registrar.AssemblyInfo.cs(11,1): hidden CS8019: Unnecessary using directive.
C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\obj\Debug\net6.0\registrar.AssemblyInfo.cs(10,1): hidden CS8019: Unnecessary using directive.
Created C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\bin\sc\registrar.nef
Created C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\bin\sc\registrar.manifest.json
Created C:\Users\harry\Source\neo\seattle\samples\registrar-sample\src\bin\sc\registrar.nefdbgnfo
```

shargon:
It depends the source code, isn't it?

devhawk:
These warnings are marked as hidden (specifically listed as `hidden CS8019`). Would be nice if `nccs` didn't show hidden warnings.

PR #940, opened 2024-02-22
Fix emitting hidden warnings

shargon:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/751

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -57,7 +57,7 @@ namespace Neo.Compiler
         private readonly Dictionary<ITypeSymbol, byte> vtables = new(SymbolEqualityComparer.Default);
         private byte[]? script;
 
-        public bool Success => diagnostics.All(p => p.Severity != DiagnosticSeverity.Error);
+        public bool Success => !diagnostics.Any(p => p.Severity == DiagnosticSeverity.Error);
         public IReadOnlyList<Diagnostic> Diagnostics => diagnostics;
         public string? ContractName => displayName ?? Options.BaseName ?? className;
         private string? Source { get; set; }
--- a/src/Neo.Compiler.CSharp/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationContext.cs
@@ -124,7 +124,7 @@ namespace Neo.Compiler
             foreach (SyntaxTree tree in compilation.SyntaxTrees)
             {
                 SemanticModel model = compilation.GetSemanticModel(tree);
-                diagnostics.AddRange(model.GetDiagnostics());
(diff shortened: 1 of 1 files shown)


src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -57,7 +57,7 @@ public class CompilationContext
         private readonly Dictionary<ITypeSymbol, byte> vtables = new(SymbolEqualityComparer.Default);
         private byte[]? script;
 
-        public bool Success => diagnostics.All(p => p.Severity != DiagnosticSeverity.Error);
+        public bool Success => !diagnostics.Any(p => p.Severity == DiagnosticSeverity.Error);
```
vncoelho, line 60: this is a good change

src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -124,7 +124,7 @@ private void Compile()
             foreach (SyntaxTree tree in compilation.SyntaxTrees)
             {
                 SemanticModel model = compilation.GetSemanticModel(tree);
-                diagnostics.AddRange(model.GetDiagnostics());
+                diagnostics.AddRange(model.GetDiagnostics().Where(u => u.Severity != DiagnosticSeverity.Hidden));
```
vncoelho, line 127: what this Hidden means?

shargon: Something that we don't need to show to the user :)

vncoelho: "These are typically used to provide suggestions or insights that could improve the code's quality, readability, or performance but are not critical errors or warnings that must be addressed for the code to run correctly."

In this case we are not talking about users, @shargon 
Devs are using these tools

vncoelho: In think the success change is good

shargon: > "These are typically used to provide suggestions or insights that could improve the code's quality, readability, or performance but are not critical errors or warnings that must be addressed for the code to run correctly."
> 
> In this case we are not talking about users, @shargon Devs are using these tools

Yes, devs hahah, but we don't need to show them these notifications https://github.com/neo-project/neo-devpack-dotnet/issues/751

vncoelho (approved):

src/Neo.Compiler.CSharp/CompilationContext.cs
Jim8y, line 60: Hey, why this change? aren't them exactly the same?

shargon: yes xD

PR #944, opened 2024-02-23
TestEngine: Allow to change the caller

shargon:
Before this change it's not possible to fake (easy) this method for testing

```csharp
public static bool ItsMe()
{
    Assert(Runtime.CallingScriptHash == Runtime.ExecutingScriptHash, "only me");
    return true;
}
```


shargon:
Note: this will mock the Syscall, but neo, and native contracts doesn't call this syscall for extract the callingScriptHash, so or we can't do it there, or we should change neo, to query the syscall...

vncoelho (approved):
Good change with being null returning the same, tests are good.

PR #953, opened 2024-02-26
Fixes for compiling multi contracts in single csproj

Hecate2:
1. I believe there are problems at https://github.com/neo-project/neo-devpack-dotnet/blob/7add5a9139a57bc0743d367aa048f4408c66ff8b/src/Neo.Compiler.CSharp/Program.cs#L139-L142  A return value of 0 means correct compilation, while the current codes immediately returns 0 on any p != 1. This does not consider the compilation of the subsequent multiple contracts, and does not handle exceptions.
2. Current codes may build `.artifacts.cs` files in the output directory. If a user just want everything to be placed in `bin/sc`, the `artifacts.cs` may be considered as contract source codes, when the compiler is run for the second time. I asked the compiler to ignore `bin/sc`, but this **DOES NOT** fix the cases where users place outputs in another directory contained by the directory of the `.csproj`.

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationEngine.cs
+++ b/src/Neo.Compiler.CSharp/CompilationEngine.cs
@@ -172,8 +172,9 @@ namespace Neo.Compiler
         {
             string folder = Path.GetDirectoryName(csproj)!;
             string obj = Path.Combine(folder, "obj");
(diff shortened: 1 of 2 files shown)


Jim8y (approved):

shargon (commented):
Good catch

PR #954, opened 2024-02-26
ensure finally executed on exceptions from catch

Hecate2:
I am directly checking `testengine.CurrentContext.LocalVariables[0]`, without making sure it is actually the variable `v`.


Jim8y:
This is adding a unit test?

Hecate2:
> This is adding a unit test?

Yes. The coverage remains the same, but I think there is no test that ensures finally on exception from catch.

Jim8y (approved):

Issue #955, opened 2024-02-26
Avoid reuse the same namespace in devpack


shargon:
We use the same namespace for `SmartContract.Framework` and Neo types, this make impossible to add the reference of both assemblies

![image](https://github.com/neo-project/neo-devpack-dotnet/assets/3167973/343f7f75-1b21-433a-bc31-93aa051a6ee9)

shargon:
@Jim8y Move the namespace to `Neo.SmartContract.Framework` in `Neo.SmartContract.Framework`?

Jim8y:
> @Jim8y Move the namespace `Neo` to `Neo.SmartContract.Framework` in `Neo.SmartContract.Framework`?

As long as it works.

PR #956, opened 2024-02-26
TestEngine: Template testing optimization

shargon:
Wait for https://github.com/neo-project/neo-devpack-dotnet/pull/958
Close https://github.com/neo-project/neo-devpack-dotnet/issues/955

@Jim8y I want to remove the msbuild event, it spend a lot of time creating the templates, my idea, test the template creation in workflow, and the unit test can have the artifacts inside the code, and compile the template source code in order to ensure that it's the same as expected, in this way we get two things:

- Speed up the compilation process.
- Add coverage compilation to Compiler source code.
- Changed the namespace from `Neo.SmartContract.Framework` to be different than `Neo`.

The problem, is that i need Neo, and Compiler (https://github.com/neo-project/neo-devpack-dotnet/issues/955)


src/Neo.SmartContract.Testing/Coverage/Formats/ConsoleFormat.cs
```diff
@@ -17,9 +17,9 @@ public partial class ConsoleFormat : CoverageFormatBase
         /// </summary>
         /// <param name="contract">Contract</param>
         /// <param name="Filter">Method Filter</param>
```
shargon, line 19:
```suggestion
        /// <param name="filter">Method Filter</param>
```

Jim8y:
Sorry, miss touched the button

src/Neo.SmartContract.Framework/Native/CalledByGroupCondition.cs
```diff
@@ -1,19 +1,17 @@
...
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+namespace Neo.SmartContract.Framework.Native
+{
+    public class CalledByGroupCondition : WitnessCondition
+    {
+        public ECPoint Group;
+    }
+}
```
Jim8y, line 17: what happend to this file?

shargon:
> why some file changed without change anything? did you changed the encoding?

changing LF back

src/Neo.SmartContract.Framework/Native/CalledByGroupCondition.cs
shargon: solved

src/Neo.Compiler.CSharp/CompilationEngine.cs
```diff
@@ -216,19 +216,27 @@ public Compilation GetCompilation(string csproj)
...
+            var sourceFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            if (!remove.Contains("*.cs"))
+            {
+                var obj = Path.Combine(folder, "obj");
+                var binSc = Path.Combine(Path.Combine(folder, "bin"), "sc");
+                foreach (var entry in Directory.EnumerateFiles(folder, "*.cs", SearchOption.AllDirectories)
+                      .Where(p => !p.StartsWith(obj) && !p.StartsWith(binSc))
+                      .Select(u => u))
+                      //.GroupBy(Path.GetFileName)
```
shargon, line 233: @Jim8y Why we group by the filename, it can exists the same name in different folders

tests/Neo.SmartContract.Template.UnitTests/templates/TestCleanup.cs
```diff
@@ -8,13 +9,51 @@
...
+
+            // Compile
+
+            var result = new CompilationEngine(new Options()
+            {
+                Debug = true,
+                Nullable = Microsoft.CodeAnalysis.NullableContextOptions.Enable,
+                GenerateArtifacts = Options.GenerateArtifactsKind.None
+            })
+            .CompileSources(
```
shargon, line 33: @Jim8y I didn't receive 3 context... no errors and no success, 0 items in the collection

src/Neo.Compiler.CSharp/CompilationEngine.cs
Jim8y: Same filename in different folder but contain the same class will cause the complie to fail. Unless you have a better method, otherwise comment this out will cause building exception.

.github/workflows/main.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         dotnet format --no-restore --verify-no-changes --verbosity diagnostic
     - name: Build Neo.Compiler.CSharp
       run: dotnet build ./src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj
-    - name: Build Neo.SmartContract.Template and generate artifacts
+    - name: Build Neo.SmartContract.Template and test template
```
shargon, line 30:
```suggestion
    - name: Build Neo.SmartContract.Template and test templates
```

tests/Neo.SmartContract.Template.UnitTests/templates/TestCleanup.cs
```diff
@@ -0,0 +1,130 @@
...
+            // Compile
+
+            //var frameworkObj = Path.Combine(frameworkPath, "obj");
+            var result = new CompilationEngine(new CompilationOptions()
+            {
+                Debug = true,
+                Nullable = Microsoft.CodeAnalysis.NullableContextOptions.Disable
+            })
+            .CompileSources(
+                ("Neo.SmartContract.Framework", "3.6.2-CI00520"),
```
shargon, line 40: @Jim8y it won't work until merge .... please review without workflow, we will need to fix it later

shargon: The namespace change make impossible to pass the `CheckNep17Compliant` because the type is Any instead of `Hash160` for `balanceOf` because it takes the framework source (without the namespace change) and the type it's different.

tests/Neo.SmartContract.Template.UnitTests/templates/TestCleanup.cs
```diff
@@ -45,21 +47,21 @@ public void EnsureArtifactsUpToDate()
 
             var root = Path.GetPathRoot(templatePath) ?? "";
             var content = File.ReadAllText(Path.Combine(artifactsPath, "neocontractnep17/TestingArtifacts/Nep17TemplateContract.artifacts.cs"));
-            var artifact = CreateArtifact(result[0], "Nep17TemplateContract");
+            var artifact = CreateArtifact(result[0], nameof(Nep17ContractTemplate));
```
Jim8y, line 56: How about using CreateArtifact<Nep17ContractTemplate>, instead of adding nameof.  just make it as easy as possible.

Jim8y (approved):

src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -79,10 +79,13 @@ public static ContractParameterType GetContractParameterType(this ITypeSymbol ty
                 case "string": return ContractParameterType.String;
                 case "byte[]": return ContractParameterType.ByteArray;
                 case "object": return ContractParameterType.Any;
-                case "Neo.Cryptography.ECC.ECPoint": return ContractParameterType.PublicKey;
+                case "Neo.ECPoint": // Old Neo.SmartContract.Framework
```
shargon, line 82:
```suggestion
                case "Neo.Cryptography.ECC.ECPoint": // Old Neo.SmartContract.Framework
```

tests/Neo.SmartContract.Template.UnitTests/templates/TestCleanup.cs
shargon: I will clean it

shargon: Done

Jim8y:
does this mean all existing smart contract must update if they want to compile with the new compiler?

shargon:
> does this mean all existing smart contract must update if they want to compile with the new compiler?

I think they should work

Jim8y (approved):

PR #959, opened 2024-02-26
[Compiler Fix + Framework Add] add oracle interface and fix issue

Jim8y:
This pr fix the compiler issue where failed context will not be fetched.
and add IOracle interface.

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationEngine.cs
+++ b/src/Neo.Compiler.CSharp/CompilationEngine.cs
@@ -21,6 +21,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Xml.Linq;
+using Akka.Util.Internal;
 using BigInteger = System.Numerics.BigInteger;
 
 namespace Neo.Compiler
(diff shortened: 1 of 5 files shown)


src/Neo.SmartContract.Framework/Interfaces/IOracle.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <param name="jsonString">The oracle response data in format of json string</param>
+    ///
+    /// <example>
+    /// public static void OnOracleResponse(string url, object userData, OracleResponseCode code, string result)
+    /// {
+    ///     if (Runtime.CallingScriptHash != Oracle.Hash) throw new Exception("Unauthorized!");
+    ///     Storage.Put(Storage.CurrentContext, PreData, result);
+    /// }
+    /// </example>
+    public void OnOracleResponse(
```
shargon, line 25: Can we add:

if (Runtime.CallingScriptHash != Oracle.Hash)
                throw new System.Exception("Unauthorized!");

in the interface?

src/Neo.Compiler.CSharp/CompilationEngine.cs
```diff
@@ -21,6 +21,7 @@
 using System.IO;
 using System.Linq;
 using System.Xml.Linq;
+using Akka.Util.Internal;
```
shargon, line 24:
```suggestion
```

shargon (approved):

src/Neo.SmartContract.Framework/Interfaces/IOracle.cs
Jim8y: Its an interface, no implementation can be added.

PR #960, opened 2024-02-26
Fix optimizations using `CompilationEngine`

shargon:
Some optimizations are lost if you use `CompilationEngine` by yourself


shargon:
+3.1% coverage!

vncoelho (approved):

PR #961, opened 2024-02-27
test assert & abort

Hecate2:
I am making use of the optimizer and assembly analyzers...
Will push more tests about abort & assert later.

### DIFF
--- a/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
@@ -176,7 +176,7 @@ namespace Neo.Optimizer
 
         public static (int start, int end) GetMethodStartEndAddress(string name, JToken debugInfo)
         {
-            name = name[0].ToString().ToUpper() + name.Substring(1);  // first letter uppercase
+            name = name.Length == 0 ? string.Empty : name[0].ToString().ToUpper() + name.Substring(1);  // first letter uppercase
             int start = -1, end = -1;
             foreach (JToken? method in (JArray)debugInfo["methods"]!)
             {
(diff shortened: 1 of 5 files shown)


Jim8y:
This is not done yet by us, but may you please add comments to the test contract, explaining the expected behavior?

Hecate2:
> This is not done yet by us, but may you please add comments to the test contract, explaining the expected behavior?

OK

src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
```diff
@@ -174,6 +174,32 @@ string TryComment(int catchOffset, int finallyOffset)
             }
         }
 
+        public static (int start, int end) GetMethodStartEndAddress(string name, JToken debugInfo)
+        {
+            name = name[0].ToString().ToUpper() + name.Substring(1);  // first letter uppercase
```
shargon, line 179:
```suggestion
            name = name.Length == 0 ? string.Empty : name[0].ToString().ToUpper() + name.Substring(1);  // first letter uppercase
```

shargon (approved):

Jim8y:
@shargon also a suggestion to you, may you please add comments to every methods that you add to the code base? especially public methods. I am not going to require you of doing that, but suggest you to do it.

src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
```diff
@@ -174,6 +174,32 @@ string TryComment(int catchOffset, int finallyOffset)
             }
         }
 
+        public static (int start, int end) GetMethodStartEndAddress(string name, JToken debugInfo)
+        {
+            name = name.Length == 0 ? string.Empty : name[0].ToString().ToUpper() + name.Substring(1);  // first letter uppercase
```
Jim8y, line 179: @shargon Nullorempty?

PR #964, opened 2024-02-27
Allow to choose the kind of optimization

shargon:
While we are testing the new optimization, we should be able to disable it.

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationOptions.cs
+++ b/src/Neo.Compiler.CSharp/CompilationOptions.cs
@@ -10,15 +10,26 @@
 
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using System;
 using System.Collections.Generic;
 
 namespace Neo.Compiler
 {
     public class CompilationOptions
     {
+        [Flags]
+        public enum OptimizationType : byte
+        {
+            None = 0,
+            Basic = 1,
+            Experimental = 2,
+
+            All = Basic | Experimental
+        }
+
         public NullableContextOptions Nullable { get; set; }
         public bool Debug { get; set; }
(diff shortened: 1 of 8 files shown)


src/Neo.Compiler.CSharp/CompilationOptions.cs
```diff
@@ -10,15 +10,26 @@
...
 namespace Neo.Compiler
 {
     public class CompilationOptions
     {
+        [Flags]
+        public enum OptimizationType : byte
+        {
+            None = 0,
+            Basic = 1,
+            Experimental = 2,
```
vncoelho, line 25: Expiremental in the release as well?

shargon: Yes

Jim8y: A different name please, too long~~

shargon: Ideas? people use experimental always

Jim8y: o0, o1, o2, o3

Jim8y: I think this is commonly used by compilers.

Jim8y: I don't think it's a good thing to ask user type a long word. But I am not going to ask you to do it.

Jim8y (approved):

src/Neo.Compiler.CSharp/CompilationOptions.cs
shargon: I don't think that no one should use only experimental, just use `all`

PR #966, opened 2024-02-28
fixes on optimizer about CATCH, ENDFINALLY and PUSHA

Hecate2:


Hecate2:
There should be only "CALLA not supported" exception in the optimizer now.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Abort.cs
```diff
@@ -10,6 +10,7 @@ namespace Neo.Compiler.CSharp.UnitTests
     public class UnitTest_Abort
     {
         TestEngine testengine;
+        bool[] falseTrue = new bool[] { false, true };
         [TestInitialize]
```
shargon, line 14:
```suggestion
        bool[] falseTrue = new bool[] { false, true };

        [TestInitialize]
```

shargon (approved):

Hecate2:
Commit https://github.com/neo-project/neo-devpack-dotnet/commit/cac675919f3f90e57723b6b301942f77f9fbf215 introduces dead loop in CoverInstructions, when a method calls itself, or multiple methods call each other cyclically. Will fix in 24 hours.

Hecate2:
> Commit [cac6759](https://github.com/neo-project/neo-devpack-dotnet/commit/cac675919f3f90e57723b6b301942f77f9fbf215) introduces dead loop in `CoverInstruction`.

Sorry but I just tested recursion and verified that the optimizer does **NOT** run into exceptions. Yet asserting a recursively called branch to throw nothing and never aborting may be dogmatic.

PR #967, opened 2024-02-28
Update native contracts

shargon:
- Added `CryptoLib.keccak256`.
- Added `NeoToken.CommitteeChanged` event.


src/Neo.SmartContract.Testing/Native/ContractManagement.cs
```diff
@@ -6,33 +6,42 @@ namespace Neo.SmartContract.Testing.Native;
...
...""parameters"":[],""returntype"":""Integer"",""offset"":42,""safe"":true},{""name"":""hasMethod"",""parameters"":[{""name"":""hash"",""type"":""Hash160""},{""name"":""method"",""type"":""String""},{""name"":""pcount"",""type"":""Integer""}],""returntype"":""Boolean"",""offset"":49,""safe"":true},{""name"":""setMinimumDeploymentFee"",""parameters"":[{""name"":""value"",""type"":""Integer""}],""returntype"":""Void"",""offset"":56,""safe"":false},{""name"":""update"",""parameters"":[{""name"":""nefFile"",""type"":""ByteArray""},{""name"":""manifest"",""type"":""ByteArray""}],""returntype"":""Void"",""offset"":63,""safe"":false},{""name"":""update"",""parameters"":[{""name"":""nefFile"",""type"":""ByteArray""},{""name"":""manifest"",""type"":""ByteArray""},{""name"":""data"",""type"":""Any""}],""returntype"":""Void"",""offset"":70,""safe"":false}],""events"":[{""name"":""Deploy"",""parameters"":[{""name"":""Hash"",""type"":""Hash160""}]},{""name"":""Update"",""parameters"":[{""name"":""Hash"",""type"":""Hash160""}]},{""name"":""Destroy"",""parameters"":[{""name"":""Hash"",""type"":""Hash160""}]}]},""permissions"":[{""contract"":""*"",""methods"":""*""}],""trusts"":[],""extra"":null}");
```
Jim8y, line 11: we need another way to replace a hardcode manifest. can be fixed later.

shargon: It's better to have the nef and manifest inside the artifacts, in case of native contracts can be generated calling neo for avoid this issue, but it can be not synchronized with the methods

shargon (commented):
It was autogenerated with the extension, but we can remove the nullable types when is not in native contracts

Jim8y (approved):

PR #968, opened 2024-02-28
TestEngine: Migrate current Framework's ut

shargon:
Better to merge https://github.com/neo-project/neo-devpack-dotnet/pull/969 before
Part of https://github.com/neo-project/neo-devpack-dotnet/issues/949

This pr migrate the unit tests from `SmartContract.Framework` to the new `TestEngine`.

It allow to check the coverage of the tested contracts in Coveralls, easier to use and easier to test.

Note: Most of the changed files are artifacts autogenerated from the tested nefFiles.

PR #969, opened 2024-02-29
TestEngine: Some fixes

shargon:
In order to reduce the changes from https://github.com/neo-project/neo-devpack-dotnet/pull/968 I will separate the fixes found during the process

- Fixed console coverage output.
- Added `CallFlags` in TestEngine.

### DIFF
--- a/src/Neo.SmartContract.Testing/TestEngine.cs
+++ b/src/Neo.SmartContract.Testing/TestEngine.cs
@@ -173,6 +173,11 @@ namespace Neo.SmartContract.Testing
         public UInt160 Sender => Transaction.Sender;
 
         /// <summary>
+        /// Call flags
+        /// </summary>
+        public CallFlags CallFlags { get; set; } = CallFlags.All;
+
+        /// <summary>
         /// Native artifacts
         /// </summary>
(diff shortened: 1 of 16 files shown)


src/Neo.SmartContract.Testing/TestEngine.cs
```diff
@@ -653,6 +665,14 @@ public StackItem Execute(Script script)
             };
         }
 
+        /// <summary>
+        /// Clear Transaction Signers
+        /// </summary>
+        public void SetTransactionSigners()
```
Jim8y, line 671: Set or Clear?

Jim8y (approved):

src/Neo.SmartContract.Testing/TestEngine.cs
shargon: We can call them clear if you prefer

shargon, line 671:
```suggestion
        public void ClearTransactionSigners()
```

shargon:
@Jim8y renamed

Jim8y (approved):

PR #974, opened 2024-03-02
[Compiler Add]allow direct assignment to UInt160, UInt256, and ECPoint.

Jim8y:
This pr allows user to directly initialize contract UInt and ECPoint field and property.

Before:

```C#
        [PublicKey("024700db2e90d9f02c4f9fc862abaca92725f95b4fddcc8d7ffa538693ecf463a9")]
        private static readonly ECPoint eCPoint = default;

        [Hash160("NXV7ZhHiyM1aHXwpVsRZC6BwNFP2jghXAq")]
        private static readonly UInt160 uInt160 = default;
```

Now:

```C#
      private static readonly ECPoint eCPoint = "024700db2e90d9f02c4f9fc862abaca92725f95b4fddcc8d7ffa538693ecf463a9";
      private static readonly UInt160 uInt160 = "NXV7ZhHiyM1aHXwpVsRZC6BwNFP2jghXAq";
```


src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -39,5 +40,15 @@ public abstract class ECPoint : ByteString
...
+
+        /// <summary>
+        /// Implicitly converts a hexadecimal string to a PublicKey object.
+        /// Assumes the string is a valid hexadecimal representation.
+        /// </summary>
+        public static implicit operator ECPoint(string value)
+        {
+            return value;
+        }
+
```
shargon, line 52:
```suggestion
```

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -71,5 +72,23 @@ public string ToAddress(byte version)
             data = Helper.Concat(data, this);
             return StdLib.Base58CheckEncode((ByteString)data);
         }
+
+
```
shargon, line 76:
```suggestion
```

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -71,5 +72,23 @@ public string ToAddress(byte version)
...
+            // Convert the hexadecimal string to a byte array.
+            // Ensure the byte array is of the expected size for a UInt160.
+            if (value.Length != 20) // UInt160 is expected to be 20 bytes.
+            {
+                throw new Exception("The provided string does not represent a valid UInt160 value.");
+            }
+            // Use the explicit byte[] to UInt160 conversion defined earlier.
+            return value;
+        }
+
```
shargon, line 92:
```suggestion
```

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -71,5 +72,21 @@ public string ToAddress(byte version)
...
+        /// <summary>
+        /// Implicitly converts a hexadecimal string to a UInt160 object.
+        /// Assumes the string is a valid hexadecimal representation.
+        /// </summary>
+        public static implicit operator UInt160(string value)
+        {
+            // Convert the hexadecimal string to a byte array.
+            // Ensure the byte array is of the expected size for a UInt160.
+            if (value.Length != 20) // UInt160 is expected to be 20 bytes.
```
shargon, line 84: Do the same for UInt256?

src/Neo.SmartContract.Framework/ECPoint.cs
```diff
@@ -39,5 +40,14 @@ public abstract class ECPoint : ByteString
 
         [OpCode(OpCode.CONVERT, StackItemType.Buffer)]
         public static extern explicit operator byte[](ECPoint value);
+
+        /// <summary>
+        /// Implicitly converts a hexadecimal string to a PublicKey object.
+        /// Assumes the string is a valid hexadecimal representation.
+        /// </summary>
+        public static implicit operator ECPoint(string value)
+        {
```
shargon, line 49: Length check?

Jim8y: length check here is useless, cause the internal logic never take effect. Can you please help me update it?

src/Neo.SmartContract.Framework/UInt160.cs
Jim8y: Only if we have it. But should be done in another pr by you to add it maybe, cause currently we don't support it.

Jim8y:
@shargon good to go.

Jim8y:
This might not be right. Cause user may want to use it in contract at runtime, but this only work at compile time. I will further update it.

shargon (approved):

PR #976, opened 2024-03-02
TestingEngine: Native contracts manifest reuse

shargon:
Wait for https://github.com/neo-project/neo/pull/3171

- Added TestingEngine tests to workflow.
- Reused manifest from neo core.
- Renamed Native contracts to be the same as core.


Jim8y:
I have a problem, every time i run, changes like this happen. Please fix.

image

Jim8y:
Its in testingArtifacts/Contract_Contract.cs and other two. Also in template/neocontractxxx/testingArtigfactxxxx

shargon:
> I have a problem, every time i run, changes like this happen. Please fix.
> 
> image

Looks like your build It's not deterministic ..

shargon:
@Jim8y could you check if it's your compiler version inside the neffile?

Jim8y:
> @Jim8y could you check if it's your compiler version inside the neffile?

Why it's related to compiler version? Isn't it's built from the source code?

shargon:
> > @Jim8y could you check if it's your compiler version inside the neffile?
> 
> Why it's related to compiler version? Isn't it's built from the source code?

Because it's inside the `nefFile`

shargon:
@neo-project/core http://seed2t5.neo.org:20332 `TotalSupply` it's not 100m?

shargon:
@Jim8y the conflicts are fixed

Jim8y (approved):

PR #985, opened 2024-03-06
[Compiler fix pattern]pattern

Jim8y:


src/Neo.SmartContract.Framework/ByteString.Extension.cs
```diff
@@ -57,7 +57,7 @@ public static bool IsAlphabet(this ByteString byteString)
     {
         foreach (var value in byteString)
         {
-            if (!((value >= 65 && value <= 90) || (value >= 97 && value <= 122)))
+            if (value is (< 65 or > 90) and (< 97 or > 122))
```
shargon, line 60: Add ut?

Jim8y: This is for showing you where is the problem.

shargon: But we should add unit test to ensure that is fixed

PR #986, opened 2024-03-06
[Compuler Bug] fix void bug

Jim8y:
This pr fix a bug in the compiler where 

```C#
  public static void a()=> i+=1;
```

will leave a return type in the stack even its defined as void.


src/Neo.Compiler.CSharp/MethodConvert/CallHelpers.cs
```diff
@@ -180,6 +180,7 @@ private void EmitCall(MethodConvert target)
                 AddInstruction(target._instructions[i].Clone());
         else
             Jump(OpCode.CALL_L, target._startTarget);
+
```
shargon, line 183:
```suggestion
```

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_StaticClass.cs
```diff
@@ -0,0 +1,30 @@
...
+using Neo.VM.Types;
+using System;
+using Neo.SmartContract.TestEngine;
+
+namespace Neo.Compiler.CSharp.UnitTests
+{
+    [TestClass]
+    public class UnitTest_StaticClass
+    {
+
```
shargon, line 11:
```suggestion
```

src/Neo.Compiler.CSharp/MethodConvert/SourceConvert.cs
```diff
@@ -42,7 +43,19 @@ private void ConvertSource(SemanticModel model)
...
             case BaseMethodDeclarationSyntax syntax:
                 if (syntax.Body is null)
+                {
                     ConvertExpression(model, syntax.ExpressionBody!.Expression);
+                    // If the method has no return value,
+                    // but the expression body has a return value, example: a+=1;
+                    // drop the return value
+                    // Problem:
+                    //   public void Test() => a+=1; // this will push a int value to the stack
+                    //   public void Test() { a+=1; } // this will not push value to the stack
```
shargon, line 53: This case was not affected?

Jim8y: Only public void Test() => a+=1;

Cause => itself represents a return expression. You either take it or not, but since we do not verify the return type, void or int was the same before this pr.

shargon (approved):

Issue #992, opened 2024-03-14
NCCS: Helper.GetPatternType() not supported ByteString and BigInteger


chenzhitong:
https://github.com/neo-project/neo-devpack-dotnet/blob/4661dd2c28ea23a3ed1aa3082db0c1ca340ae81d/src/Neo.Compiler.CSharp/Helper.cs#L126-L137

The code above, looks like it supports `ByteString` and `BigInteger`, but in fact, if it's written this way in the contract, he'll report an error somewhere else.

```c#
public void M(object o1)
{
    switch (o1)
    {
        case byte[]: break;       // Compile Successfully
        case string: break;       // Compile Successfully
        case bool: break;         // Compile Successfully
        case ByteString: break;   //Compile Failure
        case BigInteger: break;   //Compile Failure
    }
}
```

<img width="1431" alt="image" src="https://github.com/neo-project/neo-devpack-dotnet/assets/6295602/4482c833-2773-4244-a4ef-38913ca53948">
This method(ConvertIdentifierNameExpression) throws an exception

Hecate2:
I think it is not supported by the syntax analyzer [roslyn](https://github.com/dotnet/roslyn/blob/f520a73885f163f6b9f08b9c64eb0fa381ca60ed/src/Compilers/CSharp/Portable/Syntax/Syntax.xml#L2904-L2950) to judge whether `o1` is of type `Class` by writing `case Class:`.
By writing `case ByteString:`, we require the variable `o1` to equal the class `ByteString`.
`case ByteString _:` is OK

shargon:
@chenzhitong could you create the unit test contract where it fail in order to fix it easier?

PR #994, opened 2024-03-14
[Compiler Fix] https://github.com/neo-project/neo-devpack-dotnet/issues/992

Jim8y:
Closes https://github.com/neo-project/neo-devpack-dotnet/issues/992

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Expression/IdentifierNameExpression.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Expression/IdentifierNameExpression.cs
@@ -93,6 +93,10 @@ partial class MethodConvert
                     Call(model, property.GetMethod!);
                 }
                 break;
+            case ITypeSymbol type:
+                IsType(type.GetPatternType());
+                Push(true);
+                break;
             default:
                 throw new CompilationException(expression, DiagnosticId.SyntaxNotSupported, $"Unsupported symbol: {symbol}");
         }
--- a/tests/Neo.Compiler.CSharp.TestContracts/Contract_Pattern.cs
+++ b/tests/Neo.Compiler.CSharp.TestContracts/Contract_Pattern.cs
@@ -70,11 +70,9 @@ namespace Neo.Compiler.CSharp.UnitTests.TestClasses
         {
             return t switch
             {
-                byte[] => 0,
-                string => 1,
-                bool => 2,
-                ByteString => 3,
-                BigInteger => 4,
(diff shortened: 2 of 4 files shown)


tests/Neo.Compiler.CSharp.TestContracts/Contract_Pattern.cs
```diff
@@ -65,14 +66,15 @@ public static void TestDeclarationPattern()
...
             {
-                case byte[]: break;
-                case string: break;
-                case bool: break;
-            }
+                byte[] or string or ByteString => 0,
+                bool => 1,
+                BigInteger => 2,
+                _ => 5
+            };
```
AnnaShaleva, line 77: It would be nice not to touch the existing test case, but create a new one, and add tests for both. Because sometimes (not in this case) some old code may stop working after new changes, especially in the compiler.

src/Neo.Compiler.CSharp/MethodConvert/Expression/IdentifierNameExpression.cs
```diff
@@ -93,6 +93,10 @@ private void ConvertIdentifierNameExpression(SemanticModel model, IdentifierName
                     Call(model, property.GetMethod!);
                 }
                 break;
+            case ITypeSymbol type:
+                IsType(type.GetPatternType());
+                Push(true);
```
AnnaShaleva, line 98: Just wondering, why it's true?

Jim8y: cause its equal check following this.

tests/Neo.Compiler.CSharp.TestContracts/Contract_Pattern.cs
chenzhitong: > It would be nice not to touch the existing test case, but create a new one, and add tests for both. Because sometimes (not in this case) some old code may stop working after new changes, especially in the compiler.

This is the test case that was just added yesterday to cover this code (GetPatternType)

shargon: agree with @AnnaShaleva , we should add tests, not touch the previous on

Jim8y: Fixed

Jim8y:
All fixed

shargon (approved):

PR #995, opened 2024-03-14
[Compiler Add ]Recursive pattern

Jim8y:
This pr adds `RecursivePatternSyntax` to compiler

```C#
newOwner is { IsValid: true, IsZero:false}
```

### DIFF
--- a/tests/Neo.Compiler.CSharp.TestContracts/Contract1.cs
+++ b/tests/Neo.Compiler.CSharp.TestContracts/Contract1.cs
@@ -1,3 +1,5 @@
+using Neo.SmartContract.Framework;
+
 namespace Neo.Compiler.CSharp.UnitTests.TestClasses
 {
     public class Contract1 : SmartContract.Framework.SmartContract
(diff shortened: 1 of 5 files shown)


tests/Neo.Compiler.CSharp.TestContracts/Contract1.cs
```diff
@@ -1,3 +1,5 @@
+using Neo.SmartContract.Framework;
+
 namespace Neo.Compiler.CSharp.UnitTests.TestClasses
```
shargon, line 3:
```suggestion
namespace Neo.Compiler.CSharp.UnitTests.TestClasses
```

tests/Neo.Compiler.CSharp.TestContracts/Contract1.cs
```diff
@@ -40,5 +42,6 @@ public static int testArgs4(int a, int b)
             a = a + 2;
             return a + b;
         }
+
```
shargon, line 45:
```suggestion
```

PR #993, opened 2024-03-14
[Framework and Examples] Update examples

Jim8y:
This pr adds an Oracle method name and update the Examples

PR #998, opened 2024-03-19
use abstract method to define static oracle callback

Jim8y:
This pr updates the IOracle to define the oracle callback interface as static https://github.com/neo-project/neo-devpack-dotnet/pull/993#discussion_r1529952710


cschuchardt88 (approved):

shargon:
But why we want to use static interface in oracle, and not in others? we should use the same way

tests/Neo.SmartContract.Framework.UnitTests/TestingArtifacts/Contract_IOracle.cs
cschuchardt88, line 1: I thought we were using `block namespace` scoping?

Jim8y:
@shargon i am sorry to say this but i think it is a mess at all cause static or non-static all works as interface..... People can use it as static or non-static... how do we make an interace that both works for static and non-static @cschuchardt88

shargon:
> @shargon i am sorry to say this but i think it is a mess at all cause static or non-static all works as interface..... People can use it as static or non-static... how do we make an interace that both works for static and non-static @cschuchardt88 

We should try to focus on instance

Jim8y:
Use instance interface instead.

PR #1016, opened 2024-04-11
update copyright for nccs and framework

chenzhitong:


src/Neo.Compiler.CSharp/Optimizer/Analysers/InstructionCoverage.cs
```diff
@@ -1,3 +1,14 @@
+// Copyright (C) 2015-2024 The Neo Project.
+//
+// Helper.cs file belongs to the neo project and is free
```
Jim8y, line 3: file name is not right

src/Neo.Compiler.CSharp/Optimizer/Analysers/JumpTarget.cs
```diff
@@ -1,3 +1,14 @@
+// Copyright (C) 2015-2024 The Neo Project.
+//
+// Helper.cs file belongs to the neo project and is free
```
Jim8y, line 3:
```suggestion
// JumpTarget.cs file belongs to the neo project and is free
```

Jim8y (approved):

Jim8y:
@shargon please review

shargon (approved):

PR #1018, opened 2024-04-11
Migrate old ut

shargon:
Last migration unit test part, it will be divided in different pr because it's too long, included in this PR:
- Migrate BasicBlock optimizer unit tests.
- Moved old test to OldEngine namespace.
- Created artifacts for future PR.
- Separate contracts that are designed not to be compiled into folders.


tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            // Define paths
+
+            string testContractsPath = Path.GetFullPath("../../../../Neo.Compiler.CSharp.TestContracts/Neo.Compiler.CSharp.TestContracts.csproj");
+            string artifactsPath = Path.GetFullPath("../../../TestingArtifacts");
+            var root = Path.GetPathRoot(testContractsPath) ?? "";
+
+            // Compile
+
+            var results = new CompilationEngine(new CompilationOptions()
```
shargon, line 40: @Jim8y could you help me here? I'm not able to compile whole project together

> System.ArgumentException: SyntaxTree is not part of the compilation (Parameter 'syntaxTree')

Jim8y: I will check.

shargon: I think that is because there are contract designed to fail during compilation

shargon:
@Jim8y could you help me here?

shargon:
I will open a new one

PR #1028, opened 2024-04-19
Migrate old ut 

shargon:
Last migration unit test part, it will be divided in different pr because it's too long, included in this PR:

- Migrate BasicBlock optimizer unit tests.
- Moved old test to OldEngine namespace.
- Created artifacts for future PR.
- Removed wrong contracts.


shargon:
@Jim8y Still receiving compilation errors in `EnsureArtifactsUpToDateInternal` :S

Jim8y:
Weird cause there were ordinary compile errors, why they did not show in the past but appear here? Like Interface not supported.

Jim8y:
I get it now, cause many test contracts that demonstrate invalid syntax in neo contract were not compiled if there is no unit test against them. but in your new test logic, all contract will be compiled, thus many many compile errors appear.

Jim8y:
Remainning issue is in the working/bin/sc generated file, i have no idea how they are generated by your test engine, but i think you can fix them now

cschuchardt88 (changes requested):
Add the following to all **TEST PROJECTS** `*.csproj` files.
```xml
<IsPackable>false</IsPackable>
<IsTestProject>true</IsTestProject>
```

Example
---
```xml
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <IsPackable>false</IsPackable>
    <IsTestProject>true</IsTestProject>
  </PropertyGroup>
</Project>
```

tests/Neo.Compiler.CSharp.UnitTests/OldEngine/UnitTest_Types.cs
```diff
@@ -21,22 +21,22 @@ public class UnitTest_Types
...
+            // using var testengine = new TestEngine();
+            // Assert.IsFalse(testengine.AddEntryScript(Utils.Extensions.TestContractRoot + "Contract_Types_Decimal.cs").Success);
         }
 
         [TestMethod]
         public void double_Test()
         {
-            using var testengine = new TestEngine();
-            Assert.IsFalse(testengine.AddEntryScript(Utils.Extensions.TestContractRoot + "Contract_Types_Double.cs").Success);
+            // using var testengine = new TestEngine();
```
shargon, line 38: @Jim8y  Why these tests doesn't work?

shargon:
> I get it now, cause many test contracts that demonstrate invalid syntax in neo contract were not compiled if there is no unit test against them. but in your new test logic, all contract will be compiled, thus many many compile errors appear.

Could you move them to notWorking folder, I want to compile only the other folder

Jim8y:
> Could you move them to notWorking folder? I want to compile only the other folder

I am not sure it has any meaning or not to show `invalid` cases. Those contracts can not be compiled, but only creating confusion, such as there were an interface contract and corresponding unit tests, but actually neo does not support interface at all.

tests/Neo.Compiler.CSharp.TestContracts/Working/Contract_MemberAccess.cs
```diff
@@ -5,19 +5,17 @@ namespace Neo.Compiler.CSharp.UnitTests.TestClasses
...
-            }
+            var my = new MyClass();
+            Runtime.Log(my.Data1.ToString());
+            Runtime.Log(MyClass.Data2);
+            //Runtime.Log(MyClass.Data3.ToString());
+            Runtime.Log(my.Data4);
+            Runtime.Log(my.Method());
         }
 
+
```
shargon, line 18:
```suggestion
```

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        internal static CompilationContext[] EnsureArtifactsUpToDateInternal()
+        {
+            if (DebugInfos.Count > 0) return compilationContexts; // Maybe a UT call it
+
+            // Define paths
+
+            var artifactsPath = Path.GetFullPath("../../../TestingArtifacts");
+            var testContractsPath = Path.GetFullPath("../../../../Neo.Compiler.CSharp.TestContracts/Neo.Compiler.CSharp.TestContracts.csproj");
+            var root = Path.GetPathRoot(testContractsPath) ?? "";
```
cschuchardt88, line 37: You should use `FileInfo`
```csharp
var fileInfo = new FileInfo(fileName);
```

Jim8y (approved):

Issue #1027, opened 2024-04-19
Move examples to a diferent sln


shargon:
After merge examples in this repository, every compilation spend a lot of time, it could be good if we move them to a diferent project (sln)

Jim8y:
make the workflow runs based on the commit. if no change to the examples, just not build it.

shargon:
> make the workflow runs based on the commit. if no change to the examples, just not build it.

My problem is with local compilations

cschuchardt88:
@shargon 

https://github.com/neo-project/neo-devpack-dotnet/pull/1029#issuecomment-2067602831

Jim8y:
> > make the workflow runs based on the commit. if no change to the examples, just not build it.
> 
> My problem is with local compilations

pr solve that as well. It takes very long no matter where you test them.

PR #1029, opened 2024-04-20
[Examples] remove examples from workflow and create a solution for examples.

Jim8y:
Closes https://github.com/neo-project/neo-devpack-dotnet/issues/1027


Jim8y:
because examples will create artifacts for every example, it takes a long time, thus need to remove from the workflow. we can run it locally.

cschuchardt88:
@Jim8y What is the problem with using
---
image

cschuchardt88 (changes requested):
Please don't change the formatting...

examples/Example.SmartContract.ContractCall.UnitTests/Example.SmartContract.ContractCall.UnitTests.csproj
```diff
@@ -1,34 +1,34 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
+    <PropertyGroup>
```
shargon, line 3: Agree with @cschuchardt88 and the format

superboyiii (approved):

superboyiii:
Looks good on my local environment.

vncoelho (approved):

shargon (commented):
Good change

PR #1030, opened 2024-04-21
Support record

Ashuaidehao:
Support record struct.


shargon (dismissed):

Jim8y (dismissed):

shargon:
Please wait for https://github.com/neo-project/neo-devpack-dotnet/pull/1028

src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
```diff
@@ -226,12 +226,31 @@ private void ConvertExpression(SemanticModel model, ExpressionSyntax syntax, Syn
...
+    /// </summary>
+    /// <param name="model"></param>
+    /// <param name="expression"></param>
+    private void ConvertWithExpressionSyntax(SemanticModel model, WithExpressionSyntax expression)
+    {
+        //load record
+        ConvertExpression(model, expression.Expression);
+        //clone record struct
+        AddInstruction(new Instruction { OpCode = OpCode.UNPACK });
+        AddInstruction(new Instruction { OpCode = OpCode.PACKSTRUCT });
```
AnnaShaleva, line 249: Why not to use `DUP`+`NIP` instead? It's cheaper.

Jim8y: Record is readonly type, using DUP NIP will make it updatable.

Jim8y (dismissed):

Jim8y (approved):

src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
Ashuaidehao: `with` operation should return a new instance with same content. This is the simplest way to implement this since I cannot call `clone` method directly.

tests/Neo.Compiler.CSharp.TestContracts/Contract_Record.cs
```diff
@@ -0,0 +1,56 @@
+using Neo.SmartContract.Framework.Native;
+using Neo.SmartContract.Framework.Services;
+using System;
+using System.Numerics;
+
+namespace Neo.Compiler.CSharp.UnitTests.TestClasses
```
shargon, line 6:
```suggestion
namespace Neo.Compiler.CSharp.TestContracts;
```

shargon (commented):
I will move it to the new ut system, wait one minute please

shargon (approved):

PR #1032, opened 2024-04-23
update csproj format

Jim8y:


cschuchardt88 (changes requested):
Add to Test `*.csproj`
```xml
<IsPackable>false</IsPackable>
<IsTestProject>true</IsTestProject>
```

Also do some research for `msbuild` 
https://learn.microsoft.com/en-us/visualstudio/msbuild/walkthrough-using-msbuild?view=vs-2022#build-items

Jim8y:
> Add to Test `*.csproj`
> 
> ```
> <IsPackable>false</IsPackable>
> <IsTestProject>true</IsTestProject>
> ```
> 
> Also do some research for `msbuild` https://learn.microsoft.com/en-us/visualstudio/msbuild/walkthrough-using-msbuild?view=vs-2022#build-items

This pr apparently only update the format. Other issues will not be addressed here in case of being blocked.

shargon (approved):

PR #1035, opened 2024-05-06
Native: extend CryptoLib's `verifyWithECDsa` with compatible `NamedCurveHash` argument

AnnaShaleva:
Port protocol-related part of https://github.com/neo-project/neo/pull/3209.

Note, that build and tests are failing because we need to update `neo` git submodule. We can merge https://github.com/neo-project/neo/pull/3209 firstly, and then update the `neo` submodule onto the fresh master. Then I'll rebase this PR and check that tests are passed.


Jim8y:
@AnnaShaleva anna, we are not going to add secp256k1 support here. Core only do verification, no construction.

AnnaShaleva:
This code is needed for verification only. We extend the native CryptoLib's `verifyWithECDsa` parameter, so that both Sha256 and Keccak256 hashes may be checked. We need this PR.

Jim8y (approved):

vncoelho (commented):
I will wait until the dependent PR is merged.

src/Neo.SmartContract.Framework/Native/NamedCurveHash.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// The secp256k1 curve and Keccak256 hash algorithm.
+        /// </summary>
+        secp256k1Keccak256 = 122,
+
+        /// <summary>
+        /// The secp256r1 curve, which known as prime256v1 or nistP-256, and Keccak256 hash algorithm.
+        /// </summary>
+        secp256r1Keccak256 = 123
```
shargon, line 39: Why don't just add these values to the previous enum?

cschuchardt88: Hash isn't related to a/the curve. So I don't know why you need `secp256k1` in the name. `Keccak-256` is a hash function. If we decide to support on `secp256r1` curves. Than you would have to add `secp256r1SHA256`, `secp256k1Keccak256`, etc; should only be two values `Keccak256` and `SHA256` for `hasher enum`. Why do we need all this extra stuff?

If it's to indicate a `prefix` within the hash than a more appropriate name should be used.

Jim8y (dismissed):

AnnaShaleva:
Can’t be merged, I need to update dependencies.

vncoelho:
I imagined that, fetch git as well

AnnaShaleva:
Updated, ready for review and merge.

cschuchardt88 (approved):

Jim8y (approved):

PR #1040, opened 2024-05-13
Fixed Workflow for nuget publish

cschuchardt88:
# Release
### Removed solution projects for workflow to publish nuget packages, so that we don't republish `neo` repo packages. Also publish to nuget job was failing.

image


.github/workflows/main.yml
```diff
@@ -189,7 +189,7 @@ jobs:
     - name: Publish to NuGet
       if: steps.check_tag.outputs.statusCode == '404'
       run: |
-        dotnet pack -o out -c Release
-        dotnet nuget push "out/*.nupkg" -s https://api.nuget.org/v3/index.json -k ${NUGET_TOKEN} --skip-duplicate
+        dotnet pack release.sln -o pub -c Release
```
shargon, line 192: Why create a different sln?

shargon: If it's required to remove a specific project, I prefer to remove the project from the current sln in workflow

cschuchardt88: It would be all submodules. Which is all the dependencies from the `dependencies` folder (all `neo` repo projects).

shargon: one sln more is weird, also for new devs, it should be solved with the workflow

shargon (approved):

Jim8y (approved):

PR #1041, opened 2024-05-19
wildcard to any and add nep24

Jim8y:
This pr use Any to replace wildcard, and also add support to NEP24.

Also remove dedundent file


cschuchardt88 (dismissed):
I don't why you guys enabled `nullable` in `*.csproj` file, when your not using it. You have all these `Possible null reference return.` warning.

src/Neo.SmartContract.Framework/Interfaces/INEP24.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Numerics;
+
+namespace Neo.SmartContract.Framework.Interfaces;
+
+/// <summary>
+/// Interface of supporting royalty information for Non-Fungible Tokens (NFTs)
+/// </summary>
+public interface INep24
+{
+
```
shargon, line 21:
```suggestion
```

shargon (approved):

PR #1046, opened 2024-05-22
Coveralls and packages update

shargon:
- Updated coveralls git action.
- Consolidate nuget versions.


Jim8y (approved):

.github/workflows/main.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 env:
   DOTNET_VERSION: 8.0.x
+  EXCLUDE_COVERAGE: "[Neo.SmartContract.TestEngine]*,[Neo.Compiler.CSharp.UnitTests]*,[Neo]*,[Neo.IO]*,[Neo.Json]*,[Neo.VM]*,[Neo.Extensions]*,[Neo.Cryptography.BLS12_381]*"
```
vncoelho, line 10: very good.

vncoelho (changes requested):
Standardize this commands as well:

`/p:CoverletOutput=${{ github.workspace }}/coverage-join/ \`
`/p:MergeWith=${{ github.workspace }}/coverage-join/coverage.json \`

This will surely avoid us to make mistakes on that with copy and past. Thus, the standard will be default.

shargon:
> Standardize this commands as well:
> 
> `/p:CoverletOutput=${{ github.workspace }}/coverage-join/ \` `/p:MergeWith=${{ github.workspace }}/coverage-join/coverage.json \`
> 
> This will surely avoid us to make mistakes on that with copy and past. Thus, the standard will be default.

I think that save 3 lines in one is not an issue related to this pr

vncoelho:
> I think that save 3 lines in one is not an issue related to this pr

This is aligned with the change you made for the ENV variable for EXCLUDED_FILES.

vncoelho:
> > > I think that save 3 lines in one is not an issue related to this pr
> > 
> > 
> > This is aligned with the change you made for the `ENV` variable for `EXCLUDED_FILES`.
> 
> but excluded is the same for all calls, the others it depends if it's the first call, the middle or the last one

just the first, the other need to merge.
The first also needs to explicit CoverletOutput.

vncoelho:
> > > I think that save 3 lines in one is not an issue related to this pr
> > 
> > 
> > This is aligned with the change you made for the `ENV` variable for `EXCLUDED_FILES`.
> 
> but excluded is the same for all calls, the others it depends if it's the first call, the middle or the last one

If you want I can push it here.

vncoelho:
> > > > I think that save 3 lines in one is not an issue related to this pr
> > > 
> > > 
> > > This is aligned with the change you made for the `ENV` variable for `EXCLUDED_FILES`.
> > 
> > but excluded is the same for all calls, the others it depends if it's the first call, the middle or the last one
> 
> If you want I can push it here.

I will solve asap.

.github/workflows/main.yml
```diff
@@ -74,7 +75,7 @@ jobs:
         /p:CollectCoverage=true \
         /p:CoverletOutput=${{ github.workspace }}/coverage-join/ \
         /p:MergeWith=${{ github.workspace }}/coverage-join/coverage.json \
-        /p:Exclude=\"[Neo.SmartContract.TestEngine]*,[Neo.Compiler.CSharp.UnitTests]*,[Neo]*,[Neo.IO]*,[Neo.Json]*,[Neo.VM]*,[Neo.Extensions]*,[Neo.Cryptography.BLS12_381]*\"
+        /p:Exclude=\"${{ env.EXCLUDE_COVERAGE }}\"
```
vncoelho, line 79: Why do we need `-e "COVERAGE_MERGE_JOIN=${{ github.workspace }}/coverage-join/coverage.json" \` and `/p:MergeWith=${{ github.workspace }}/coverage-join/coverage.json \` here in this workflow?

It looks a little more complex than in neo-core.

vncoelho (commented):
@shargon ,

this file is a little strange here compared to the neo-core.
Should not we standardize this? There are strange additional parameters which I am not sure they are necessary.

vncoelho:
Commit [c1dbb2e](https://github.com/neo-project/neo-devpack-dotnet/pull/1046/commits/c1dbb2e3d83d30b8c8249abcdf614f972126a63f) unified and removed some "Probably" extra parameters that were duplicated

.github/workflows/main.yml
```diff
@@ -7,6 +7,10 @@ on:
 
 env:
   DOTNET_VERSION: 8.0.x
+  DOTNET_TEST_PARAMETERS: --no-build  /p:CollectCoverage=true -l "console;verbosity=detailed"
+  EXCLUDE_COVERAGE: "[Neo.SmartContract.TestEngine]*,[Neo.Compiler.CSharp.UnitTests]*,[Neo]*,[Neo.IO]*,[Neo.Json]*,[Neo.VM]*,[Neo.Extensions]*,[Neo.Cryptography.BLS12_381]*"
```
vncoelho, line 11: Maybe change to the same way I pushed for the other ones

vncoelho, line 11:
```suggestion
EXCLUDE_COVERAGE: "/p:Exclude=[Neo.SmartContract.TestEngine]*,[Neo.Compiler.CSharp.UnitTests]*,[Neo]*,[Neo.IO]*,[Neo.Json]*,[Neo.VM]*,[Neo.Extensions]*,[Neo.Cryptography.BLS12_381]*"
```

vncoelho (dismissed):
@shargon ,

The file was, indeed, a little strange and not very well organized.
There were additional parameters. 
Check the last two commits.

I think that it is cleaner now.

As a last modification for now, maybe apply my last suggestion.

.github/workflows/main.yml
```diff
@@ -7,6 +7,10 @@ on:
 
 env:
   DOTNET_VERSION: 8.0.x
+  DOTNET_TEST_PARAMETERS: --no-build  /p:CollectCoverage=true -l "console;verbosity=detailed"
```
shargon, line 10:
```suggestion
  DOTNET_TEST_PARAMETERS: --no-build /p:CollectCoverage=true -l "console;verbosity=detailed"
```

.github/workflows/main.yml
shargon: It is because we need to merge it with the smart contract coverage

shargon: It is because we need to merge it with the smart contract coverage

.github/workflows/main.yml
shargon: @vncoelho please use that way, otherwise now looks different

vncoelho: ok, changing here

vncoelho (dismissed):

vncoelho:
I will just wait test.

vncoelho (approved):

PR #1052, opened 2024-05-26
Migrate more uts

shargon:


Jim8y (commented):
why this change?

shargon:
> why this change?

The smart contract or which one?

shargon:
@Jim8y if is the SmartContract change, there are some UT that send the StackItem type as an argument, We can change the unit test, but I think that is good to be able to do that

tests/Neo.Compiler.CSharp.TestContracts/Contract_NULL.cs
```diff
@@ -5,27 +5,27 @@ namespace Neo.Compiler.CSharp.TestContracts
 {
     public class Contract_NULL : SmartContract.Framework.SmartContract
     {
-        public static bool IsNull(byte[]? value)
+        public static bool IsNull(object? value)
```
Jim8y, line 8: I mean this change

Jim8y (approved):

tests/Neo.Compiler.CSharp.TestContracts/Contract_NULL.cs
shargon: It checked that the object is null, but it receive a buffer, we can revert it if you prefer

Jim8y: i dont like non related change. but its ok.

PR #1054, opened 2024-05-31
TestEngine: Migrate uts

shargon:
- switch contracts was joined in the same smart contract


Jim8y:
just a recommendation, maybe we should use biginteger as much as possible. its calculation is much cheaper.

Jim8y (approved):

shargon:
> just a recommendation, maybe we should use biginteger as much as possible. its calculation is much cheaper.

where?

Jim8y:
> > just a recommendation, maybe we should use biginteger as much as possible. its calculation is much cheaper.
> 
> where?

Anywhere that has no overflow concerns. Not necessarily related to this pr. Maybe another pr that we can update all examples and tests, use BigInteger instead.

vncoelho (approved):

cschuchardt88 (approved):

PR #1056, opened 2024-06-05
[Neo Framework]add IWithdrawable

Jim8y:
This pr standarize the withdrawable interface and add its comments.


cschuchardt88 (approved):

src/Neo.SmartContract.Framework/Interfaces/IWithdrawable.cs
```diff
@@ -0,0 +1,39 @@
...
+// modifications are permitted.
+
+#nullable enable
+
+namespace Neo.SmartContract.Framework.Interfaces;
+
+/// <summary>
+/// Interface of method that indicate a contract can be withdrawn.
+/// </summary>
+public interface IWithdrawable
```
shargon, line 19: Iverificable? Verify is not only for withdraw

Jim8y: This is for IWithdrawable, not for verify. Just verify happens to be IWithdrawable method. 

For the ease of helping developers to implement contract that is withdrawable, not a contract that is verificable.

shargon: But it won't be better if we create IVerify? that it can be used for multiple things, like withdraw GAS and NEO

Jim8y: Well yes, but if its purpose is withdrawing, why call it verify?

Jim8y: @shargon my suggestion is create another pr that create IVerify, then IWithdrawable  inherits IVerify

Jim8y: @shargon Go ahead with what you want. But please make sure to have detailed comments for IWithdrawable.

PR #1057, opened 2024-06-05
Update SampleRoyaltyNEP11Token.cs

chenzhitong:
I think it makes more sense that the default `royaltyAmount` is scaled by `salePrice`, otherwise a deal with NEO would be 700 NEO and a deal with GAS would be 700 Datoshi


shargon (approved):

examples/Example.SmartContract.SampleRoyaltyNEP11Token/SampleRoyaltyNEP11Token.cs
```diff
@@ -170,7 +170,7 @@ public static void SetRoyaltyInfo(ByteString tokenId, Map<string, object>[] roya
             {
                 var royaltyInfo = new Map<string, object>();
                 royaltyInfo["royaltyRecipient"] = InitialRecipient;
-                royaltyInfo["royaltyAmount"] = InitialFactor;
+                royaltyInfo["royaltyAmount"] = salePrice / InitialFactor;
```
Jim8y, line 173: Can this be 0?

Jim8y (approved):

examples/Example.SmartContract.SampleRoyaltyNEP11Token/SampleRoyaltyNEP11Token.cs
chenzhitong: I think it makes sense to round down.

Jim8y (approved):

shargon (approved):

PR #1060, opened 2024-06-06
TestEngine: Migrate UTS

shargon:
Problems found:

- https://github.com/neo-project/neo-devpack-dotnet/blob/5334d70d8375fb86f8f68554e5f3454ad383a544/tests/Neo.Compiler.CSharp.TestContracts/Contract_Types_ECPoint.cs#L15 this cast is not the expected as ECPoint.ToString, it converts the ECPoint to byte[] representation, and when is decoded, it throw an exception, `DecoderFallbackException` because is not a valid UTF8 string
- I added a permision in https://github.com/neo-project/neo-devpack-dotnet/pull/1060/files#diff-91f0ea4f6691d93c4ed6f90c777c9ea646fad8ba76100f594370b2f609f15dc4R10 because otherwise is not possible to call a created contract, I don't know how it was possible to work before
- Added `StringInterpreter` for set the parse method used for string types.


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Types_ECPoint.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        [TestMethod]
+        public void ECPoint_test()
+        {
+            Assert.IsFalse(Contract.IsValid(InvalidECPoint.InvalidLength));
+            Assert.IsFalse(Contract.IsValid(InvalidECPoint.InvalidType));
+            Assert.ThrowsException<TestException>(() => Contract.IsValid(InvalidECPoint.Null));
+            Assert.IsTrue(Contract.IsValid(Cryptography.ECC.ECPoint.Parse("024700db2e90d9f02c4f9fc862abaca92725f95b4fddcc8d7ffa538693ecf463a9", Cryptography.ECC.ECCurve.Secp256r1)));
+
+            Assert.AreEqual("024700db2e90d9f02c4f9fc862abaca92725f95b4fddcc8d7ffa538693ecf463a9", Contract.Ecpoint2String()); // ERROR: Is converted to string, but in bytes
```
shargon, line 22: @Jim8y this doesn't work :S, I don't know why work before also

Jim8y: I will check

Jim8y:
git --no-optional-locks -c color.branch=false -c color.diff=false -c color.status=false -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree commit -q -F /var/folders/ry/g6fy9gt1599246fzxndmrl700000gn/T/SourceTreeTemp.A0bQto 


git --no-optional-locks -c color.branch=false -c color.diff=false -c color.status=false -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree push -v --tags --set-upstream upstream refs/heads/migrate_uts:refs/heads/migrate_uts 
Pushing to github.com:neo-project/neo-devpack-dotnet.git
ERROR: Permission to neo-project/neo-devpack-dotnet.git denied to Jim8y.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Completed with errors, see above

Jim8y:
Screenshot 2024-06-06 at 22 53 57
Screenshot 2024-06-06 at 22 54 10


The issue is that you are trying to convert byte array directly to string. While original code returns buffer, though the return type is string.

shargon:
> Screenshot 2024-06-06 at 22 53 57 Screenshot 2024-06-06 at 22 54 10
> The issue is that you are trying to convert byte array directly to string. While original code returns buffer, though the return type is string.

seems weird to me return in a different format if it FAULT

Jim8y:
> > Screenshot 2024-06-06 at 22 53 57 Screenshot 2024-06-06 at 22 54 10
> > The issue is that you are trying to convert byte array directly to string. While original code returns buffer, though the return type is string.
> 
> seems weird to me return in a different format if it FAULT

neo does not.really check the return type, neighter will it convert it to specified type, but your test engine has this system. as you can see the original tests, its using getspan then to hex, instead of directly getstring.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Types_ECPoint.cs
```diff
@@ -0,0 +1,30 @@
...
+        public void ECPoint_test()
+        {
+            Assert.IsFalse(Contract.IsValid(InvalidECPoint.InvalidLength));
+            Assert.IsFalse(Contract.IsValid(InvalidECPoint.InvalidType));
+            Assert.ThrowsException<TestException>(() => Contract.IsValid(InvalidECPoint.Null));
+            Assert.IsTrue(Contract.IsValid(Cryptography.ECC.ECPoint.Parse("024700db2e90d9f02c4f9fc862abaca92725f95b4fddcc8d7ffa538693ecf463a9", Cryptography.ECC.ECCurve.Secp256r1)));
+
+            Engine.StringEncoder = Encoding.ASCII;
+
+            Assert.AreEqual("0247003f2e3f3f3f2c4f3f3f623f3f3f27253f5b4f3f3f3f7f3f533f3f3f3f633f", Encoding.ASCII.GetBytes(Contract.Ecpoint2String()!)?.ToHexString());
```
shargon, line 25: @Jim8y any idea about why now this value is different? 🤔

Jim8y:
@shargon

shargon:
@Jim8y ready to merge

shargon:
@Jim8y merge?

Jim8y (approved):

PR #1065, opened 2024-06-28
[DevPack Compiler Optimize] fix push bool

Jim8y:
This pr mainly fix the push bool issue, previously, it pushes 1 or 0 and then convert it to boolean, but actually we can directly use PUSHT and PUSHF.


src/Neo.Compiler.CSharp/MethodConvert/StackHelpers.cs
```diff
@@ -22,8 +22,8 @@ partial class MethodConvert
 {
     private void Push(bool value)
     {
-        AddInstruction(value ? OpCode.PUSH1 : OpCode.PUSH0);
```
Jim8y, line 25: @shargon do you have any idea why this part was like this?

shargon: It was less optimal, but should work exactly the same

src/Neo.Compiler.CSharp/MethodConvert/StackHelpers.cs
```diff
@@ -22,8 +22,8 @@ partial class MethodConvert
...
-        ChangeType(VM.Types.StackItemType.Boolean);
+        AddInstruction(value ? OpCode.PUSHT : OpCode.PUSHF);
+        // ChangeType(VM.Types.StackItemType.Boolean);
```
shargon, line 26:
```suggestion
```

shargon (approved):

cschuchardt88 (approved):

steven1227:
Awesome fix!

PR #1071, opened 2024-07-12
Set optimization level from command line

Hecate2:
It seems command line users cannot trigger --no-optimize or the experimental optimizers🤣, although they can be set in tests.
To test this PR, use the following flags:
` --debug --ex-optimize --assembly`
` --debug --no-optimize --assembly`
` --debug --no-optimize --ex-optimize --assembly` (should fail)
` --debug --ex-optimize --no-optimize --assembly` (should fail)
and read `bin/sc/{contract}.nef.txt` to see whether the contract has been correctly optimized.


src/Neo.Compiler.CSharp/CompilationOptions.cs
```diff
@@ -33,6 +33,30 @@ public enum OptimizationType : byte
         public OptimizationType Optimize { get; set; } = OptimizationType.Basic;
         public bool Checked { get; set; }
         public bool NoInline { get; set; }
+        protected bool _NoOptimize;
+        public bool NoOptimize
+        {
+            get => _NoOptimize; set
+            {
+                _NoOptimize = value;
```
shargon, line 41:
```suggestion
        public bool NoOptimize
        {
            get => Optimize == OptimizationType.None; set
            {
```

Jim8y (dismissed):

shargon:
> It seems command line users cannot trigger --no-optimize 

What about use --optimize=none ?

Hecate2:
> > It seems command line users cannot trigger --no-optimize 
> 
> What about use --optimize=none ? 

That's great for me

shargon:
I think that we should instruct the user to use --optimize=Level

Hecate2:
` --debug --optimize=All --assembly`
` --debug --optimize=Basic --assembly`
` --debug --optimize=None --assembly`

shargon (approved):

Jim8y (approved):

PR #1076, opened 2024-07-23
[Neo Devpack Tests] add gas consume to the test

Jim8y:
This pr adds gas consumption to the tests, by having gas cost, we can see which syntax consumes how many datoshis, and can further help us learn and optimize the compiler.


Jim8y:
@shargon why every test consumes 10_0xxx_xxxx? Do you know which operation costs that 10_0000_0000?

shargon:
I have in mind a different approach for this with an Attribute like [GasCost(xxx)]. Let me check it...

Jim8y:
some test may contain multiple contract calls, attribute may work for some tests. will not be conflict with this one.

shargon:
It must be the native contract initialization, use `CreateGasWatcher` instead

shargon:
> > It must be the native contract initialization, use `CreateGasWatcher` instead
> 
> Sorry, i dont get it, can you give me an example?

Done

shargon:
Maybe we can substract the native cost initialization from the engine

shargon:
@Jim8y take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/1078

shargon:
Why still so high?

Jim8y:
@shargon  fixed and updated.

tests/Neo.SmartContract.Testing.UnitTests/NativeArtifactsTests.cs
```diff
@@ -25,9 +25,9 @@ public void TestInitialize()
 
             // Check symbols
 
-            engine.FeeConsumed.Value = 0;
+            var fee = engine.CreateGasWatcher();
```
shargon, line 28:
```suggestion
            using var fee = engine.CreateGasWatcher();
```

src/Neo.SmartContract.Testing/TestEngine.cs
```diff
@@ -180,7 +180,7 @@ public long Fee
         /// <summary>
         /// Fee Consumed (In the unit of datoshi, 1 datoshi = 1e-8 GAS)
         /// </summary>
-        public FeeWatcher FeeConsumed { get; }
+        public FeeWatcher? FeeConsumed { get; set; }
```
shargon, line 183: I prefer to avoid the nullability, please check https://github.com/neo-project/neo-devpack-dotnet/pull/1076 for this change

Jim8y: > I prefer to avoid the nullability, please check #1076 for this change

what is the reason of avoiding nullable? performance? possible issue?

Jim8y:
@shargon good to go now

src/Neo.SmartContract.Testing/TestEngine.cs
```diff
@@ -180,7 +180,7 @@ public long Fee
         /// <summary>
         /// Fee Consumed (In the unit of datoshi, 1 datoshi = 1e-8 GAS)
         /// </summary>
-        public FeeWatcher FeeConsumed { get; }
+        public FeeWatcher FeeConsumed { get; set; }
```
shargon, line 183:
```suggestion
        public FeeWatcher FeeConsumed { get; }
```

src/Neo.SmartContract.Testing/TestEngine.cs
```diff
@@ -626,9 +625,8 @@ public StackItem Execute(Script script, int initialPosition = 0, Action<Applicat
             ApplicationEngine.Notify += ApplicationEngineNotify;
 
             // Execute
-
-            if (ResetFeeConsumed) FeeConsumed.Reset();
             beforeExecute?.Invoke(engine);
+            if (ResetFeeConsumed) FeeConsumed.Reset();
```
shargon, line 629: You don't want to take care the event in the comoutation?

Jim8y: Its before the execution, do we need to consider that as well?

shargon: it's used sometimes to prepare the script, for example, add Interop items to the stack

Jim8y: then it is not necessary, we need the gas consumtion for optimize the converter, the preparation process will not be involved.

Jim8y: This pr is getting bigger and bigger because of the coverage increase, is becoming unreviewable.

Jim8y:
This pr contains new unit tests that increases the coverage, should be merged before other prs.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ABI_Safe.cs
```diff
@@ -21,6 +21,12 @@ public void Method1Test()
             Assert.AreEqual(1, Contract.UnitTest_001());
         }
 
+        // [TestMethod]
+        // public void Method2Test()
+        // {
+        //     Assert.AreEqual(2, Contract.UnitTest_002());
+        // }
+
```
shargon, line 29: Needed?

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_IndexOrRange.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Neo.Compiler.CSharp.UnitTests
+{
+    [TestClass]
+    public class UnitTest_IndexOrRange() : TestBase<Contract_IndexOrRange>(Contract_IndexOrRange.Nef, Contract_IndexOrRange.Manifest)
+    {
+        [TestMethod]
+        public void Test_Main()
+        {
+            // Contract.TestMain();
```
shargon, line 13: should do anything, not needed for this PR

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ABI_Safe.cs
Jim8y: No, wanted to test it but the test engine wont allow me to, maybe you should investage on it.

shargon (dismissed):

vncoelho (approved):

shargon (commented):
Lgtm now

PR #1081, opened 2024-07-24
[Devpack Compiler] Nullable

Jim8y:
This pr completes the nullable.  This pr fixed and tested some of the nullable methods, but more work need to be done. So this pr can be considered to be part of the job. 

However, i am not able to complete it in one pr, its too complex and need to carefully check the compiler, will take a while. Thus, i will consider this pr as has completed its target.


shargon:
Blocked by?

Jim8y:
@shargon good to go

tests/Neo.SmartContract.Framework.UnitTests/NullableTest.cs
```diff
@@ -0,0 +1,322 @@
...
+namespace Neo.SmartContract.Framework.UnitTests
+{
+    [TestClass]
+    public class NullableTest : TestBase<Contract_Nullable>
+    {
+        public NullableTest() : base(Contract_Nullable.Nef, Contract_Nullable.Manifest)
+        {
+        }
+
+        [TestMethod]
```
shargon, line 15:
```suggestion
    public class NullableTest : TestBase<Contract_Nullable>(Contract_Nullable.Nef, Contract_Nullable.Manifest)
    {
        [TestMethod]
```

tests/Neo.SmartContract.Framework.TestContracts/Contract_Nullable.cs
```diff
@@ -0,0 +1,230 @@
...
+            return a != b && !a.Equals(b) && !b.Equals(a) && b != a;
+        }
+
+        public static bool BoolNullableEqualNull(bool? a)
+        {
+            return a == null && !a.HasValue;
+        }
+
+
+        public static bool ByteNullableToString(byte? a)
```
shargon, line 180:
```suggestion
        public static bool ByteNullableToString(byte? a)
```

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
```diff
@@ -332,12 +377,6 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
                 if (arguments is not null)
                     PrepareArgumentsForMethod(model, symbol, arguments);
                 return true;
-            //Converts the string representation of a number to its BigInteger equivalent.
-            case "System.Numerics.BigInteger.Parse(string)":
-                if (arguments is not null)
-                    PrepareArgumentsForMethod(model, symbol, arguments);
-                CallContractMethod(NativeContract.StdLib.Hash, "atoi", 1, true);
-                return true;
```
shargon, line 340: This was removed?

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
```diff
@@ -379,12 +420,78 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
...
+            case "char?.Value.get":
+            case "System.Numerics.BigInteger?.Value.get":
+                {
+                    if (instanceExpression is not null)
+                        ConvertExpression(model, instanceExpression);
+                    AddInstruction(OpCode.DUP);
+                    AddInstruction(OpCode.ISNULL);
+                    JumpTarget endTarget = new();
+                    Jump(OpCode.JMPIFNOT, endTarget);
+                    AddInstruction(OpCode.THROW);
```
shargon, line 465: if we use ? it should return a null if is null, isn't it?

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
Jim8y: Its here https://github.com/neo-project/neo-devpack-dotnet/pull/1081/files/2f04774f4a2543776eec7d064e5f6e147c05cab2#diff-c3d4fe6fb638262e5b22f1ffdd0ba20cfdac4ddae36f711f70003b893fb7ad62R239, moved to stay with other `parse` methods. I did this cause i was about to seperate them into different methods, then i realized that would be hard to review.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
Jim8y: this is the Value property from the dotnet.System.Nullable

```C#
        public readonly T Value
        {
            get
            {
                if (!hasValue)
                {
                    ThrowHelper.ThrowInvalidOperationException_InvalidOperation_NoValue();
                }
                return value;
            }
        }
```

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
```diff
@@ -594,6 +727,25 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
...
+                    AddInstruction(OpCode.ISNULL);
+                    Jump(OpCode.JMPIF_L, nullTarget);
+                    Jump(OpCode.JMPIF_L, trueTarget);
+                    Push("False");
+                    Jump(OpCode.JMP_L, endTarget);
+                    trueTarget.Instruction = Push("True");
+                    Jump(OpCode.JMP_L, endTarget);
+                    nullTarget.Instruction = AddInstruction(OpCode.NOP);
+                    AddInstruction(OpCode.DROP);
+                    Push("");
```
shargon, line 745: Why this push?

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
```diff
@@ -604,10 +756,32 @@ private bool TryProcessSystemMethods(SemanticModel model, IMethodSymbol symbol,
...
+                    if (instanceExpression is not null)
+                        ConvertExpression(model, instanceExpression);
+                    AddInstruction(OpCode.DUP);
+                    AddInstruction(OpCode.ISNULL);
+                    Jump(OpCode.JMPIF_L, nullTarget);
+                    ChangeType(StackItemType.ByteString);
+                    Jump(OpCode.JMP_L, endTarget);
+                    nullTarget.Instruction = AddInstruction(OpCode.NOP);
+                    AddInstruction(OpCode.DROP);
+                    Push("");
```
shargon, line 780: Why this push?

Jim8y: nullable type tostring returns `""` if null

cschuchardt88: why not `PUSHNULL` opcode?

Jim8y: benefits of pushnull?

cschuchardt88: No `GarbageCollection` or `Allocated Memory`. `bool?.ToString()` is nullable to begin with and saves additional checks for value. Also comes the question did user supply an empty string if its nullable and user sets `bool? b = null`, then they get an empty string.

shargon (approved):

shargon:
@Jim8y ut fault

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.cs
Jim8y: the problem is how do you know if its null or not? you have to do the check in neo vm, with opcode. GarbageCollection and Allocated Memory more like C# vm thing.

Jim8y: > No `GarbageCollection` or `Allocated Memory`. `bool?.ToString()` is nullable to begin with and saves additional checks for value. Also comes the question did user supply an empty string if its nullable and user sets `bool? b = null`, then they get an empty string.

I will merge this one since i have another pr that depends on this pr, maybe you can update what you are saying in another pr.

Jim8y: Oh, BTW, if the nullable value is null already, there will be a null already pushed to the stack by the compiler, here the logic is check if its a null or not.

PR #1082, opened 2024-07-25
Remove debuginfo param in optimizer

Hecate2:
Remove unnecessary debuginfo param in optimizer
Add startAddr for BasicBlock
Rename all `starting` in optimizer to `start`
Add tests for HasCallA

Without relying on debuginfo we may analyze all deployed contracts on mainnet in the future.


Jim8y (approved):

shargon (commented):
LGTM

PR #1084, opened 2024-07-26
[Neo DevPack Compiler] fix type issue and test issue

Jim8y:
We can not properly compile nullable types, they are compiled to `List` (Array) types and is not correct, thus having this pr that compiles nullable value to `Any`.

And also enables the test engine to update multiple artifacts instead of fail on every artifact update.


src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -63,6 +63,7 @@ public static bool IsInternalCoreMethod(this IMethodSymbol method)
 
         public static ContractParameterType GetContractParameterType(this ITypeSymbol type)
         {
+            if (type.ToString()!.EndsWith('?')) return ContractParameterType.Any;
```
shargon, line 66: All the types inside VM are nullable, we can lost information here

Jim8y: removed this change, will demonstrate in another pr. So this pr will only be allowing the test engine to update multiple artifacts and run them in parallel.

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
```diff
@@ -58,29 +63,53 @@ internal static CompilationContext[] EnsureArtifactsUpToDateInternal()
...
+                            }
+                        }
+                        else
+                        {
+                            updatedArtifactNames.TryAdd(res!);
+                        }
+                    });
+                }).ToArray();
+
+            Task.WhenAll(tasks).Wait();
```
cschuchardt88, line 99:
```suggestion
            Task.WhenAll(tasks).GetAwaiter().GetResult();
```

cschuchardt88 (dismissed):

cschuchardt88 (approved):

Jim8y:
@Hecate2 now you can run the tests twice and update all artifacts.

shargon (commented):
LgTM

PR #1085, opened 2024-07-27
demonstrate error type

Jim8y:
@shargon  this pr demonstrate the problem of compiling nullable type.

methods

```C#
    public static void TestBigInteger(BigInteger? a)
    public static void TestInt(int? a)
    public static void TestUInt(uint? a)
    public static void TestLong(long? a)
    public static void TestULong(ulong? a)
    public static void TestShort(short? a)
    public static void TestUShort(ushort? a)
    public static void TestSByte(sbyte? a)
    public static void TestByte(byte? a)
    public static void TestChar(char? a)
    public static void TestBool(bool? a)
    public static void TestUInt160(UInt160? a)
    public static void TestUInt256(UInt256? a)
    public static void TestUInt160Array(UInt160[] a)
    public static void TestUInt256Array(UInt256[] a)
    public static void TestByteArray(byte[] a)
    public static void TestString(string a)
    public static void TestObject(object a)
```

compiled:

```C#
    public abstract void TestBigInteger(IList<object>? a);
    public abstract void TestBool(IList<object>? a);
    public abstract void TestByte(IList<object>? a);
    public abstract void TestByteArray(byte[]? a);
    public abstract void TestChar(IList<object>? a);
    public abstract void TestInt(IList<object>? a);
    public abstract void TestLong(IList<object>? a);
    public abstract void TestObject(object? a = null);
    public abstract void TestSByte(IList<object>? a);
    public abstract void TestShort(IList<object>? a);
    public abstract void TestString(string? a);
    public abstract void TestUInt(IList<object>? a);
    public abstract void TestUInt160(object? a = null);
    public abstract void TestUInt160Array(IList<object>? a);
    public abstract void TestUInt256(object? a = null);
    public abstract void TestUInt256Array(IList<object>? a);
    public abstract void TestULong(IList<object>? a);
    public abstract void TestUShort(IList<object>? a);
```

PR #1086, opened 2024-07-27
Fix nullable type

Jim8y:
this pr fixes https://github.com/neo-project/neo-devpack-dotnet/pull/1085


Jim8y:
```C#
    public abstract void TestBigInteger(object? a = null);
    public abstract void TestBool(object? a = null);
    public abstract void TestByte(object? a = null);
    public abstract void TestByteArray(byte[]? a);
    public abstract void TestChar(object? a = null);
    public abstract void TestInt(object? a = null);
    public abstract void TestLong(object? a = null);
    public abstract void TestNullableString(string? a);
    public abstract void TestObject(object? a = null);
    public abstract void TestSByte(object? a = null);
    public abstract void TestShort(object? a = null);
    public abstract void TestString(string? a);
    public abstract void TestUInt(object? a = null);
    public abstract void TestUInt160(UInt160? a);
    public abstract void TestUInt160Array(IList<object>? a);
    public abstract void TestUInt256(UInt256? a);
    public abstract void TestUInt256Array(IList<object>? a);
    public abstract void TestULong(object? a = null);
    public abstract void TestUShort(object? a = null);
```

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Pattern.cs
```diff
@@ -67,7 +67,7 @@ public abstract class Contract_Pattern : Neo.SmartContract.Testing.SmartContract
     /// Unsafe method
     /// </summary>
     [DisplayName("testNotPattern")]
-    public abstract bool? TestNotPattern(IList<object>? x);
+    public abstract bool? TestNotPattern(object? x = null);
```
shargon, line 70: Array changed?

Jim8y: It is a `bool?`, supposed to be `object?`, it has being wrong since the begining.

PR #1087, opened 2024-07-27
[Compiler Fix] complete char type support

Jim8y:
This pr completes the support for `char` type. 

Char in C# has 2 bytes, and it has an range of [0, 65535].

It contains non-utf-8 characters, thus tostring may fail.


shargon (commented):
ut fault

Jim8y:
> ut fault

fixed

shargon (approved):

shargon (approved):

PR #1088, opened 2024-07-29
improve CheckStandards CompilationException error message by removing unnecessary exceptions and catches

Hecate2:
It seems that the additional catch just makes the error message on the console less clear.
Besides, could we keep compiling even if the contract does not match NEP standards?

### DIFF
--- a/src/Neo.Compiler.CSharp/ContractManifestExtension.cs
+++ b/src/Neo.Compiler.CSharp/ContractManifestExtension.cs
@@ -22,7 +22,8 @@ namespace Neo.Compiler
 {
     internal static class ContractManifestExtensions
     {
-        private static void CheckNep11Compliant(this ContractManifest manifest)
+        private static System.Collections.Generic.List<CompilationException>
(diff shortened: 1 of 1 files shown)


Hecate2:
Example output:
```
  Determining projects to restore...
  All projects are up-to-date for restore.
  Determining projects to restore...
  All projects are up-to-date for restore.
error NC3006: Incomplete or unsafe NEP standard NEP-11 implementation: symbol
error NC3006: Incomplete or unsafe NEP standard NEP-11 implementation: decimals
Examples:
        public override string Symbol
        {
            [Safe]  // Do not drop `[Safe]`!
            get => "GAS";
        }
        public override byte Decimals
        {
            [Safe]  // Do not drop `[Safe]`!
            get => 8;
        }
Do not write `[Safe]` for `Transfer` method! `[Safe]` forbids writing to storage and emitting events.

Created ..../..../..../Contract.nef
Created .manifest.json
Created ...
Compilation completed successfully.
```

Jim8y (approved):

shargon (approved):

PR #1089, opened 2024-07-29
[DevPack TestEngine] debug single contract

Jim8y:
This pr enables us to convinently debug the compilation of a specified contact.


src/Neo.Compiler.CSharp/CompilationEngine.cs
```diff
@@ -128,6 +129,21 @@ public List<CompilationContext> CompileProject(string csproj)
...
+        /// Debug the compilation of a specific contract.
+        /// If this method is called, the compiler will
+        /// only compile the specified contract.
+        /// Effect only under debug mode.
+        /// </summary>
+        /// <param name="contractName">The name of the contract to be debugged.</param>
+        /// <returns>The current <see cref="CompilationEngine"/>.</returns>
+        internal CompilationEngine SetDebugContract(string contractName)
+        {
+            if (Options.Debug)
```
shargon, line 142: it's a different thing, maybe you want to compile in release only one contract

Jim8y: No, i want to debug the complication process of a single contract, of course it should be under Debug mode.

shargon: But you want to compile only one smart contract, I don't understand well the debug concept here

shargon: Why not use CompileSource, instead of folder?

Jim8y: this is just for debugging the unit test contracts, otherwise i have to do the same thing on every pr manually. Not supposed to be used by users.

shargon: So you want to test only one UT, without calling the `AssemblyCleanup`?

Jim8y: > So you want to test only one UT, without calling the `AssemblyCleanup`?

Yes, but not testing one UT, but debug the complication process of a single contract. 

Currently if we run the test, all contracts will be compiled together, making it impossible to debug one specific contract.

Jim8y: > So you want to test only one UT, without calling the `AssemblyCleanup`?

But it would be great if you can find a better way to achieve it, mine is stuipd.

shargon: You create one UT for with this value (contract), and you remove this ut later?

shargon: Please take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/1109

PR #1091, opened 2024-07-29
[TestEngine] Add `ResetFeeConsumed`

shargon:
Related to https://github.com/neo-project/neo-devpack-dotnet/pull/1076

### DIFF
--- a/src/Neo.SmartContract.Testing/TestEngine.cs
+++ b/src/Neo.SmartContract.Testing/TestEngine.cs
@@ -183,6 +183,11 @@ namespace Neo.SmartContract.Testing
         public FeeWatcher FeeConsumed { get; }
 
         /// <summary>
+        /// Reset FeeConsumed on each Execution
+        /// </summary>
+        public bool ResetFeeConsumed { get; set; }
+
+        /// <summary>
         /// Sender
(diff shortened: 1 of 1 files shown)


src/Neo.SmartContract.Testing/TestEngine.cs
```diff
@@ -182,6 +182,11 @@ public long Fee
         /// </summary>
         public FeeWatcher FeeConsumed { get; }
 
+        /// <summary>
+        /// Reset FeeConsumed on each Execution
+        /// </summary>
+        public bool ResetFeeConsumed { get; set; }
```
vncoelho, line 188: where will you set it?

shargon: https://github.com/neo-project/neo-devpack-dotnet/pull/1076 Jimmy want to use it, it can be true by default

shargon, line 188:
```suggestion
        public bool ResetFeeConsumed { get; set; } = true;
```

vncoelho (approved):

Jim8y (approved):

PR #1092, opened 2024-07-30
RemoveUnnecessaryJumps optimizer

Hecate2:
Remove JMP and JMP_L that just jumps to the next instruction of itself.
If the removed JMP is the target of another jumping instruction, re-target to the next instruction after the JMP.


Hecate2:
Yet I do not know how to generate the test contract artifacts. Tests will fail.

Jim8y:
> Yet I do not know how to generate the test contract artifacts. Tests will fail.

Manually construct scripts then.

Hecate2:
Do not know why tests got `Coverage not found`

shargon:
> Do not know why tests got `Coverage not found`

I will deal with this today

Jim8y:
why gas consumed is not affected?

tests/Neo.SmartContract.Framework.UnitTests/Services/RuntimeTest.cs
```diff
@@ -79,7 +79,7 @@ public void Test_Random()
             Engine.SetTransactionSigners(Alice);
             Engine.Transaction.Nonce = 0x01020304;
             Engine.PersistingBlock.Nonce = 0x01020304;
-            Assert.AreEqual(BigInteger.Parse("205214629647252551804410215848131909835"), Contract.GetRandom());
+            Assert.AreEqual(BigInteger.Parse("170860635854782763966488234028122619308"), Contract.GetRandom());
```
Jim8y, line 82: why this one is changed?

Hecate2: > why this one is changed?

The contract content affects the random number

Jim8y (dismissed):

shargon:
> why gas consumed is not affected?

This is a good question

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
```diff
@@ -21,7 +21,7 @@ public class TestCleanup : TestCleanupBase
         private static readonly object RootSync = new();
 
         [AssemblyCleanup]
-        public static void EnsureCoverage() => EnsureCoverageInternal(Assembly.GetExecutingAssembly(), 0.76M);
+        public static void EnsureCoverage() => EnsureCoverageInternal(Assembly.GetExecutingAssembly(), 0.755M);
```
shargon, line 24: Less coverage?

cschuchardt88:
You shouldn't add this because if i want to compile (_see below_). I should be able to

```csharp
//do something
for (;;) {}
//do something
```

Hecate2:
> ```cs
> do_work:
> for (;;) { goto do_work; }
> for (;;) {}
> ```

I have re-targeted other jumping instructions that points to the deleted JMP. Infinite loops and goto's should still work.
```csharp
        public void InfiniteLoop()          // line 18
        {
        do_work:
            for (; ; ) { goto do_work; }    // line 21
            for (; ; ) { }
        }                                   // line 23
```
with All optimization
```python
# Method Start txHash.txHash.InfiniteLoop()
000 INITSLOT 00-01 # 0 local variables, 1 arguments
003 JMP 04 # pos: 7 (offset: 4)
# Code txHash.cs line 21: "goto do_work;"
# Code txHash.cs line 23: "}"
005 JMP FE # pos: 3 (offset: -2)
# Method End txHash.txHash.InfiniteLoop()
```
with only Basic optimization:
```python
# Method Start txHash.txHash.InfiniteLoop()
000 INITSLOT 00-01 # 0 local variables, 1 arguments
003 JMP 04 # pos: 7 (offset: 4)
# Code txHash.cs line 21: "goto do_work;"
005 JMP FE # pos: 3 (offset: -2)
007 JMP FE # pos: 5 (offset: -2)
009 JMP 02 # pos: 11 (offset: 2)  # deleted, because never reached
011 JMP 00 # pos: 11 (offset: 0)  # deleted, because never reached
# Method End txHash.txHash.InfiniteLoop()
```

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
Hecate2: > Less coverage?

I am really having trouble to reach the coverage in HasCallATests

Hecate2:
> why gas consumed is not affected?

It does affect gas consumption. I will have to change all the hundreds of gas consumption asserts in https://github.com/neo-project/neo-devpack-dotnet/pull/1076

Hecate2:
> Yet I do not know how to generate the test contract artifacts. Tests will fail.

In CreateArtifact<T>, comment Assert.Fail($"{typeof(T).Name} artifact was wrong");, and run test multiple times

Hecate2:
Jim8y has found a bug in optimizer, and I am fixing it here.
When no exception is thrown from try catch finally, the codes after finally should be executed. But my `CoverInstruction` forces to throw exception from try, in order to visit catch. This makes finally covered without covering codes after finally. When a later branch tries to visit finally without exception, the branch sees finally as visited, and quit, leaving the codes after finally unvisited.

Jim8y (approved):

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
Jim8y: > Less coverage?

No need to worry, i have added tons of unit tests in another pr, can increase later.

Jim8y:
merge as there is a bug fix, and need it to move on adding unit tests.

shargon:
> why gas consumed is not affected?

it was merged without answer this.....

shargon (commented):
Big PR merged to soon, in my opinion should be reviewed deeper...

shargon:
it doesn't compile master with this change....

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
shargon: For me is something to worry, because if this PR reduce the opcodes, how it's possible that the coverage of the smart contract will be less than before?

PR #1096, opened 2024-07-31
Unit tests try catch

Jim8y:
This pr increase coverage of try-catch.

PR #1098, opened 2024-07-31
fix deleting codes after finally when `catch` may throw

Hecate2:
Sorry for still having bugs...
https://github.com/neo-project/neo-devpack-dotnet/pull/1096
https://github.com/Jim8y/neo-devpack-dotnet/blob/95a718cc2d10c3cc66b5b7a6c7587a82c0c91622/tests/Neo.Compiler.CSharp.TestContracts/Contract_TryCatch.cs#L101-L120
This method demonstrates the error of the optimizer in master. The optimizer deletes all the codes after finally because it visits the throw in catch, and then visits finally (without visiting codes after finally). Then when we do not throw in catch, the optimizer thinks finally has been visited well, and never visits the codes after finally.


shargon:
UT Fault

shargon (commented):
LgTM

PR #1102, opened 2024-08-01
delete codes after finally when finally throws or aborts

Hecate2:


Hecate2:
There is still bug. `throwed` should be associated with FINALLY stack, in order to judge if the code had throwed in try/catch corresponding to the finally.

shargon:
> There is still bug. `throwed` should be associated with FINALLY stack, in order to judge if the code had throwed in try/catch corresponding to the finally. Otherwise we cannot judge the meaning of `throwed` in finally.

Could you open a issue for it?

Hecate2:
> > There is still bug. `throwed` should be associated with FINALLY stack, in order to judge if the code had throwed in try/catch corresponding to the finally. Otherwise we cannot judge the meaning of `throwed` in finally.
> 
> Could you open a issue for it?

Already fixed in the latest commit c0a0eefc4ba929384c87385d76659a4ccfd7b8c5

Hecate2:
ABORT in finally is not handled well

Jim8y (approved):

shargon (commented):
LGTM

PR #1103, opened 2024-08-01
Use Original Type to Replace Nullable Type

Jim8y:
This pr uses the original type of both value-type and non-value type to replce its nullable type.


tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
```diff
@@ -20,8 +20,8 @@ public class TestCleanup : TestCleanupBase
         private static CompilationContext[]? compilationContexts;
         private static readonly object RootSync = new();
 
-        [AssemblyCleanup]
-        public static void EnsureCoverage() => EnsureCoverageInternal(Assembly.GetExecutingAssembly(), 0.77M);
+        // [AssemblyCleanup]
```
shargon, line 23: Revert this change?

shargon (approved):

PR #1108, opened 2024-08-01
[FIX] fix string empty

Jim8y:
This pr fix the string.Empty. Previously it returns "NULL" cause the compiler has filed to process this syntax, which actually should return "".


src/Neo.Compiler.CSharp/MethodConvert/Expression/MemberExpression.cs
```diff
@@ -55,6 +55,16 @@ private void ConvertMemberAccessExpression(SemanticModel model, MemberAccessExpr
                 }
                 else if (field.IsStatic)
                 {
+                    // Have to process the string.Empty specially since it has no AssociatedSymbol
+                    // thus will return directly without this if check.
+                    if (field.ContainingType.ToString() == "string" && field.Name == "Empty")
+                    {
+                        // preInitialize?.Invoke();
```
Jim8y, line 62:
```suggestion
```

src/Neo.Compiler.CSharp/MethodConvert/Expression/MemberExpression.cs
```diff
@@ -55,6 +55,15 @@ private void ConvertMemberAccessExpression(SemanticModel model, MemberAccessExpr
                 }
                 else if (field.IsStatic)
                 {
+                    // Have to process the string.Empty specially since it has no AssociatedSymbol
+                    // thus will return directly without this if check.
+                    if (field.ContainingType.ToString() == "string" && field.Name == "Empty")
+                    {
+                        Push(string.Empty);
+                        // postInitialize?.Invoke();
```
Jim8y, line 63:
```suggestion
```

src/Neo.Compiler.CSharp/MethodConvert/Expression/MemberExpression.cs
```diff
@@ -55,6 +55,14 @@ private void ConvertMemberAccessExpression(SemanticModel model, MemberAccessExpr
                 }
                 else if (field.IsStatic)
                 {
+                    // Have to process the string.Empty specially since it has no AssociatedSymbol
+                    // thus will return directly without this if check.
+                    if (field.ContainingType.ToString() == "string" && field.Name == "Empty")
```
shargon, line 60: lowercase comparation and without the namespace?

Jim8y: exactly, this is a built in static field of string,, will never change.

shargon (approved):

PR #1110, opened 2024-08-02
[FIX] fix call virtaual

Jim8y:
It is possible to compile unsupported syntax and cause error script. thus having this pr to fix it.

error NC2001: Unsupported syntax: object.ToString().


src/Neo.Compiler.CSharp/MethodConvert/CallHelpers.cs
```diff
@@ -224,6 +224,8 @@ private void CallVirtual(IMethodSymbol symbol)
         IFieldSymbol[] fields = members.OfType<IFieldSymbol>().ToArray();
         IMethodSymbol[] virtualMethods = members.OfType<IMethodSymbol>().Where(p => p.IsVirtualMethod()).ToArray();
         int index = Array.IndexOf(virtualMethods, symbol);
+        if (index < 0)
+            throw new CompilationException(symbol, DiagnosticId.SyntaxNotSupported, $"Unsupported syntax: {symbol.OriginalDefinition}.");
```
Hecate2, line 228: Could we just say `Cannot call virtual method {...}`?

Jim8y: To be honest, i dont really know what this syntax means. though its virtualmethods, but i have no idea why and how these code works.

shargon: where did you get this error?

Jim8y: object.tostring

shargon (approved):

PR #1109, opened 2024-08-01
[TestTengine] Test only one contract

shargon:
Alternative to https://github.com/neo-project/neo-devpack-dotnet/pull/1089

This pr avoid `AssemblyCleanup` when we run only one unit test

PR #1111, opened 2024-08-02
ensure correct order

Jim8y:
this pr ensures that the coverage is calculated after the artifacts is updated.


Jim8y:
The github action often fails because of coverage null. thus having this pr that ensures coverage is calculated after the compilation.

Jim8y:
action was executed multiple times, it works.

shargon (commented):
like this is always checked that the artifacts are the same, previously is like a unit test, you can skip that part

shargon:
I understand the problem, if you choose one UT, is not selected the cleanUp

shargon:
@Jim8y you prefer to work on this one or in #1109, because it's almost the same

Jim8y:
> @Jim8y you prefer to work on this one or in #1109, because it's almost the same, (I need this code to fix the compilation of unique unit test)

just finish yours, mine can wait here to see if your work fulfill my need. But this pr mainly focus on solve the github action fail issue, not sure if your pr can do the same.

Jim8y:
With https://github.com/neo-project/neo-devpack-dotnet/pull/1109 this pr is no longer needed.

PR #1113, opened 2024-08-02
[Update] default constructor

Jim8y:
This pr makes it easier to create unit test class.


src/Neo.SmartContract.Testing/TestingStandards/TestBase.cs
```diff
@@ -26,7 +28,24 @@ public class TestBase<T> where T : SmartContract
...
+
+        var manifestField = typeof(T).GetField("Manifest", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static);
+        if (manifestField == null)
+        {
+            throw new InvalidOperationException($"The type {typeof(T).Name} does not have a static Manifest field.");
+        }
+        var manifest = (ContractManifest)manifestField.GetValue(null)!;
+
+        TestBaseSetup(nef, manifest);
+    }
```
Jim8y, line 48: Core logic is here.

Jim8y:
@shargon you can see the github action often fails, and need to run multiple times to make it work, pr https://github.com/neo-project/neo-devpack-dotnet/pull/1111 is for this purpose. 

  Failed Test_HasCallA [1 ms]
  Error Message:
   Assembly Cleanup method TestCleanup.EnsureCoverage failed. Error Message: System.ArgumentNullException: Value cannot be null.. StackTrace:     at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at Neo.SmartContract.Testing.TestingStandards.TestCleanupBase.EnsureCoverageInternal(Assembly assembly, Decimal requiredCoverage) in /home/runner/work/neo-devpack-dotnet/neo-devpack-dotnet/src/Neo.SmartContract.Testing/TestingStandards/TestCleanupBase.cs:line 25
   at Neo.Compiler.CSharp.UnitTests.TestCleanup.EnsureCoverage() in /home/runner/work/neo-devpack-dotnet/neo-devpack-dotnet/tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs:line 24
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

  Stack Trace:
     at Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.TestAssemblyInfo.ExecuteAssemblyCleanup() in /_/src/Adapter/MSTest.TestAdapter/Execution/TestAssemblyInfo.cs:line 315
   at Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.Execution.UnitTestRunner.RunRequiredCleanups(ITestContext testContext, TestMethodInfo testMethodInfo, TestMethod testMethod, UnitTestResult[] results) in /_/src/Adapter/MSTest.TestAdapter/Execution/UnitTestRunner.cs:line 227

src/Neo.SmartContract.Testing/TestingStandards/TestBase.cs
shargon: We can cache the fields in a static var?

src/Neo.SmartContract.Testing/SmartContract.cs
```diff
@@ -9,6 +9,7 @@
 using System.Numerics;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using Neo.SmartContract.Manifest;
```
shargon, line 12: Required?
```suggestion
```

src/Neo.SmartContract.Testing/TestingStandards/TestBase.cs
shargon: Or use an interface ? https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/static-abstracts-in-interfaces

Jim8y: Tried, wont work, T will get smartcontract fields and meethods, instead of the contacts. 

Maybe you can try it in another pr as an update of this one.

Jim8y: Now i have a new idea, will try.

Jim8y:
Weird, my local test has no problem, and git has no log of artifact update.....

src/Neo.SmartContract.Testing/Neo.SmartContract.Testing.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.1;net8.0</TargetFrameworks>
+        <TargetFramework>net8.0</TargetFramework>
```
shargon, line 4: Not possible with net standard, right?

Jim8y: to use static in interface, no we cant.

cschuchardt88: Why do you need static in `interface` just use a `abstract class`. That's why they have it. Shouldn't unlimit the program.

Jim8y: > abstract class

Need to get the static field or property difrectly from the passed in contract class. With abstract class, cannot achieve that. 

>You can't access a static method through a generic type parameter even if it's constrained to a type. 

https://stackoverflow.com/questions/9146533/how-to-access-static-methods-of-generic-types

Jim8y: Core reason is here, this line:

https://github.com/neo-project/neo-devpack-dotnet/pull/1113/files#diff-c3a87e7e4709d175a18f40e4852b3997b25d14abe5badff88de213916e833f91R34

cschuchardt88: Just use `interface` without `static` doesn't have to be `static` methods for `SmartContract`

Just look at _below_. Its `non-static`
https://github.com/neo-project/neo-devpack-dotnet/blob/0ec0b8e1a3d632b046887a903fc9dc631e0613cd/src/Neo.SmartContract.Framework/Nep11Token.cs#L51

Jim8y: I need it to be static to get them from the contract artifact class. 

```C#
public abstract class Contract_ABIAttributes3 : Neo.SmartContract.Testing.SmartContract, IContractInfo
{
    #region Compiled data

    public static Neo.SmartContract.Manifest.ContractManifest Manifest => Neo.SmartContract.Manifest.ContractManifest.Parse(@"{""name"":""Contract_ABIAttributes3"",""groups"":[],""features"":{},""supportedstandards"":[],""abi"":{""methods"":[{""name"":""test"",""parameters"":[],""returntype"":""Integer"",""offset"":0,""safe"":false}],""events"":[]},""permissions"":[{""contract"":""0x01ff00ff00ff00ff00ff00ff00ff00ff00ff00a4"",""methods"":""*""}],""trusts"":[],""extra"":{""nef"":{""optimization"":""All""}}}");

    /// <summary>
    /// Optimization: "All"
    /// </summary>
    public static Neo.SmartContract.NefFile Nef => Neo.IO.Helper.AsSerializable<Neo.SmartContract.NefFile>(Convert.FromBase64String(@"TkVGM1Rlc3RpbmdFbmdpbmUAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIQQEYexxg="));

```


I need to read Nef and Manifest from Artifacts of contracts.

cschuchardt88: It's only for testing framework, so I guess it is ok.

shargon: Maybe can have a property with the interface, that call the static field

Jim8y: We need to get the static field via the class, cant be a property.

shargon: But we can use both. Something like: this.GetContract(out nef, out manifest){ ... }

Jim8y: but currently we dont need to support netstandard2.1 at all,,,,

src/Neo.SmartContract.Testing/TestingStandards/TestBase.cs
```diff
@@ -26,7 +28,11 @@ public class TestBase<T> where T : SmartContract
     /// <summary>
     /// Empty constructor
     /// </summary>
-    public TestBase() { }
+    public TestBase()
+    {
+
+        TestBaseSetup(T.Nef, T.Manifest);
```
shargon, line 34:
```suggestion
        TestBaseSetup(T.Nef, T.Manifest);
```

shargon (approved):

Issue #1080, opened 2024-07-23
Fix not covered contracts


shargon:
After migrate the whole tests to the new system, some contracts have been detected without any kind of test (or not covered).

- [x] Coverage not found for Neo.SmartContract.Testing.Contract_ABIAttributes2
- [x] Coverage not found for Neo.SmartContract.Testing.Contract_ABIAttributes3
- [x] Coverage not found for Neo.SmartContract.Testing.Contract_Event
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_MultipleA
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_MultipleB
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_OnDeployment1
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_OnDeployment2
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_PostfixUnary
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_NEP17
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_Throw
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_String
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_IndexOrRange
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_Logical
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_MemberAccess
- [x]  Coverage not found for Neo.SmartContract.Testing.Contract_NEP11

Jim8y:
@shargon need this, please start working on them,

PR #1116, opened 2024-08-05
[TestEngine] Add coverage

shargon:
Add coverage according to https://github.com/neo-project/neo-devpack-dotnet/issues/1080

### DIFF
--- a/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Logical.cs
+++ b/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Logical.cs
@@ -1,0 +1,85 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.SmartContract.Testing;
+using Neo.SmartContract.Testing.TestingStandards;
+
+namespace Neo.Compiler.CSharp.UnitTests
+{
+    [TestClass]
+    public class UnitTest_Logical : TestBase<Contract_Logical>
+    {
+        public UnitTest_Logical() : base(Contract_Logical.Nef, Contract_Logical.Manifest) { }
+
+        [TestMethod]
+        public void Test_TestConditionalLogicalAnd()
+        {
+            foreach (var x in new bool[] { true, false })
+                foreach (var y in new bool[] { true, false })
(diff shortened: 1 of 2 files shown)


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Logical.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        [TestMethod]
+        public void Test_TestConditionalLogicalAnd()
+        {
+            foreach (var x in new bool[] { true, false })
+                foreach (var y in new bool[] { true, false })
+                {
+                    var result = Contract.TestConditionalLogicalAnd(x, y);
+                    Assert.AreEqual(x && y, result);
+                    // TODO: 1047210
```
shargon, line 20: @Jim8y why this test change the cost according to the values? could you check it please?

Jim8y: Cause true or false have different jmp pass. So you can not really do this gas check in a for loop.

Jim8y: Must do it this way:

```C#
        [TestMethod]
        public void Test_TestConditionalLogicalOr()
        {
            var result = Contract.TestConditionalLogicalOr(true, true);
            Assert.AreEqual(true || true, result);
            Assert.AreEqual(1047210, Engine.FeeConsumed.Value);

            result = Contract.TestConditionalLogicalOr(true, false);
            Assert.AreEqual(true || false, result);
            Assert.AreEqual(1047210, Engine.FeeConsumed.Value);

            result = Contract.TestConditionalLogicalOr(false, true);
            Assert.AreEqual(false || true, result);
            Assert.AreEqual(1047180, Engine.FeeConsumed.Value);

            result = Contract.TestConditionalLogicalOr(false, false);
            Assert.AreEqual(false || false, result);
            Assert.AreEqual(1047180, Engine.FeeConsumed.Value);
        }
```

Jim8y (approved):

PR #1129, opened 2024-08-08
[Compiler Tools] add some tools to wrtite instructions.

Jim8y:
This pr add some tool methods that concludes the instruction sets. 

These tool methods can not only help us to write instructions, but also help us to optimize the price (with cheaper way of achieving) in the future.

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
@@ -27,7 +27,6 @@ using System.Linq;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using Neo.Optimizer;
 
 namespace Neo.Compiler
 {
@@ -668,11 +667,8 @@ namespace Neo.Compiler
 
         private Instruction Jump(OpCode opcode, JumpTarget target)
         {
-            if (!OpCodeTypes.conditionalJump.Contains(opcode) && !OpCodeTypes.unconditionalJump.Contains(opcode) &&
-                !OpCodeTypes.conditionalJump_L.Contains(opcode))
-            {
-                throw new Exception("Invalid opcode for Jump");
-            }
+            if (!Neo.Optimizer.JumpTarget.SingleJumpInOperand(opcode))
+                throw new Exception($"Invalid opcode {opcode} for {nameof(Jump)}");
 
             return AddInstruction(new Instruction
             {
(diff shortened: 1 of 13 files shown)


src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -667,13 +667,99 @@ private void CreateObject(SemanticModel model, ITypeSymbol type, InitializerExpr
...
 
+        /// <summary>
+        /// Duplicate an item in the stack without change anything else
+        ///  For example, dup(2)
+        ///     a - b - c => a - b - b - c.
+        /// </summary>
+        /// <param name="index">Index of items to be duped.</param>
+        private void DupAt(int index)
+        {
+            if (index <= 0) throw new Exception("Invalid index for Dup");
```
shargon, line 688: Switch?

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -667,13 +667,99 @@ private void CreateObject(SemanticModel model, ITypeSymbol type, InitializerExpr
...
+
+        /// <summary>
+        /// Drop items at index n from the stack.
+        /// </summary>
+        /// <param name="index">Index of the items to be dropped.</param>
+        /// <exception cref="Exception">When index is invalid.</exception>
+        private void DropAt(int index)
+        {
+            if (index <= 0) throw new Exception("Invalid index for DropAt");
+            if (index == 1)
```
shargon, line 732: Why index doesn't start in 0?

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_String.cs
```diff
@@ -100,10 +100,10 @@ public void Test_TestEndWith()
         public void Test_TestContains()
         {
             Assert.IsTrue(Contract.TestContains("hello world"));
-            Assert.AreEqual(2032920, Engine.FeeConsumed.Value);
+            Assert.AreEqual(2032800, Engine.FeeConsumed.Value);
```
shargon, line 103: Why changed the gas in the methods are not used?

Hecate2: > Why changed the gas in the methods are not used?

I do not really know what happened, but we do not have `TestInsert` method in the master branch, and we have an artifact Contract_String that does not match the test contract source codes

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
Jim8y: these does not mater at all, can be this way, can be that way. Wont affect what this pr intents for.

PR #1136, opened 2024-08-13
add opcode methods to make it easier to construct instruction scripts.

Jim8y:


src/Neo.Compiler.CSharp/MethodConvert/StackHelpers.OpCodes.cs
```diff
@@ -0,0 +1,508 @@
...
+    private Instruction ReverseN(int? count)
+    {
+        if (count.HasValue)
+            Push(count.Value);
+        return AddInstruction(OpCode.REVERSEN);
+    }
+
+    #endregion
+
+
```
shargon, line 128:
```suggestion
```

src/Neo.Compiler.CSharp/MethodConvert/StackHelpers.OpCodes.cs
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    private Instruction PackStruct()
+    {
+        return AddInstruction(OpCode.PACKSTRUCT);
+    }
+
+    private Instruction Pack(int? size)
+    {
+
```
shargon, line 377:
```suggestion
```

src/Neo.Compiler.CSharp/MethodConvert/StackHelpers.OpCodes.cs
```diff
@@ -0,0 +1,507 @@
...
+        if (size.HasValue)
+            Push(size.Value);
+        return AddInstruction(OpCode.PACK);
+
```
shargon, line 381:
```suggestion
```

shargon:
Change the previous calls?

Jim8y:
> Change the previous calls?

Will work on new syscalls and see the experience,,,,,,

shargon:
> > Change the previous calls?
> 
> Will work on new syscalls and see the experience,,,,,,

I don't understand the addiction of new methods if they are not used

Jim8y:
> > > Change the previous calls?
> > 
> > Will work on new syscalls and see the experience,,,,,,
> 
> I don't understand the addiction of new methods if they are not used

its ok if you dont like it, not a necessary pr, it just same my effort to add new syacalls. no need to write addinatruction, but directly use the corresponding opcode methods.

shargon (approved):

Jim8y:
fix conflict

Jim8y:
please review again, there was a conflic and i have to update it. @shargon

shargon (dismissed):

Jim8y:
@shargon I am crying on it,,,, conflict again,,,, i will break prs into even smaller if you dont mind,,,,, too many conflicts now.

shargon (dismissed):

shargon:
It doesn't build..

shargon (approved):

PR #1140, opened 2024-08-19
[Compiler Fix] fix push integer

Jim8y:
This pr fixes a bug in the compiler, the `push(BigInteger)` are wrong at processing negative value:

it got 15777216 when push -1000000
and got 280474976710656 when push -1000000000000

res = ~value -1


src/Neo.Compiler.CSharp/MethodConvert/Helpers/StackHelpers.cs
```diff
@@ -70,25 +70,45 @@ private void Push(bool value)
...
+            {
+                OpCode = OpCode.PUSHINT64,
+                Operand = PadRight(buffer, bytesWritten, 8, number.Sign < 0).ToArray()
+            },
+            16 => new Instruction
+            {
+                OpCode = OpCode.PUSHINT128,
+                Operand = PadRight(buffer, bytesWritten, 16, number.Sign < 0).ToArray()
+            },
+            _ => new Instruction
```
Hecate2, line 104: throw new ArgumentOutOfRangeException when the number overflows 32 bytes?

Jim8y: done

tests/Neo.Compiler.CSharp.TestContracts/Contract_Stack.cs
```diff
@@ -0,0 +1,26 @@
...
+        return (byte.MinValue, byte.MaxValue, sbyte.MinValue, sbyte.MaxValue, short.MinValue, short.MaxValue, ushort.MaxValue, uint.MaxValue, int.MinValue, int.MaxValue, ulong.MaxValue, long.MinValue, long.MaxValue);
+    }
+
+    public static (int, long) Test_External()
+    {
+        const int intValue = -1000000;
+        const long longValue = -1000000000000;
+
+        return (intValue, longValue);
+
```
shargon, line 24:
```suggestion
```

tests/Neo.Compiler.CSharp.TestContracts/Contract_Stack.cs
```diff
@@ -0,0 +1,26 @@
+using System.Numerics;
+
+namespace Neo.Compiler.CSharp.TestContracts;
+
+public class Contract_Stack : SmartContract.Framework.SmartContract
+{
+
```
shargon, line 7:
```suggestion
```

src/Neo.Compiler.CSharp/MethodConvert/Helpers/StackHelpers.cs
```diff
@@ -70,25 +70,46 @@ private void Push(bool value)
 
     private Instruction Push(BigInteger number)
     {
-        if (number == BigInteger.MinusOne) return AddInstruction(OpCode.PUSHM1);
-        if (number >= BigInteger.Zero && number <= 16) return AddInstruction(OpCode.PUSH0 + (byte)number);
-        byte n = number.GetByteCount() switch
+        if (number >= -1 && number <= 16) return AddInstruction(OpCode.PUSH0 + (byte)(int)number);
```
shargon, line 73: Where is pushed -1?

Jim8y: number>=-1

shargon: Return? Is not pushM1?

Jim8y: emmmm, let me check, maybe should be sbyte instead.

shargon: Sorry, my bad, it was indented wrong in my phone, i readed just return

Jim8y: > Return? Is not pushM1?

i think you find a bug in the core, vm also convert it to byte, instead of sbyte. do you mind to fix it in the core/vm/scriptbuilder?

shargon: But yes, must be signed 😅

shargon: > > Return? Is not pushM1?
> 
> i think you find a bug in the core, vm also convert it to byte, instead of sbyte. do you mind to fix it in the core/vm/scriptbuilder?

In both, isn't it?

Jim8y: yes.

shargon (approved):

PR #1145, opened 2024-08-22
update workflow to enable  only run failed jobs

Jim8y:
This pr split the job into several sub jobs, enabling the action to only rerun the failed sub jobs. 

this problem is coverage is always fail and need to rerun multiple times, while the tests are actually correct, but now we have to rerun everyting since they are of the same job.


Jim8y:
This will make the action time unnecessarily long, and i can not fix it.

PR #1146, opened 2024-08-23
[Github workflow] update format

Jim8y:
only run dotnet format on changed file to save the dotnet format time


Jim8y:
Why having this pr? I am working to fix the run only failed jobs, turns out having multiple sub jobs will increase the workflow executing time, making it necessary to optimize the time as many as possible. This pr saves 50% of the format time.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_MultipleB.cs
```diff
@@ -12,5 +12,11 @@ public void Test()
             Assert.IsFalse(Contract.Test());
             Assert.AreEqual(984060, Engine.FeeConsumed.Value);
         }
+
+        [TestMethod]
+        public void Test2()
+        {
+            // Test if the workflow format is working
+        }
```
Jim8y, line 20:
```suggestion
```

shargon (commented):
Json is not checked?

Jim8y:
> Json is not checked?

do we have json in devpack?

shargon:
> > Json is not checked?
> 
> do we have json in devpack?

No, but this will be copied to neo?

Jim8y:
> > > Json is not checked?
> > 
> > do we have json in devpack?
> 
> No, but this will be copied to neo?

sure, i will update later, can you help to check the coverage failure in other prs? i was trying to fix that problem thus having this one.

shargon (approved):

shargon (approved):

PR #1147, opened 2024-08-23
[Workflow] Cancel existing workflow

Jim8y:
If a new commit is pushed to the pr, previous unfinished workflow execution will be canceled to save our free action time.


shargon (approved):
Nice!

PR #1148, opened 2024-08-23
[Devpack Compiler] add  nested out support

Jim8y:
This pr adds nested out parameter support. 

Since out parameter need to connect the symbols in/out method, we need to set a global field to store the value, however, it is possible that different methods set different fields, but actually those fields values are also connected by nested method calls, thus we also need to figure out a way to connect different fields. 

all of these challenges make it hard to fully support out keyword, and this pr completes it.


src/Neo.Compiler.CSharp/CompilationContext.cs
```diff
@@ -56,8 +56,9 @@ public class CompilationContext
         private readonly System.Collections.Generic.List<byte> _anonymousStaticFields = new();
         private readonly Dictionary<ITypeSymbol, byte> _vtables = new(SymbolEqualityComparer.Default);
         private readonly Dictionary<ISymbol, byte> _capturedStaticFields = new(SymbolEqualityComparer.Default);
-        // This dictionary is used to map out parameters/arguments to local variables (in the method where the out argument is declared)
-        internal Dictionary<IParameterSymbol, ILocalSymbol> OutParamToLocal { get; } = new(SymbolEqualityComparer.Default);
+        // This dictionary is used to sync value from key symbol to value symbol
+        // We need to sync the value symbol to the key symbol when the key symbol is updated
+        internal Dictionary<ISymbol, System.Collections.Generic.List<ISymbol>> OutStaticFieldsSync { get; } = new Dictionary<ISymbol, System.Collections.Generic.List<ISymbol>>(SymbolEqualityComparer.Default);
```
Jim8y, line 61: this helps us to connect different filed, key are the symbol of called method parameter, value contains the symbols that is connected to the parameter and need to sync its value with parameter.

src/Neo.Compiler.CSharp/MethodConvert/Helpers/CallHelpers.cs
```diff
@@ -196,7 +213,37 @@ private void ProcessOutIdentifier(SemanticModel model, IParameterSymbol paramete
...
+        }
+        else if (!parameterCaptured && symbolCaptured)
+        {
+            _context.AssociateCapturedStaticField(parameter, symbolIndex);
+        }
+        else if (parameterCaptured && symbolCaptured && parameterIndex != symbolIndex)
+        {
+            // both values are already captured in different indirectly connected methods,
+            // but they are different, thus need to sync value from symbol to parameter
+            if (!_context.OutStaticFieldsSync.TryGetValue(parameter, out var syncList))
```
Jim8y, line 235: in/out parameters are registerd to different field, which means we need to connect them.

shargon (dismissed):

shargon (approved):

PR #1150, opened 2024-08-23
update compiler project struct

Jim8y:
This pr change the structure of the compiler to group files by its type, which  makes the compiler structure more clear and help us to  locate what we need.


.github/workflows/main.yml
```diff
@@ -30,38 +30,9 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: ${{ env.DOTNET_VERSION }}
-    - name: Get changed files
-      id: changed-files
+    - name: Check format
       run: |
-        if [ "${{ github.event_name }}" == "pull_request" ]; then
```
Jim8y, line 36: Sorry i made this stupid mistake, it does not work, but will actually take longer to run. Short only for pr that does not touch C#.

shargon (approved):

PR #1153, opened 2024-08-24
[Devpack Testing] update artifacts

Jim8y:
use primary constructor to make it easier to create artifacts.

I have to create one placeholder artifact file for every new test contract, this pr makes it easier for me to create that placeholder artifacts.


src/Neo.SmartContract.Testing/Extensions/ArtifactExtensions.cs
```diff
@@ -436,5 +428,22 @@ private static string TypeToSource(ContractParameterType type)
...
+            var formattedInheritance = new List<string>();
+            foreach (var type in inheritance)
+            {
+                if (type == typeof(SmartContract))
+                {
+                    formattedInheritance.Add($"{type.FullName}({initializeParam})");
+                }
+                else
+                {
+                    formattedInheritance.Add(type.FullName);
```
shargon, line 443: Null check?

Jim8y: in this pr no logic is changed. you want me to change the logic here?

shargon: Yes,  just for for avoid the notification

shargon (approved):

PR #1154, opened 2024-08-24
[Compiler Add] add pop count

Jim8y:
This pr adds nemuric method PopCount.


src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
```diff
@@ -570,4 +570,51 @@ private static void HandleBigIntegerEquals(MethodConvert methodConvert, Semantic
...
+        var endIntCheck = new JumpTarget();
+        methodConvert.Jump(OpCode.JMPIFNOT, endIntCheck);
+
+        // If within int range, mask with 0xFFFFFFFF
+        methodConvert.Push(0xFFFFFFFF);
+        methodConvert.AddInstruction(OpCode.AND);
+        var endMask = new JumpTarget();
+        methodConvert.Jump(OpCode.JMP, endMask);
+
+        // If larger than int, throw exception, cause too many check will make the script too long.
```
Jim8y, line 593: out scirpt size is limited, can not put too many checking logic here, thus only deal 32 bits value.

src/Neo.Compiler.CSharp/MethodConvert/StackHelpers.OpCodes.cs
```diff
@@ -6,6 +6,47 @@ namespace Neo.Compiler;
 
 partial class MethodConvert
 {
+
```
shargon, line 9:
```suggestion
```

shargon (approved):

PR #1155, opened 2024-08-25
update system methods.

Jim8y:


src/Neo.Compiler.CSharp/MethodConvert/Instruction.cs
```diff
@@ -59,6 +59,12 @@ static Instruction()
             }
         }
 
+        internal Instruction AddTarget(JumpTarget target)
+        {
+            target.Instruction = this;
```
shargon, line 64: This doesn't add the NOP, isn't it?

shargon (commented):
You also moved some methods and it complicate the review :(

Jim8y:
> You also moved some methods and it complicate the review :(

I know,,, i made this super huge pr,,,,,, it just keep growing while i was trying to make it work....... have being working on this for 15 hours without resting......

src/Neo.Compiler.CSharp/MethodConvert/Instruction.cs
Jim8y: This one will not, we can set jump target to existing opcode, if there is none, we can add a nop and attach the jumptarget to it.. This one inside the Instruction is for adding jump target to existing opcode, while the one with the same name `AddTarget` is for situations where we have to set jumptarget to Nop.

I made this change to make it easier to write the code, and make instructoins look shorter.....

Jim8y:
I tried my best to make it easier for reviewing, and avoided changes to the compile result.....

BUT, there is still a case that will change the artifact,,, which is I added `exception message`  while calling `throw` for nemuric methods.

src/Neo.Compiler.CSharp/MethodConvert/Instruction.cs
shargon: But it's working, isn't it?

Jim8y:
this will not be merged.

PR #1160, opened 2024-09-05
optimize HandleMathClamp

Hecate2:
It's not only a performance optimization. max == min should be allowed. This is the same as the behavior of C# https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/Math.cs#L636C13-L650C26


Hecate2:
OpCode.THROW also exists in many other files, mainly for checking value overflows.

Jim8y:
you can disable the gasassert for this pr

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Math.cs
```diff
@@ -306,41 +306,33 @@ private static void HandleMathClamp(MethodConvert methodConvert, SemanticModel m
...
-        methodConvert.AddInstruction(OpCode.REVERSE3);
-        methodConvert.AddInstruction(OpCode.DROP);
-        methodConvert.AddInstruction(OpCode.DROP);
-        methodConvert.Jump(OpCode.JMP, endTarget);
-        maxTarget.Instruction = methodConvert.AddInstruction(OpCode.NOP);
-        methodConvert.AddInstruction(OpCode.SWAP);
-        methodConvert.AddInstruction(OpCode.DROP);
-        methodConvert.Jump(OpCode.JMP, endTarget);
-        endTarget.Instruction = methodConvert.AddInstruction(OpCode.NOP);
+        methodConvert.AddInstruction(OpCode.REVERSE3);  // 10 0 5
```
Hecate2, line 319: //We do not use MAX and MIN after REVERSE3 because it's more expensive. MAX&MIN costs 1<<3 each
        //methodConvert.AddInstruction(OpCode.MAX);  // 10 5
        //methodConvert.AddInstruction(OpCode.MIN);  // 5
        //methodConvert.AddInstruction(OpCode.RET);

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Math.cs
```diff
@@ -306,41 +306,36 @@ private static void HandleMathClamp(MethodConvert methodConvert, SemanticModel m
...
-        methodConvert.AddInstruction(OpCode.ROT);// 5 10 10 0
-        methodConvert.AddInstruction(OpCode.DUP);// 5 10 10 0 0
-        methodConvert.AddInstruction(OpCode.ROT);// 5 10 0 0 10
-        methodConvert.Jump(OpCode.JMPLT, exceptionTarget);// 5 10 0
+        // Evaluation stack: value=5 min=0 max=10 <- top
+        methodConvert.AddInstruction(OpCode.OVER);  // 5 0 10 0
+        methodConvert.AddInstruction(OpCode.OVER);  // 5 0 10 0 10 <- top
+        methodConvert.Jump(OpCode.JMPLE, exceptionTarget);  // 5 0 10  // if 0 <= 10, continue execution
+        //methodConvert.Push("min>max");
         methodConvert.AddInstruction(OpCode.THROW);
```
shargon, line 315: but this use THROW also, isn't it?

Hecate2: Jim8y asked me to still use THROW

shargon: Could you update the description?

Jim8y (approved):

shargon (approved):

PR #1162, opened 2024-09-06
add feature as a standalone .nef file optimizer

Hecate2:
This allows any contract .nef file written in any language to be optimized.
Will conflict against https://github.com/neo-project/neo-devpack-dotnet/pull/1159 . Kept as draft for now.
Test: 
```bash
nccs SomeContract.nef --assembly --security-analysis
```
(There may be no additional print for --security-analysis if there is no vulnerability)


Hecate2:
Maybe nef and manifest need cloning before optimization as well

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -159,11 +159,7 @@ internal void Compile()
                 try
                 {
 #pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
-                    (nef, manifest, debugInfo) = Reachability.RemoveUnnecessaryJumps(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.ReplaceJumpWithRet(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.RemoveUnnecessaryJumps(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.ReplaceJumpWithRet(nef, manifest, debugInfo!.Clone() as JObject);
+                    (nef, manifest, debugInfo) = Neo.Optimizer.Optimizer.Optimize(nef, manifest, debugInfo: debugInfo!, optimizationType: Options.Optimize);
```
shargon, line 162:
```suggestion
                    (nef, manifest, debugInfo) = Neo.Optimizer.Optimizer.Optimize(nef, manifest, debugInfo: debugInfo!.Clone(), optimizationType: Options.Optimize);
```

src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -92,7 +92,43 @@ private static void Handle(RootCommand command, Options options, string[]? paths
...
+                        File.WriteAllText(Path.Combine(directory, filename + ".optimized.nef.txt"), dumpnef);
+                    }
+                    if (debugInfo != null)
+                        File.WriteAllBytes(Path.Combine(directory, filename + ".optimized.nefdbgnfo"), DumpNef.ZipDebugInfo(debugInfo.ToByteArray(true), filename + ".optimized.debug.json"));
+                    Console.WriteLine($"Optimization finished.");
+                    if (options.SecurityAnalysis)
+                        ReEntrancyAnalyzer.AnalyzeSingleContractReEntrancy(nef, manifest, debugInfo).GetWarningInfo(print: true);
+                    return;
+                }
+                if (extension != ".cs")
```
shargon, line 131:
```suggestion
                else if (extension != ".cs")
```

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -159,11 +159,7 @@ internal void Compile()
                 try
                 {
 #pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
-                    (nef, manifest, debugInfo) = Reachability.RemoveUnnecessaryJumps(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.ReplaceJumpWithRet(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.RemoveUncoveredInstructions(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.RemoveUnnecessaryJumps(nef, manifest, debugInfo!.Clone() as JObject);
-                    (nef, manifest, debugInfo) = Reachability.ReplaceJumpWithRet(nef, manifest, debugInfo!.Clone() as JObject);
+                    (nef, manifest, debugInfo) = Neo.Optimizer.Optimizer.Optimize(nef, manifest, debugInfo: debugInfo!.Clone(), optimizationType: Options.Optimize);
```
Hecate2, line 162:
```suggestion
                    (nef, manifest, debugInfo) = Neo.Optimizer.Optimizer.Optimize(nef, manifest, debugInfo: debugInfo!.Clone() as JObject, optimizationType: Options.Optimize);
```

src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -92,7 +92,43 @@ private static void Handle(RootCommand command, Options options, string[]? paths
...
+                    JObject? debugInfo;
+                    if (File.Exists(debugInfoPath))
+                        debugInfo = (JObject?)JObject.Parse(DumpNef.UnzipDebugInfo(File.ReadAllBytes(debugInfoPath)));
+                    else
+                        debugInfo = null;
+                    (nef, manifest, debugInfo) = Neo.Optimizer.Optimizer.Optimize(nef, manifest, debugInfo, optimizationType: CompilationOptions.OptimizationType.All);
+                    // Define the optimization type inside the manifest
+                    manifest.Extra ??= new JObject();
+                    manifest.Extra["nef"] = new JObject();
+                    manifest.Extra["nef"]!["optimization"] = CompilationOptions.OptimizationType.All.ToString();
```
shargon, line 116: It should take the value from args, isn't it?

Hecate2: > It should take the value from args, isn't it?

After considering twice, I think we may have to still set optimization to Experimental, regardless of user input, because only Experimental is supported for now. I am pushing a new commit on this

shargon: We allow `Basic` also, so it should use the configured in the arguments, otherwise we will forgot it in the future.

Hecate2: > We allow `Basic` also, so it should use the configured in the arguments, otherwise we will forgot it in the future.

Current basic optimizer cannot track and modify the manifest and the debuginfo, and cannot be used for optimizing .nef file directly.

Jim8y (approved):

shargon (approved):

PR #1165, opened 2024-09-11
optimizer supporting CALLA

Hecate2:


src/Neo.Compiler.CSharp/Optimizer/Analysers/InstructionCoverage.cs
```diff
@@ -77,6 +77,7 @@ public class InstructionCoverage
         /// value: sources of that jump target
         /// </summary>
         public Dictionary<Instruction, HashSet<Instruction>> jumpTargetToSources { get; init; }
+        public Dictionary<int, EntryType> pushaTargets { get; init; }
         public InstructionCoverage(NefFile nef, ContractManifest manifest)
```
shargon, line 81:
```suggestion

        public InstructionCoverage(NefFile nef, ContractManifest manifest)
```

src/Neo.Compiler.CSharp/Optimizer/Analysers/InstructionCoverage.cs
```diff
@@ -238,8 +241,23 @@ public BranchType CoverInstruction(int addr, Stack<TryStack>? stack = null)
...
+                    BranchType returnedType;
+                    if (instruction.OpCode == OpCode.CALLA)
+                    {
+                        returnedType = BranchType.ABORT;
+                        foreach (int callaTarget in pushaTargets.Keys)
+                        {
+                            BranchType singleCallaResult = CoverInstruction(callaTarget, stack);
+                            if (singleCallaResult < returnedType)
+                                returnedType = singleCallaResult;
+                            // TODO: if a PUSHA cannot be covered, do not add it as a CALLA target
```
shargon, line 253: It will work without this?

Hecate2: A way to find all impossible PUSHA is
1. Cover codes assuming all PUSHAs are valid
2. Find PUSHAs that are actually not covered
3. Cover codes, excluding all the uncovered PUSHAs
4. ...Until no new PUSHA is uncovered

shargon (approved):

src/Neo.Compiler.CSharp/Optimizer/Analysers/InstructionCoverage.cs
shargon: But it will work, without the optimizer, isn't it?

Hecate2: > But it will work, without the optimizer, isn't it?

Yes it works with current codes

Jim8y (approved):

PR #1170, opened 2024-09-18
test large overflows and underflows for EnsureIntegerInRange

Hecate2:
The correctness of OpCode.AND depends on the representation of BigInteger in bytes, and I am really suspicious whether AND is correct. Here are some examples of bytes representations.
https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.tobytearray?view=net-8.0
MOD is safe, no matter how BigInteger is implemented in bytes.


shargon (dismissed):

Jim8y:
Need unit test to demonstrate the problem.

Hecate2:
> Need unit test to demonstrate the problem. 

I have not come up with any test case to demonstrate that AND is wrong, but I will try it in 24 hrs

vncoelho:
For BigIntenger Negative values it may be the error if I understood correctly

shargon (commented):
We should ensure that is well covered with the current UT

Hecate2:
In tests I found that AND works well, while MOD is wrong. In C#, (-1 mod 5) returns -1. In other languages the result is often 4 😰

nan01ab:
> In tests I found that AND works well, while MOD is wrong. In C#, (-1 mod 5) returns -1. In other languages the result is often 4 😰

Other languages are Python?  C, Java, Rust etc. have the same semantics as C#.

Jim8y:
@Hecate2 test should cover signed and unsigned.

Hecate2:
> @Hecate2 test should cover signed and unsigned. 

ok

shargon (approved):

PR #1173, opened 2024-09-19
optimize DivRem

Hecate2:
Using MOD instead of l-l/r*r
Original implementation check and throw when the remainder overflows. However it is impossible for the remainder to overflow, because
```python
if MinValue < 0
  abs(remainder) < abs(MinValue)
if MinValue == 0
  0 < remainder < MaxValue
```
Therefore I cancelled the check.
But it is possible for the quotient to overflow in a single case, when the allowed range is [-128, 127] and I call DivRem(-128, -1). I did not check and throw in this case, because it is rare and expensive, and the original implementation did not check it.


Jim8y:
@Hecate2 please fix conflict

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
```diff
@@ -488,21 +488,22 @@ private static void HandleMathBigIntegerDivRem(MethodConvert methodConvert, Sema
...
         // Perform division
-        methodConvert.AddInstruction(OpCode.DUP);
-        methodConvert.AddInstruction(OpCode.ROT);
-        methodConvert.AddInstruction(OpCode.TUCK);
-        methodConvert.AddInstruction(OpCode.DIV);
+        methodConvert.AddInstruction(OpCode.DUP); // r, l, l
+        methodConvert.Push(2);
+        methodConvert.AddInstruction(OpCode.PICK);// r, l, l, r
+        methodConvert.AddInstruction(OpCode.DIV);  // r, l, l/r
+        // TODO: for types that is restricted by range, check l/r <= MaxValue
```
Jim8y, line 497: Are you going to work on this todo？

Hecate2: No. I think we can just allow overflow.

Jim8y: Will this be the same as C# behavior?

Jim8y: then can you please explain this as comment where others can learn.

Hecate2: > then can you please explain this as comment where others can learn.

Done

shargon (approved):

Jim8y (approved):

PR #1177, opened 2024-09-24
generate basic block jumps from coverage; enhance basic block analysis for contracts with NOP

Hecate2:
I found it wrong to generate basic block jumping graphs with only the last instruction of each block. The graph has to be generated from `InstructionCoverage`. This PR also enhances reentrancy analyzer.


shargon (approved):

Hecate2:
I think it still has problem in basic block tests. Yet the exceptions were catched in tests.

Hecate2:
I got the reason and the fix. Basic block analyzer does not consider jump targets that are not covered. We need to optimize the nef with RemoveUncoveredInstructions before running basic block tests.

shargon (approved):

Jim8y (approved):

PR #1178, opened 2024-09-24
Optimize BigInteger.Parse

shargon:
Now it cost less than 50% if the argument is a constant


src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
```diff
@@ -170,7 +170,25 @@ private static void HandleBigIntegerToByteArray(MethodConvert methodConvert, Sem
...
+            if (arguments.Count == 1 && arguments[0] is ArgumentSyntax { NameColon: null } arg)
+            {
+                // Optimize call when is a constant string
+
+                Optional<object?> constant = model.GetConstantValue(arg.Expression);
+
+                if (constant.HasValue && constant.Value is string strValue && BigInteger.TryParse(strValue, out var bi))
+                {
+                    // Insert a sequence point for debugging purposes
+                    using var sequence = methodConvert.InsertSequencePoint(arg.Expression);
```
Jim8y, line 183: just  a question, when should we insert the point?

Jim8y (approved):

Jim8y:
amazing, nice work.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
shargon: when you want to add the syntax to debug.info, more sequence points, will reference better the source with the nef, so debuggers or coverage will be better

PR #1184, opened 2024-09-29
[Compiler] fix object creation

Jim8y:
This pr fix the problem that we can not create Map or List object with initializer. 

The reason of this is that Map and List has no explicitly defined Property.


tests/Neo.Compiler.CSharp.TestContracts/Contract_PropertyMethod.cs
```diff
@@ -13,15 +17,44 @@ public static void testProperty2()
...
+        {
+            1, 2, 3, 4, 5
+        };
+    }
+
     public class Person
     {
         public string Name { get; set; }
-        public int Age { get; }
+        public int Age { get; set; }
```
shargon, line 48: We remove the test without the setter

tests/Neo.Compiler.CSharp.TestContracts/Contract_PropertyMethod.cs
```diff
@@ -1,3 +1,7 @@
+
```
shargon, line 1:
```suggestion
```

tests/Neo.Compiler.CSharp.TestContracts/Contract_PropertyMethod.cs
Jim8y: i dont get that?

shargon: This ensure that a peoperty without set, initialized in the constructor works

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Property_Method.cs
```diff
@@ -1,7 +1,10 @@
+using System;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Neo.SmartContract.Testing;
 using Neo.VM.Types;
 using System.Numerics;
+using Akka.Util.Internal;
```
shargon, line 6:
```suggestion
```
It's required?

shargon (approved):

PR #1185, opened 2024-09-30
cancel unnecessary constructors

Hecate2:
TODO in the future: regex may not be reliable enough


Jim8y (approved):

src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
```diff
@@ -34,7 +34,7 @@ public class CompilationEngine(CompilationOptions options)
         internal CompilationOptions Options { get; private set; } = options;
         private static readonly MetadataReference[] CommonReferences;
         private static readonly Dictionary<string, MetadataReference> MetaReferences = [];
-        private static readonly Regex s_pattern = new(@"^(Neo\.SmartContract\.Framework\.SmartContract|SmartContract\.Framework\.SmartContract|Framework\.SmartContract|SmartContract)$");
+        internal static readonly Regex s_pattern = new(@"^(Neo\.SmartContract\.Framework\.SmartContract|SmartContract\.Framework\.SmartContract|Framework\.SmartContract|SmartContract|Neo\.SmartContract\.Framework\.Nep17Token|Neo\.SmartContract\.Framework\.TokenContract|Neo.SmartContract.Framework.Nep11Token<.*>)$");
```
shargon, line 37: Maybe we should try to remove this regex

shargon (approved):

src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
shargon: @Hecate2 this is needed? it should check the hierarchy

shargon:
```
  static bool IsDerivedFromSmartContract(INamedTypeSymbol classSymbol, Regex pattern)
  {
      var baseType = classSymbol.BaseType;
      while (baseType != null)
      {
          if (pattern.IsMatch(baseType.ToString() ?? string.Empty))
          {
              return true;
          }
          baseType = baseType.BaseType;
      }
      return false;
  }
  ```

Hecate2: IsDerivedFromSmartContract returns true even when a asmart contract is derived for many times. e.g. 
class A: SmartContract
class B: A
class C: B
IsDerivedFromSmartContract returns true for B and C. But in B and C we may call `base`, or utilize the constructor of A.

PR #1187, opened 2024-09-30
[Compiler] add fix nep

Jim8y:
This pr adds some newly added NEP standards, and also updated the NEP checking error message.

I find it very long and hard to review, so i will only do NEP11 error message update in this pr, other updte will be in other coming prs.


src/Neo.SmartContract.Framework/NEPStandard.cs
```diff
@@ -23,19 +23,34 @@ public enum NepStandard
...
+        // NEP-30: Contract witness verification callback
+        // This standard defines a method that contracts can implement to allow their addresses
+        // to be used as transaction signers and pass witness checks in other code.
+        // It enables contracts to act as transaction senders, covering fees from their GAS balance,
+        // and to be verified in contexts like NEP-17 transfers.
+        // The 'verify' method is automatically called during transaction verification with the Verification trigger.
+        // Defined at https://github.com/neo-project/proposals/blob/master/nep-30.mediawiki
+        Nep30,
+        // This is the nickname of NEP-30.
+        Withdrawable
```
shargon, line 53: Maybe we should use Nep30, without the nickname

src/Neo.SmartContract.Framework/Interfaces/IWithdrawable.cs
```diff
@@ -19,7 +19,8 @@ namespace Neo.SmartContract.Framework.Interfaces;
 public interface IWithdrawable
```
shargon, line 19: I think that we should remove it and use only INEP30

src/Neo.SmartContract.Framework/NEPStandard.cs
Jim8y: withdrwable is being used before nep30 is defined, and i believe it is more readable than nep30. and moat importantly, it will not cause any coding problem to the contract.

src/Neo.SmartContract.Framework/Interfaces/IWithdrawable.cs
Jim8y: done, removed

shargon (approved):

PR #1189, opened 2024-10-02
reentrancy analyzer test

Hecate2:


shargon (commented):
It doesn't check if the method has the attribute

Hecate2:
> It doesn't check if the method has the attribute

The analyzer intakes .nef file instead of source code, and is unable to be aware of the `[NoReentrant]` attribute for now.

Jim8y:
is that possible that we add attribute to the debuginfo?@Shargon

Hecate2:
> is that possible that we add attribute to the debuginfo?@Shargon

I think this may not be the best methodology. Verifying the codes is better than trusting the manifest. And it may be not easy for other languages to have the same attributes

Jim8y:
> > is that possible that we add attribute to the debuginfo?@Shargon
> 
> I think this may not be the best methodology. Verifying the codes is better than trusting the manifest. And it may be not easy for other languages to have the same attributes

here we dont consider other languages.

roman-khimov:
NEP-25 and NEP-19 are shared by all of the ecosystem, so any changes there should be discussed separately.

shargon:
How it works this reentrancy analyzer? how it check that it has a protection?

Jim8y:
> NEP-25 and NEP-19 are shared by all of the ecosystem, so any changes there should be discussed separately.

@roman-khimov  This is C# compiler, nothing to do with NEP25 or NEP19 standards. And core team is not maintaining any languagues other than C#, we are not responsible for other languages.

Hecate2:
I think the best way is to symbolically execute the assembly of a method, and infer that the method cannot be re-entered.

roman-khimov:
>  This is C# compiler

I know, but

> is that possible that we add attribute to the debuginfo?

Looks like NEP-19 to me. While in fact what you want with attribute and what is suggested by @Hecate2 by

> trusting the manifest

is even closer to NEP-25. And this rings the bell for me.

Jim8y:
> NEP-19

Problem all mine, i did not know debuginfo is also an NEP~~~

Jim8y (approved):

tests/Neo.Compiler.CSharp.UnitTests/SecurityAnalyzer/UnitTest_Reentrancy.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Neo.Compiler.CSharp.UnitTests.SecurityAnalyzer
+{
+    [TestClass]
+    public class ReentrancyTests : DebugAndTestBase<Contract_Reentrancy>
+    {
+        [TestMethod]
+        public void Test_HasReentrancy()
+        {
+            ReEntrancyAnalyzer.ReEntrancyVulnerabilityPair v =
+                ReEntrancyAnalyzer.AnalyzeSingleContractReEntrancy(NefFile, Manifest);
```
shargon, line 15: AnalyzeSingleContractReEntrancy checks that the key is the same?

Hecate2: > AnalyzeSingleContractReEntrancy checks that the storage key is the same?

No. This is a test for the reentrancy analyzer, instead of the [NoReentrant] attribute

shargon (approved):

PR #1190, opened 2024-10-02
fix basicBlockContinuation after else

Hecate2:
In the original codes, if I have a contract like this:
```python
if True:
    A()
else:
    B()
C()
```
A() is covered, and then the address jumps to C(). When C() is covered and returned to the jump stack at the end of A(), we try to cover B(). B() ends at the beginning of C(), hitting the coverage cache at C() (because C() had been covered from A()), and forgets to register the basic block continuation from B() to C().


Jim8y:
any way to test your change?

Jim8y (approved):

Hecate2:
Except Contract_Abort, there is no compiler test contract that runs a complete if-else. However Contract_Abort tests will conflict with https://github.com/neo-project/neo-devpack-dotnet/pull/1185 . The instruction positions are changed without class contructors.

shargon (approved):

PR #1191, opened 2024-10-02
[Test] add instructions to the contract methods

Jim8y:
This pr disassemble the script and add corresponding instructions to the artifacts. 

This can be very useful to add new SystemCalls by implementing them in contract and compile to get instructions directly.

Its getting harder and harder to implement system methods, now with this pr, it can be done eaisly and directly in C#.


Jim8y:
Core part of the disassember is copied from Optimizer implemented by @Hecate2 but i have to move and adapt it into a seperate project due to cycle dependency issue.

src/Neo.Disassembler.CSharp/Disassembler.cs
```diff
@@ -0,0 +1,60 @@
...
+    private static readonly Regex RangeRegex = new(@"(\d+)\-(\d+)", RegexOptions.Compiled);
+
+    public static List<Instruction> ConvertScriptToInstructions(byte[] script)
+    {
+        var res = EnumerateInstructions(script);
+
+        return res.Select(x => x.instruction).ToList();
+    }
+
+    public static List<(int address, Instruction instruction)> ConvertMethodToInstructions(NefFile nef, JToken DebugInfo, string method)
```
Jim8y, line 22: this is the core part of this pr

shargon (commented):
Is not better to add the comment above the method, inside the summary or in a different  tag?

src/Neo.Disassembler.CSharp/Disassembler.cs
```diff
@@ -0,0 +1,60 @@
...
+        var address = 0;
+        var opcode = OpCode.PUSH0;
+        Instruction instruction;
+        for (; address < script.Length; address += instruction.Size)
+        {
+            instruction = script.GetInstruction(address);
+            opcode = instruction.OpCode;
+            yield return (address, instruction);
+        }
+        if (opcode != OpCode.RET)
```
shargon, line 57: Technically this is out of the script

Jim8y:
> Is not better to add the comment above the method, inside the summary or in a different tag?

I thought about that as well, but you have comments and artibute above method.

shargon:
```
/// <summary>
/// This summary.
/// </summary>
/// <remarks>
/// <---- HERE?
/// </remarks>
public void MyMethod()
{
    // Method implementation
}
```

Jim8y:
> ```
> /// <summary>
> /// This summary.
> /// </summary>
> /// <remarks>
> /// <---- HERE?
> /// </remarks>
> public void MyMethod()
> {
>     // Method implementation
> }
> ```
> 
> @Jim8y what do you think?

as long as it works, accept your suggestion as usual.

src/Neo.Disassembler.CSharp/Disassembler.cs
shargon: Some of them end's with ret, and other don't, I think this line should be removed

Jim8y: > Some of them end's with ret, and other don't, I think this line should be removed

I will check it, this code was adapted from the optimizer

Jim8y:
```C#
        for (; address < script.Length; address += instruction.Size)
        {
            instruction = script.GetInstruction(address);
            opcode = instruction.OpCode;
            yield return (address, instruction);
        }
```

will process instructions already, without caring about Ret. 

```C#
        if (opcode != OpCode.RET)
            yield return (address, Instruction.RET);
```
might be for ensuring the complete parsing.

shargon (approved):

shargon (dismissed):

src/Neo.SmartContract.Testing/Extensions/ArtifactExtensions.cs
```diff
@@ -36,10 +102,9 @@ public static class ArtifactExtensions
         /// <param name="nef">Nef file</param>
         /// <param name="generateProperties">Generate properties</param>
         /// <returns>Source</returns>
-        public static string GetArtifactsSource(this ContractManifest manifest, string? name = null, NefFile? nef = null, bool generateProperties = true)
+        public static string GetArtifactsSource(this ContractManifest manifest, string? name, NefFile? nef = null, bool generateProperties = true, JToken? debugInfo = null)
```
shargon, line 105:
```suggestion
        public static string GetArtifactsSource(this ContractManifest manifest, string? name = null, NefFile? nef = null, bool generateProperties = true, JToken? debugInfo = null)
```

Issue #1186, opened 2024-09-30
The implementation of  `MethodConvert.HandleStringTrim` in `SystemCall.String.cs`  has different semantics with C# stdlib


nan01ab:
`string.Trim()` is used to remove `whitespace` characters before and after a string.

https://github.com/neo-project/neo-devpack-dotnet/blob/add21354263b658f46c66b4738d0296c08fb3b7d/src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs#L298

The definition of `whitespace` character in C#, contains more than just space character(i.e. `' '`),  but also other characters(for example: `\t`, `\n`, detail: https://learn.microsoft.com/zh-cn/dotnet/api/system.string.trim?view=net-8.0).

But `MethodConvert.HandleStringTrim` just remove the space character(i.e. `' '`).

In fact,  the `trim`(or other names) method does not just remove space character(`' '`) in common programming languages.

Jim8y:
will fix it.

PR #1194, opened 2024-10-03
fix: string.Trim only trim ' '

nan01ab:
fix of #1186 

`string.Trim` uses same semantics with `char.IsWhiteSpace` to determine an ASCII char that is whitespace or not.


nan01ab:
@Jim8y

Jim8y:
need unit.test as well

shargon (commented):
UT required

vncoelho (commented):
I believe the change is correct following the logic here, but in fact a UT is necessary

nan01ab:
UT will be added later.

Jim8y:
> UT will be added later.

Wait, i am working on fixing the method itself, your UT will not work before that.

Jim8y:
@shargon @vncoelho @nan01ab please check the latest update and unit test.

Jim8y (approved):

shargon (commented):
I don't like inline methods here, but it works for you @Jim8y, I'm not against to

shargon (approved):

Jim8y:
> I don't like inline methods here, but it works for you @Jim8y, I'm not against to

@shargon I know what you mean clearly, and i actually also don't like it, but doing it this way makes it eaiser to implement the method and easier to review the code.

Implement these methods purely in opcode is very hard to maintain and review. I am thinking that maybe when we have multiple these unly implementations, we can group them and make these methods into something like `suger syntax`. 

We can figure out a better way later.

PR #1196, opened 2024-10-04
Avoid regex for smart contract checking

shargon:


src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
```diff
@@ -319,12 +318,12 @@ void Visit(INamedTypeSymbol classSymbol)
...
         }
 
-        internal static bool IsDerivedFromSmartContract(INamedTypeSymbol classSymbol, Regex pattern)
+        internal static bool IsDerivedFromSmartContract(INamedTypeSymbol classSymbol)
         {
             var baseType = classSymbol.BaseType;
             while (baseType != null)
             {
-                if (pattern.IsMatch(baseType.ToString() ?? string.Empty))
+                if (baseType.ToString() == "Neo.SmartContract.Framework.SmartContract")
```
shargon, line 326: This method checks the hierarchy

src/Neo.Compiler.CSharp/MethodConvert/SourceConvert.cs
```diff
@@ -168,7 +171,7 @@ internal static bool NeedInstanceConstructor(IMethodSymbol symbol)
             .DeclaringSyntaxReferences.Length == 0)
         // No explicit non-static constructor in class
         {
-            if (CompilationEngine.s_pattern.IsMatch(containingClass.BaseType?.ToString() ?? string.Empty))
+            if (s_pattern.IsMatch(containingClass.BaseType?.ToString() ?? string.Empty))
                 // class itself is directly inheriting smart contract; cannot have more base classes
```
shargon, line 175: We still need the regex for the last @Hecate2 changes, but I hope we can remove it.
@Hecate2, if `IsDerivedFromSmartContract` this shoud be true, isn't it?

Jim8y (approved):

src/Neo.Compiler.CSharp/MethodConvert/SourceConvert.cs
Hecate2: Yes

PR #1202, opened 2024-10-10
compile multiple .csproj in subdirectories

Hecate2:
Could be a benefit for large projects. Input a path of multiple contracts in multiple directories, and all of them get compiled.


Jim8y:
i am not sure if we need it or not, up to you @shargon to decide

shargon (approved):

shargon:
It doesn't hurt

PR #1203, opened 2024-10-12
replace STSFLD with PUSH const

Hecate2:
We have been always doing STSFLD in _initialize method for all static variables. They can be replaced with PUSH consts when the static variables are read. The major changes are in Peephole.cs  

Future steps:
- For expensive PUSHDATA instructions, if the pushed static variable is read very often, store it to static field at the beginning of method.


steven1227:
Nice step

shargon:
Love it!!!

shargon:
I have different concerns:
- There are case where it can be more expensive, some pushes are more expensive. Can we ensure that we choose the cheap one?.
- the script will be bigger, we have a limmit script size. If this limit is reached, how can we avoid this mechanism?
- whats happene in this case:
```csharp
class myClass
{
public int MyVar { get; } = 7 ;
public myClass()
   {
   MyVar = 3;
   }
}
```

src/Neo.Compiler.CSharp/Optimizer/AssetBuilder/OptimizedScriptBuilder.cs
```diff
@@ -48,7 +48,11 @@ public static Script BuildScriptWithJumpTargets(
                         throw new BadScriptException($"Target instruction of {i.OpCode} at new address {a} is deleted");
                     }
                     if (i.OpCode == OpCode.JMP || conditionalJump.Contains(i.OpCode) || i.OpCode == OpCode.CALL || i.OpCode == OpCode.ENDTRY)
-                        simplifiedScript.Add(BitConverter.GetBytes(delta)[0]);
+                        if (sbyte.MinValue <= delta && delta <= sbyte.MaxValue)
+                            simplifiedScript.Add(BitConverter.GetBytes(delta)[0]);
+                        else
+                            // TODO: build with _L version
```
shargon, line 54: Previously this also doesn't work, isn't it?

Hecate2: JMP_L problem has been solved in https://github.com/neo-project/neo-devpack-dotnet/pull/1203/commits/faf8dc5f611af8af3009e6979e70f89a5a9bad44

Hecate2:
> ```cs
>     public class MyClass : SmartContract.Framework.SmartContract
>     {
>         public static int MyVar = 7;
>         public MyClass()
>         {
>             MyVar = 3;
>         }
>         public int Main() => MyVar;
>     }
> ```

```python
# Method Start Neo.Compiler.CSharp.TestContracts.MyClass.Main()
00 INITSLOT 00-01 # 0 local variables, 1 arguments
# Code test-compiler.cs line 13: "MyVar"
03 LDSFLD0
# Method End Neo.Compiler.CSharp.TestContracts.MyClass.Main()
04 RET
# Method Start Neo.Compiler.CSharp.TestContracts.MyClass.MyClass()
05 INITSLOT 00-01 # 0 local variables, 1 arguments
08 LDARG0
09 CALL 05 # pos: 14 (offset: 5)
# Code test-compiler.cs line 11: "3"
# Code test-compiler.cs line 11: "MyVar = 3"
11 PUSH3
# Code test-compiler.cs line 11: "MyVar = 3"
# Code test-compiler.cs line 11: "MyVar = 3"
12 STSFLD0
# Method End Neo.Compiler.CSharp.TestContracts.MyClass.MyClass()
# Code test-compiler.cs line 12: "}"
13 RET
14 INITSLOT 00-01 # 0 local variables, 1 arguments
17 RET
18 INITSSLOT 01 # 1 static variables
20 PUSH7
21 STSFLD0
22 RET
23 NEWARRAY0
24 DUP
25 CALL EC # pos: 5 (offset: -20)
27 JMP E5 # pos: 0 (offset: -27)
29 RET
```

Hecate2:
>     * There are cases where it can be more expensive, some pushes are more expensive. Can we ensure that we choose the cheap one?

            [OpCode.PUSHINT128] = 1 << 2,
            [OpCode.PUSHINT256] = 1 << 2,
            [OpCode.PUSHDATA1] = 1 << 3,
            [OpCode.PUSHDATA2] = 1 << 9,
            [OpCode.PUSHDATA4] = 1 << 12,
These OpCodes are more expensive than LDSFLD and STSFLD (1 << 1).  

For PUSHDATA2 and 4, if they are executed twice in a method, it can be more expensive than LDSFLD. I have no immediate solution in a short time, but I have not really seen any contract in production that has to read a single huge piece of static DATA for multiple times.  

For PUSHDATA1, we spend 16+8+2 (INITSSLOT + PUSHDATA1 + STSFLD) to prepare it in the static fields, and get it at a price of 2 (LDSFLD) each time. If we always use PUSHDATA1 when we read it, we can have (16+8+2)/(8-2) = 4.3333 times before it get more expensive than INITSSLOT.

Also we can have (16+4+2)/(4-2) = 11 PUSHINT128s if we do not push it in the _initialize method.

Jim8y:
> I have different concerns:
> - There are cases where it can be more expensive, some pushes are more expensive. Can we ensure that we choose the cheap one?
> - the script will be bigger, we have a limmit script size. If this limit is reached, how can we avoid this mechanism?
> - whats happene in this case:
> ```csharp
> class myClass
> {
> public int MyVar { get; } = 7 ;
> public myClass()
>    {
>    MyVar = 3;
>    }
> }
> ```

this can be further optimized  for instance do a script size check.

Jim8y:
following this path, you can further optimize static fileds that are only updated in one method as the local const variable of that method. cause when i was imolememting out keyword, i use global static variable to store the out value, many things like this can only be done post compile

Jim8y:
assign myself a task as updating analyzer to apply relevent check.

Hecate2:
> following this path, you can further optimize static fileds that are only updated in one method as the local const variable of that method. cause when i was imolememting out keyword, i use global static variable to store the out value, many things like this can only be done post compile

Complex optimizations require delicate data flow analysis on the nef. Maybe I need to transform all writes (to eval stack, slot and storage) to static single assignment (SSA) form. I may be able to prepare it in half a year.

Hecate2:
>     * There are cases where it can be more expensive, some pushes are more expensive. Can we ensure that we choose the cheap one?

Is it OK that I do not optimize the static fields that are too expensive to push, and are pushed in a loop or for too many times?

Jim8y (dismissed):

Jim8y:
@shargon

src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
```diff
@@ -53,9 +54,16 @@ public BasicBlock(Dictionary<int, Instruction> instructions)
...
         }
 
-        //public void SetNextBasicBlock(BasicBlock block) => this.nextBlock = block;
-        //public void SetJumpTargetBlock1(BasicBlock block) => this.jumpTargetBlock1 = block;
-        //public void SetJumpTargetBlock2(BasicBlock block) => this.jumpTargetBlock2 = block;
+        public int FindFirstOpCode(OpCode opCode, ReadOnlyMemory<byte>? operand = null)
+        {
+            int addr = this.startAddr;
+            foreach (Instruction i in this.instructions)
+                if (i.OpCode == opCode && (operand == null || operand.Equals(i.Operand)))
```
vncoelho, line 61: There is no coverage for this method.

Also, maybe something like `(!operand.HasValue || operand.Value.Span.SequenceEqual(i.Operand.Span))`

Hecate2: Ok. I did plan to use this method but I found it meaningless later.

src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -48,7 +48,7 @@ public static bool IsSubclassOf(this ITypeSymbol type, string baseTypeName, bool
 
         public static bool IsVirtualMethod(this IMethodSymbol method)
         {
-            return method.IsAbstract || method.IsVirtual || method.IsOverride;
+            return method.IsAbstract || method.IsVirtual;
```
shargon, line 51: This change seems that affect different things, It's currently covered by UT, override with base call?

Hecate2: I think this is used to judge whether we need to PUSHA for a virtual or abstract base method. A method overriding another does not need a PUSHA to the address of itself. This is mainly used for [NoReentrant] Enter and Exit that overrides others. If override is considered virtual, the compiler adds an array of unnecessary PUSHA in _initialize to the address  of Enter and Exit.

shargon: Could you ensure that is covered the `override` by any unit test?

Hecate2: > Could you ensure that is covered the `override` by any unit test?

OK. Probably in a few hours.

Hecate2: Done in https://github.com/neo-project/neo-devpack-dotnet/commit/7edd7c173bac802137b1602e4b2dba6e8a2b766c

shargon (dismissed):

Hecate2:
I am going to remove NEWARRAY0, NEWSTRUCT0 and NEWMAP from optimization, because they are mutable. Even if an array in a static field is not explicitly written, it can be appended and SETITEM.

Jim8y:
@Hecate2 please do not append anything new here, do a new pr. 

> I am going to remove NEWARRAY0, NEWSTRUCT0 and NEWMAP from optimization, because they are mutable. Even if an array in a static field is not explicitly written, it can be appended and SETITEM.

Hecate2:
> @Hecate2 please do not append anything new to a pr that is reviewed here, do a new pr next time
> 
> > I am going to remove NEWARRAY0, NEWSTRUCT0 and NEWMAP from optimization, because they are mutable. Even if an array in a static field is not explicitly written, it can be appended and SETITEM.

Ok. But it can be wrong to include NEWARRAY0

shargon:
> @Hecate2 please do not append anything new to a pr that is reviewed here, do a new pr next time
> 
> > I am going to remove NEWARRAY0, NEWSTRUCT0 and NEWMAP from optimization, because they are mutable. Even if an array in a static field is not explicitly written, it can be appended and SETITEM.

is not new, it's a fix of this pr

Jim8y (approved):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Polymorphism.cs
```diff
@@ -10,14 +10,13 @@ public class UnitTest_Polymorphism : DebugAndTestBase<Contract_Polymorphism>
...
             Assert.AreEqual(14, Contract.Sum(5, 9));
-            AssertGasConsumed(1452840);
+            AssertGasConsumed(1233330);
             Assert.AreEqual(40, Contract.Mul(5, 8));
-            AssertGasConsumed(1470180);
-            Assert.AreEqual("testFinal", Contract.Test());
-            AssertGasConsumed(1197510);
-            Assert.AreEqual("testbase2.test2.test", Contract.Test2());
-            AssertGasConsumed(2171850);
-            Assert.AreEqual("abstractTestoverridenAbstract", Contract.AbstractTest());
```
shargon: Tests removed?

shargon (approved):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Polymorphism.cs
Jim8y: @Hecate2 please take care of this in another pr.

PR #1204, opened 2024-10-15
Cancel unnecessary instance method call

Hecate2:
Test contracts and GAS assertions are not updated because they will conflict against https://github.com/neo-project/neo-devpack-dotnet/pull/1203


Jim8y:
@shargon recent prs from @Hecate2 will need to update the gas consumption regularly, thus i disabled it here, will update the gas later all at once.

shargon (approved):

PR #1205, opened 2024-10-16
AllowOverflow option to avoid checking int range

Hecate2:
unchecked((int)bigInteger))
Will conflict. Draft for now.


Jim8y:
This pr has problem

```C#
 byte x = (byte)-1;
```
Overflow in constant value computation.

Hecate2:
Changed the methodology. Skip range checks and throws with compilation option

Hecate2:
Need a way to compile neo-devpack-dotnet\tests\Neo.Compiler.CSharp.TestContracts\Contract_BigInteger.cs with SimulateOverflow=true

Hecate2:
> Need a way to compile neo-devpack-dotnet\tests\Neo.Compiler.CSharp.TestContracts\Contract_BigInteger.cs with SimulateOverflow=true

Allowed CompilationOptions to be used in tests.
TestGasConsume is set to false.

src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -41,6 +41,7 @@ public static int Main(string[] args)
                 new Option<string>(new[] { "-o", "--output" }, "Specifies the output directory."),
                 new Option<string>("--base-name", "Specifies the base name of the output files."),
                 new Option<NullableContextOptions>("--nullable", () => NullableContextOptions.Annotations, "Represents the default state of nullable analysis in this compilation."),
+                new Option<bool>("--simulate-overflow", "Indicates whether to simulate overflow and underflow. If false, --checked does not take effect."),
```
shargon, line 44: --allow-overflow ?

By default we should test the contracts without this option

Jim8y:
This name is so scary....

Jim8y:
this can save tons of gas, yes,,,,,, but,,,,,,,

Jim8y:
rather than allow overflow, i think biginteger only is better

shargon:
We should not save gas if the security is involved

shargon:
Sorry miss click

Hecate2:
> We should not save gas if the security is involved 

No security problem is involved. This is just handling all int, long, short as BigInteger.

Hecate2:
> rather than allow overflow, i think biginteger only is better

Sometimes it's impossible to write BigInteger literally in codes. For example, shifting operations (>> and <<) in C# supports only int as the shifted count of bits. Then we have to write an explicit coversion (int) which wastes a range check.

Hecate2:
And our current default settings (with or without my PR) are --checked=false (--allow-overflow=false). This allows silent overflows (without or without my PR). I think this is even more dangerous. Maybe it's better to have default as --checked=false --allow-overflow=true, or --checked=true --allow-overflow=false

shargon:
we should detect `unchecked` keyword

Hecate2:
> we should detect `unchecked` keyword

We are actually detecting it (with or without this PR). But by default everything is unchecked.

Jim8y:
> > rather than allow overflow, i think biginteger only is better
> 
> Sometimes it's impossible to write BigInteger everywhere literally in codes. For example, shifting operations (>> and <<) in C# supports only int as the shifted count of bits. Also indexes of arrays have to be int. Then we have to write an explicit coversion (int) which wastes a range check.

excatly, just use what is allowed in biginteger.

Jim8y:
@shargon please review

Jim8y:
This is an option that should be false by default, so after you @shargon reviewed, please dont merge it, i will update the test to remove the option and revert gas change.

Jim8y:
> I still don't know why we need this option, which is the use case?

where user can handle the integer range and no need to do automatic range check. currently, every numerous operation requires a range check, its expensive.

shargon:
> > I still don't know why we need this option, which is the use case?
> 
> where user can handle the integer range and no need to do automatic range check. currently, every numerous operation requires a range check, its expensive.

expensive not because it's required for safety

tests/Neo.Compiler.CSharp.UnitTests/TestCleanup.cs
```diff
@@ -24,13 +24,24 @@ public class TestCleanup : TestCleanupBase
         private static readonly string TestContractsPath = Path.GetFullPath(Path.Combine("..", "..", "..", "..", "Neo.Compiler.CSharp.TestContracts", "Neo.Compiler.CSharp.TestContracts.csproj"));
         private static readonly string RootPath = Path.GetPathRoot(TestContractsPath) ?? string.Empty;
 
-        private static readonly Lazy<CompilationEngine> _compilationEngine = new(() => new CompilationEngine(new CompilationOptions
+        public static readonly CompilationOptions DefaultTestCompilationOptions = new CompilationOptions
         {
+            AllowOverflow = true,
```
shargon, line 29: We should leave the test as before, only create one test for this specific case

Hecate2: Done

Jim8y (approved):

shargon:
Can we discuss this pe in the meeting?

Hecate2:
> Can we discuss this pe in the meeting?

That's really great

Jim8y:
gas examples to demonstrate gas saving.

shargon:
Take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/1249

Jim8y:
close as can be partially replaced by using BigInteger, though not fully.

PR #1210, opened 2024-10-22
add gas price to artifact

Jim8y:
this pr adds the gas price to the artifact, such that we can easily check the optimization detail in git diff.


tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Recursion.cs
```diff
@@ -26,26 +26,26 @@ public abstract class Contract_Recursion(Neo.SmartContract.Testing.SmartContract
...
+    /// 00 : OpCode.INITSLOT 0001	[64 datoshi]
+    /// 03 : OpCode.LDARG0	[2 datoshi]
+    /// 04 : OpCode.PUSH0	[1 datoshi]
+    /// 05 : OpCode.EQUAL	[32 datoshi]
+    /// 06 : OpCode.JMPIFNOT 04	[2 datoshi]
+    /// 08 : OpCode.PUSHT	[1 datoshi]
+    /// 09 : OpCode.RET	[0 datoshi]
+    /// 0A : OpCode.LDARG0	[2 datoshi]
+    /// 0B : OpCode.PUSH0	[1 datoshi]
+    /// 0C : OpCode.LT	[8 datoshi]
```
shargon, line 38: @Jim8y Why some of them are not regenerated?

Jim8y: I am not sure what you mean, looks like all have being updated.

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Assignment.cs
```diff
@@ -55,36 +55,36 @@ public abstract class Contract_Assignment(Neo.SmartContract.Testing.SmartContrac
...
+    /// 18 : OpCode.LDLOC0	[2 datoshi]
+    /// 19 : OpCode.JMP 05	[2 datoshi]
+    /// 1B : OpCode.PUSH2	[1 datoshi]
+    /// 1C : OpCode.DUP	[2 datoshi]
+    /// 1D : OpCode.STLOC0	[2 datoshi]
+    /// 1E : OpCode.DROP	[2 datoshi]
+    /// 1F : OpCode.LDLOC0	[2 datoshi]
+    /// 20 : OpCode.PUSH1	[1 datoshi]
+    /// 21 : OpCode.EQUAL	[32 datoshi]
+    /// 22 : OpCode.ASSERT	[1 datoshi]
```
shargon, line 86: @Jim8y Why some use space and other tabs?

Jim8y:
@shargon fixed. the reason of that is because the only difference between old artifact and new artifact is now many space are there, while the comparison actually will ignore space.... thus it will not be updated.

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract2.cs
```diff
@@ -28,7 +28,7 @@ public abstract class Contract2(Neo.SmartContract.Testing.SmartContractInitializ
     /// Script: VwECDAECAwTbMHBoEs5A
     /// 00 : OpCode.INITSLOT 0102 [64 datoshi]
     /// 03 : OpCode.PUSHDATA1 01020304 [8 datoshi]
-    /// 09 : OpCode.CONVERT 30 [8192 datoshi]
+    /// 09 : OpCode.CONVERT (Buffer) [8192 datoshi]
```
shargon, line 31: 0x30=Buffer?

Jim8y: done

src/Neo.Disassembler.CSharp/Disassembler.cs
```diff
@@ -60,16 +61,47 @@ private static (int start, int end) GetMethodStartEndAddress(string name, JToken
...
+    public static string InstructionToString(this Instruction instruction, bool addPrice = true)
     {
         var opcode = instruction.OpCode.ToString();
         var operand = instruction.Operand;
 
-        if (operand.IsEmpty || operand.Length == 0)
+        var addprice = 0L;
+        string ret;
+
+        if (operand.IsEmpty)
```
shargon, line 72:
```suggestion
        if (operand.IsEmpty || operand.Length == 0)
```

shargon (approved):

Issue #1209, opened 2024-10-22
[MethodImpl(MethodImplOptions.AggressiveInlining)] does not handle methods defined by arrow expressions


Hecate2:
```csharp
[MethodImpl(MethodImplOptions.AggressiveInlining)]
private static ByteString MyMethod(object foo) => StdLib.Serialize((foo, bar));
```
https://github.com/Hecate2/neo-devpack-dotnet/blob/add21354263b658f46c66b4738d0296c08fb3b7d/src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs#L45
`syntax.Body` is null for this case, because no explicit method body is defined in a BlockSyntax {}.
`syntax.ExpressionBody` may be useful for such cases, but I do not know how to compile the method for now.

PR #1211, opened 2024-10-22
fix inline for null body method

Jim8y:
this pr fixes https://github.com/neo-project/neo-devpack-dotnet/issues/1209


Hecate2:
```csharp
        private static void ArrowInline(int a, int b) => CallMethodThatReturnsInt(a, b);
```

shargon (approved):

shargon:
> ```cs
>         [MethodImpl(MethodImplOptions.AggressiveInlining)]
>         private static void ArrowInline(int a, int b) => CallMethodThatReturnsInt(a, b);
> ```
> 
> The compiler should drop the returned value in this case.

If it's private yes, if it's public... Maybe we can create a new issue for it

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Inline.cs
```diff
@@ -35,5 +35,11 @@ public void Test_NestedInline()
             Assert.AreEqual(new BigInteger(3), Contract.TestInline("inline_nested"));
             AssertGasConsumed(1071930);
         }
+
+        [TestMethod]
+        public void Test_ArrowMethod()
+        {
+            Assert.AreEqual(new BigInteger(3), Contract.ArrowMethod());
```
shargon, line 42: artifacts not updated

Jim8y:
> ```cs
>         [MethodImpl(MethodImplOptions.AggressiveInlining)]
>         private static void ArrowInline(int a, int b) => CallMethodThatReturnsInt(a, b);
> ```
> 
> The compiler should drop the returned value in this case.

Fixed.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Inline.cs
Jim8y: done

shargon (approved):

PR #1213, opened 2024-10-24
UseNz, UseIsNull & FoldNotInJmp to handle if(a==null), if(a==0)

Hecate2:
Terrible conflicts with https://github.com/neo-project/neo-devpack-dotnet/pull/1203 . I will have to change codes when they are being merged.
Artifacts and GAS assertions not updated.


Hecate2:
When we compile `if (a == null)`, we used:
```
PUSHNULL
NOTEQUAL
JMPIFNOT
```
Now we can fold it to `ISNULL NOT JMPIFNOT`, and then to `ISNULL JMPIF`.

Hecate2:
(Will handle the GAS and test artifacts in 24 hours)

shargon (dismissed):

Jim8y:
Impreassive

Jim8y (approved):

PR #1214, opened 2024-10-24
nullable Storage.Get; refactor TokenContract

Hecate2:
This PR is a bit risky and may affect existing contract codes. I do not mind abandoning it if not good.
`Storage.Get` may actually return null. Ignoring it without generating any warning for the contract developer is not good.
Also it is not the best to blindly translate null to BigInteger 0. And it costs GAS to check whether a ByteString is null, when converting it to BigInteger.
Judging null increases contract size, but can be optimized by https://github.com/neo-project/neo-devpack-dotnet/pull/1213 .


Jim8y:
This will not affect existing contract, unless they compile it again.... so, not really a big deal. But the behavior should be properly commented. Such that developers can see how it works when they use it.

src/Neo.SmartContract.Framework/ByteString.cs
```diff
@@ -54,11 +54,6 @@ IEnumerator IEnumerable.GetEnumerator()
         [OpCode(OpCode.CONVERT, StackItemType.ByteString)]
         public static extern explicit operator ByteString(byte[] buffer);
 
-        [OpCode(OpCode.DUP)]
```
shargon, line 57: I prefer to avoid this change for the moment, only nullables

Hecate2:
NUMEQUAL can be used in TokenContract

src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -112,7 +112,9 @@ public extern ByteString this[byte[] key]
         [OpCode(OpCode.CAT)]
         [OpCode(OpCode.SWAP)]
         [Syscall("System.Storage.Get")]
-        public extern ByteString Get(ByteString key);
+#pragma warning disable CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```
shargon, line 115: Enable nullable at the beginning of the file?

src/Neo.SmartContract.Framework/TokenContract.cs
```diff
@@ -17,35 +17,49 @@ namespace Neo.SmartContract.Framework
 {
     public abstract class TokenContract : SmartContract
     {
-        protected const byte Prefix_TotalSupply = 0x00;
-        protected const byte Prefix_Balance = 0x01;
+        protected static readonly ByteString Prefix_TotalSupply = "\x00";
```
shargon, line 20: Why not const?

src/Neo.SmartContract.Framework/UInt256.cs
```diff
@@ -27,11 +27,13 @@ public extern bool IsValid
         {
             [OpCode(OpCode.DUP)]
             [OpCode(OpCode.ISTYPE, "0x28")] //ByteString
-            [OpCode(OpCode.SWAP)]
+            [OpCode(OpCode.JMPIF, "06")]  // to SIZE
```
shargon, line 30: This is cheaper?

src/Neo.SmartContract.Framework/TokenContract.cs
Hecate2: ByteString is really cheaper when used for storage.get/put. Also static readonly ByteString can be optimized to const now.

src/Neo.SmartContract.Framework/UInt256.cs
Hecate2: Yes, and better for NZ/ISNULL and folding NOT in optimization

Hecate2: Also this avoids executing SIZE on a possibly null input, and further saves `if (input is null)` before using isValid

src/Neo.SmartContract.Framework/Services/StorageMap.cs
Hecate2: OK. Maybe I shall just delete the pragma comments

src/Neo.SmartContract.Framework/UInt256.cs
shargon: This should be in a different PR

src/Neo.SmartContract.Template/templates/neocontractnep17/Nep17Contract.cs
```diff
@@ -60,14 +60,14 @@ public static void SetOwner(UInt160 newOwner)
 
         public static new void Burn(UInt160 account, BigInteger amount)
         {
-            if (IsOwner() == false)
```
Jim8y, line 63: actually `== false`  is a better format as it is read friendly, but anyway. i dont mind.

Jim8y:
@shargon please review

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_UIntTypes.cs
```diff
@@ -18,18 +18,18 @@ public void UInt160_ValidateAddress()
...
 
             // False
 
             Assert.IsFalse(Contract.ValidateAddress(InvalidUInt160.InvalidType));
-            AssertGasConsumed(1047930);
-            Assert.ThrowsException<TestException>(() => Contract.ValidateAddress(InvalidUInt160.Null));
-            AssertGasConsumed(1047330);
+            AssertGasConsumed(1047450);
+            Assert.IsFalse(Contract.ValidateAddress(InvalidUInt160.Null));
+            AssertGasConsumed(1047450);
```
shargon, line 28: There are cases where the cost is greater

Hecate2: No greater.
For invalid input,
  1047930 -> 1047450
For null input,
  Exception with 1047330 -> no exception with 1047450

shargon (commented):
I prefer to move the uint optimizations in a different PR

Hecate2:
> I prefer to move the uint optimizations in a different PR

Ok I am on it. `is null` will be still used in TokenContract for this PR.

Jim8y (dismissed):

src/Neo.SmartContract.Framework/TokenContract.cs
```diff
@@ -17,35 +17,49 @@ namespace Neo.SmartContract.Framework
...
                 throw new Exception("The argument \"owner\" is invalid.");
-            StorageMap balanceMap = new(Storage.CurrentContext, Prefix_Balance);
-            return (BigInteger)balanceMap[owner];
+            ByteString? balanceStored = Storage.Get(Prefix_Balance + owner);
+            return (BigInteger)balanceStored!;
         }
 
         protected static bool UpdateBalance(UInt160 owner, BigInteger increment)
         {
-            StorageMap balanceMap = new(Storage.CurrentContext, Prefix_Balance);
```
shargon, line 41: It's better to use Storage map, it's more secure, less humans erros using concat keys

src/Neo.SmartContract.Framework/TokenContract.cs
```diff
@@ -17,35 +17,49 @@ namespace Neo.SmartContract.Framework
...
-        protected const byte Prefix_Balance = 0x01;
+        protected static readonly ByteString Prefix_TotalSupply = "\x00";
+        protected static readonly ByteString Prefix_Balance = "\x01";
 
         public abstract string Symbol { [Safe] get; }
 
         public abstract byte Decimals { [Safe] get; }
 
-        [Stored(Prefix_TotalSupply)]
-        public static BigInteger TotalSupply { [Safe] get; protected set; }
```
shargon, line 28: Why avoid the Stored?

src/Neo.SmartContract.Framework/TokenContract.cs
Hecate2: StorageMap costs much more GAS. But I am ok to revert it to StorageMap

src/Neo.SmartContract.Framework/TokenContract.cs
Hecate2: > Why avoid the Stored?

I think it was because Stored does not support a ByteString prefix, but I actually forgot why. I will find the reason.

Hecate2: > Why avoid the Stored?

Yes it is because ByteString is not supported by `Stored`

shargon: Then better to do this change in a different PR, this is focus in be nullable

src/Neo.SmartContract.Framework/TokenContract.cs
shargon: We should improve it, but not always cheaper is better, bad storage use could cost more if they are hacked

Hecate2: > We should improve it, but not always cheaper is better, bad storage use could cost more if they are hacked

TokenContract reverted

src/Neo.SmartContract.Framework/TokenContract.cs
```diff
@@ -41,8 +41,9 @@ protected static bool UpdateBalance(UInt160 owner, BigInteger increment)
             StorageMap balanceMap = new(Storage.CurrentContext, Prefix_Balance);
             BigInteger balance = (BigInteger)balanceMap[owner];
             balance += increment;
-            if (balance < 0) return false;
-            if (balance.IsZero)
```
shargon, line 45: Maybe we should optimize IsZero to use NumEqual

shargon (dismissed):

Jim8y (approved):

PR #1221, opened 2024-11-01
fix artifact method script

Hecate2:
The scripts in artifact methods are wrong for instructions like PUSHDATA1. 
https://github.com/neo-project/neo-devpack-dotnet/blob/ba53357262da864ca9a9cc7119197d06933889f3/tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_ByteString.cs#L39
DOS9oOWlvQDCgMO/QA==
0C E4BDA0E5A5BD00C280C3BF 40
    /// 00 : OpCode.PUSHDATA1 E4BDA0E5A5BD00C280C3BF [8 datoshi]
    /// 0D : OpCode.RET [0 datoshi]
0x0C is PUSHDATA1, but we need another 1 byte to represent the length of the pushed data, instead of just starting to push the data.

Conflicts everywhere.


src/Neo.SmartContract.Testing/Extensions/ArtifactExtensions.cs
```diff
@@ -417,12 +418,10 @@ private static string CreateSourceMethodFromManifest(ContractMethodDescriptor me
                 var instructions = Disassembler.CSharp.Disassembler.ConvertMethodToInstructions(nefFile, debugInfo, method.Name);
```
shargon, line 418: Fix this method ?

Hecate2: > Fix this method ?

This method works well, but there was a mistake translating Instruction to bytes (which is really dirty work). My optimizer had codes for the translation, but I need to copy most of those codes for a simple translation in this use case.

shargon (approved):

PR #1222, opened 2024-11-04
delete unused CALLT method token in nef

Hecate2:


shargon (commented):
In what cases this happens?

Hecate2:
> In what cases this happens?

https://github.com/neo-project/neo-devpack-dotnet/blob/07720d0927cfb41fa8e337a6e124c399067565da/src/Neo.Compiler.CSharp/MethodConvert/ExternConvert.cs#L98

shargon (approved):

Jim8y (approved):

vncoelho (approved):

PR #1227, opened 2024-11-06
update nepcheck

Jim8y:
This pr updates the nep check, i forgot that this task is not finished~~ and find some issue in existing check.

### DIFF
--- a/src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.Nep11.cs
+++ b/src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.Nep11.cs
@@ -63,51 +63,56 @@ internal static partial class ContractManifestExtensions
             errors.Add(new CompilationException(DiagnosticId.IncorrectNEPStandard,
             $"Incomplete or unsafe NEP standard {NepStandard.Nep11.ToStandard()} implementation: totalSupply, it's return type is not an integer"));
 
-
-        if (balanceOfMethod1 == null)
+        if (balanceOfMethod1 == null && balanceOfMethod2 == null)
+        {
             errors.Add(new CompilationException(DiagnosticId.IncorrectNEPStandard,
             $"Incomplete or unsafe NEP standard {NepStandard.Nep11.ToStandard()} implementation: balanceOf, it is not found in the ABI"));
+        }
 
-        if (balanceOfMethod1 is { Safe: false })
-            errors.Add(new CompilationException(DiagnosticId.IncorrectNEPStandard,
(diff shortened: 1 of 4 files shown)


src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.Nep11.cs
```diff
@@ -63,51 +63,56 @@ private static System.Collections.Generic.List<CompilationException>
...
+                    $"Incomplete or unsafe NEP standard {NepStandard.Nep11.ToStandard()} implementation: balanceOf, it is not safe, you should add a 'Safe' attribute to the balanceOf method"));
 
-        if (balanceOfMethod2 is { Safe: false })
-            errors.Add(new CompilationException(DiagnosticId.IncorrectNEPStandard,
-            $"Incomplete or unsafe NEP standard {NepStandard.Nep11.ToStandard()} implementation: balanceOf, it is not safe, you should add a 'Safe' attribute to the balanceOf method"));
-
-        if (balanceOfMethod2 != null && balanceOfMethod2.ReturnType != ContractParameterType.Integer)
-            errors.Add(new CompilationException(DiagnosticId.IncorrectNEPStandard,
-            $"Incomplete or unsafe NEP standard {NepStandard.Nep11.ToStandard()} implementation: balanceOf, it's return type is not an integer"));
+            if (balanceOfMethod2 != null && balanceOfMethod2.ReturnType != ContractParameterType.Integer)
```
nan01ab, line 97: Why not use a method that like `Check(..., params (ReturnType, Expected)`  instead of these `if`  checks

Jim8y: make it work first, then optimize it later, you can do that check method.

src/Neo.Compiler.CSharp/Manifest/ContractManifestExtensions.Nep17.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        // Check Transfer event
+        var transferEvent = manifest.Abi.Events.FirstOrDefault(e =>
+            e.Name == "Transfer");
+
+        if (transferEvent == null)
+            errors.Add(new CompilationException(DiagnosticId.IncorrectNEPStandard,
+                $"Incomplete NEP standard {NepStandard.Nep17.ToStandard()} implementation: Transfer event is not found in the ABI"));
+
+        if (transferEvent != null)
```
shargon, line 127:
```suggestion
        else
```

shargon (approved):

PR #1229, opened 2024-11-06
make it easier to debug the compiler from script

Jim8y:
Its painful to debug the compiler from scripts, thus having this pr to enable us to insert debug message into script. Any other solution that works better is welcome.

### DIFF
--- a/src/Neo.Compiler.CSharp/Diagnostic/DebugInstruction.cs
+++ b/src/Neo.Compiler.CSharp/Diagnostic/DebugInstruction.cs
@@ -1,0 +1,13 @@
+using System.Runtime.CompilerServices;
+using Neo.VM;
+
+namespace Neo.Compiler;
+
+internal partial class MethodConvert
+{
+    void InstructionDebug([CallerLineNumber] int lineNumber = 0, [CallerMemberName] string caller = null)
+    {
+        Push("DEBUG: at line " + lineNumber + " (" + caller + ")");
+        AddInstruction(OpCode.DROP);
+    }
+}


src/Neo.Compiler.CSharp/Diagnostic/DebugInstruction.cs
```diff
@@ -0,0 +1,13 @@
+using System.Runtime.CompilerServices;
+using Neo.VM;
+
+namespace Neo.Compiler;
+
+internal partial class MethodConvert
+{
+    void InstructionDebug([CallerLineNumber] int lineNumber = 0, [CallerMemberName] string caller = null)
```
shargon, line 8: Why not increase the `DebugInfo` information?

Jim8y: I dont really know how to use DebugInfo~~ can you please give me an example how to debug with debuginfo?

Jim8y: Well, checked that debuginfo is not what i want, it build relationship betweeen opcode and source code, while what i want to know is where the opcode is added in the compiler....

Jim8y: Let me check if i can update debuginfo ~~

shargon: > Well, checked that debuginfo is not what i want, it build relationship betweeen opcode and source code, while what i want to know is where the opcode is added in the compiler....

DebugInfo have a reference between the source and the script

Jim8y: not the source, i need the compiler location, instead the contract source code location, studing on adding it to debuginfo.

Jim8y: Have tried to do so, but it is much harder to ahiceve that since the optimizer changes the debuginfo in multiple places that is hard to trace, yet what i want to achieve can be done easily.

shargon: But do you want to use this only for you, or do you have plans to add this call everywhere, if it's for you, I'm ok like this, otherwise, I can try to do it in `DebugInfo`

Jim8y: if it can be added to DebugInfo and shown in the artifact, it would be great, cause i need to know where the opcode is being added to the script in the compiler. If that is hard to implement, this one would be the great option since i can check it in the script itself.

shargon: Let me help you

Issue #1224, opened 2024-11-05
Mint multiple NEP11 tokens causes Peek out of bounds: 0/0


GreenCakes-png:
Copied this example "[examples/Example.SmartContract.SampleRoyaltyNEP11Token](https://github.com/neo-project/neo-devpack-dotnet/tree/master/examples/Example.SmartContract.SampleRoyaltyNEP11Token)"

I added a method to mint multiple tokens at once like this:

public static void MultiMint(UInt160 to)
{
    Mint(to);
    Mint(to);
    Mint(to);
    Mint(to);
}

However when I call this method, the NEO CLI throws "Peak out of bounds..." error

![image](https://github.com/user-attachments/assets/aed21a16-41dd-4aeb-b422-1289b85b5659)

Could someone help identify the cause or suggest a solution?
Is it not possible to mint multiple tokens at once?

Jim8y:
I'll take a look

Hecate2:
Could you provide any contract source code, the compiler version, or the .nef, .manifest.json, .nefdbgnfo?
Also you may try neo-fairy-test.

Jim8y:
@shargon can you please check this logic here, the problem is caused by this check:

https://github.com/neo-project/neo-devpack-dotnet/blob/65654df63f60c436f1c8d292975a6bbe831b652d/src/Neo.Compiler.CSharp/MethodConvert/PropertyConvert.cs#L128

if you call a `storebacked` property twice, you will be able to trigger this problem, cause the first call will initialize the property and make it none null, then the second call to the property will jump all the way to the bottom of the `get` method.

Jim8y:
simply do this will reproduce the issue. 

```C#
        public static void MultiMint(UInt160 to)
        {
            TotalSupply++;
            TotalSupply++;
        }
```

shargon:
> @shargon can you please check this logic here, the problem is caused by this check:
> 
> https://github.com/neo-project/neo-devpack-dotnet/blob/65654df63f60c436f1c8d292975a6bbe831b652d/src/Neo.Compiler.CSharp/MethodConvert/PropertyConvert.cs#L133
> 
> if you call a `storebacked` property twice, you will be able to trigger this problem, cause the first call will initialize the property and make it none null, then the second call to the property will jump all the way to the bottom of the `get` method.

this is supposed to be the instance, and because is static, should be null :S

Hecate2:
```
[...]

PR #1230, opened 2024-11-07
property `get;` loads value in stack if static field not null

Hecate2:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/1224
By the way, what if we want to `set;` the value as null? The value can only be read as 0, if it is actually null.


Jim8y:
wait a sec, i am adding ut and fixing remaining problems

shargon (approved):

Jim8y (approved):

PR #1231, opened 2024-11-07
WriteInTryAnalzyer

Hecate2:
Writing in try is dangerous, because depending on the position where the exception is thrown, it is always ambiguous whether a write in a try is reverted.

### DIFF
--- a/src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
@@ -36,6 +36,13 @@ namespace Neo.Optimizer
         public HashSet<BasicBlock> jumpTargetBlocks = new();  // jump target of the last instruction of this basic block
         public HashSet<BasicBlock> jumpSourceBlocks = new();
         public BranchType branchType = BranchType.UNCOVERED;
+        public TryType tryType = TryType.NONE;
+        public HashSet<BasicBlock> catchBlocks = new();
+        public HashSet<BasicBlock> finallyBlocks = new();
+        public HashSet<BasicBlock> endFinallyBlocks = new();
+        public HashSet<BasicBlock> trySourceBlocks = new();
+        public HashSet<BasicBlock> catchSourceBlocks = new();
+        public HashSet<BasicBlock> finallySourceBlocks = new();
 
         public BasicBlock(int startAddr, List<Instruction> instructions)
         {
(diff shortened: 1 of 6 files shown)


Hecate2:
We cannot prevent others writing a try, but we should avoid writing in try in our own contracts.
If we do not try, our contracts have no ambiguity.
```csharp
using System;
using System.ComponentModel;
using Neo;
using Neo.SmartContract.Framework.Attributes;
using Neo.SmartContract.Framework.Services;

namespace callerA
{
    [DisplayName("callerA")]
    [ManifestExtra("Author", "NEO")]
    [ManifestExtra("Email", "developer@neo.org")]
    [ManifestExtra("Description", "This is a callerA")]
    [ContractPermission("*", "*")]
    public class callerA : SmartContract
    {
        public static bool Main(UInt160 callee, bool throwInB)
        {
            Storage.Put(new byte[] { 0xce }, 1);
            Storage.Put(new byte[] { 0xcc }, 1);
            try
            {
                Storage.Put(new byte[] { 0xcc }, 2);  // valid
                Contract.Call(callee, "main", CallFlags.All, new object[] { Runtime.ExecutingScriptHash, throwInB });
                return true;
            }
            catch
            {
                return false;
            }
        }
        public static void PrivateWrite()
        {
            Storage.Put(new byte[] { 0xce }, 2);
        }
    }
}
```
```csharp
using System;
using System.ComponentModel;
using Neo;
using Neo.SmartContract.Framework.Attributes;
using Neo.SmartContract.Framework.Services;

namespace calleeB
{
    [DisplayName("calleeB")]
    [ManifestExtra("Author", "NEO")]
    [ManifestExtra("Email", "developer@neo.org")]
    [ManifestExtra("Description", "This is a calleeB")]
    [ContractPermission("*", "*")]
    public class calleeB : SmartContract
    {
        public static bool Main(UInt160 callerA, bool throw_)
        {
            Storage.Put(new byte[] { 0xcc }, 1);
            PrivateWrite();
            try
            {
                Contract.Call(callerA, "privateWrite", CallFlags.All);
                throw new Exception("B");
            }
            catch { }
            if (throw_)
                throw new Exception("B");
            return true;
        }
        public static void PrivateWrite()
        {
            Storage.Put(new byte[] { 0xce }, 2);
        }
    }
}
```
```python
from neo_fairy_client import FairyClient

c = FairyClient(fairy_session='throwLab')
calleeB = c.virutal_deploy_from_path('calleeB/bin/sc/calleeB.nef')
callerA = c.virutal_deploy_from_path('callerA/bin/sc/callerA.nef')
print(c.invokefunction('main', [calleeB, True]))
print(c.get_storage_with_session(b'\xcc', contract_scripthash=callerA))  # b'\xcc': 2
print(c.get_storage_with_session(b'\xce', contract_scripthash=callerA))  # b'\xce': 1
print(c.get_storage_with_session(b'\xcc', contract_scripthash=calleeB))  # None
print(c.get_storage_with_session(b'\xce', contract_scripthash=calleeB))  # None
```
[throwLab.zip](https://github.com/user-attachments/files/17658711/throwLab.zip)

Jim8y:
The syntax follows C#, operations before the exception in a try will not be revertred,,,, that is C# logic.....

shargon:
I think is well, the problem is if 0xA0 will be 2, but it's one, so I think snapshot is working as expected
https://github.com/neo-project/neo-devpack-dotnet/pull/1232/files

Hecate2:
There is a case to be even more ambiguous:
```python
def A():
    write("A")
    throwInA()

try:
    A()
    throwInTry()
...
```
If we throw at `throwInA()`, the write is reverted. If we do no throw at `throwInA()`, but throw at `throwInTry()`, the write is not reverted.
My analyzer should be taking this case into consideration, when A() may be called in a try block. But I am still suspecting about the correctness of my InstructionCoverage. It may be too lazy and skip registering try stack info for basic blocks in A().

Hecate2:
> I think is well, the problem is if 0xA0 will be 2, but it's one, so I think snapshot is working as expected
> https://github.com/neo-project/neo-devpack-dotnet/pull/1232/files

Yes. Current VM is working well. But if a developer writes a storage.Put in a try, we may be able to manipulate the inputs to execute the try or skip or revert it, in an unexpected way.

Hecate2:
I will write a brand-new analyzer for this. I need to analyze the scope of each try, each catch and each finally.

Hecate2:
Going to suspend this PR because it will take a longer time. May open again with a rebase in the future. Preparations at https://github.com/neo-project/neo-devpack-dotnet/pull/1234

PR #1233, opened 2024-11-08
add ut for property

Jim8y:
this pr adds ut to property to demonstrate that property actually does not work correctly.


Hecate2:
My fault, not applying `NeedInstanceConstructor` everywhere.
https://github.com/Jim8y/neo-devpack-dotnet/pull/3

shargon (approved):

shargon:
Ut fault

Hecate2:
> Ut fault

Because of calling ++ on null

Jim8y:
> Ut fault

I am debugging this, but without a proper tool, it is very hard to locate where the opcode is being processed in the compiler, very hard to debug.

Hecate2:
Will be hard to tackle. Smart contract _initialize only handles static constructors, and ignores non-static initial values.
https://github.com/Jim8y/neo-devpack-dotnet/blob/4aa527f1c7de1ddc17f46a131b48208a423ced2a/tests/Neo.Compiler.CSharp.TestContracts/Contract_Property.cs#L11
https://github.com/Jim8y/neo-devpack-dotnet/blob/4aa527f1c7de1ddc17f46a131b48208a423ced2a/src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs#L464-L466

Hecate2:
Should be ready for merge

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -416,6 +422,25 @@ private void ProcessClass(SemanticModel model, INamedTypeSymbol symbol)
...
+                    // Otherwise, use the current model (for metadata fields)
+                    var fieldModel = syntaxRef != null
+                        ? ((ISourceAssemblySymbol)field.ContainingAssembly).Compilation.GetSemanticModel(syntaxRef.SyntaxTree)
+                        : model;
+                    return (Field: field, Model: fieldModel);
+                })
+                .ToArray();
+
+            // Process each field using its symbol
+            ContractFields.ForEach(f => AddStaticField(f.Field));
```
Hecate2, line 442:
```suggestion
            foreach ((IFieldSymbol f, _) in ContractFields)
                AddStaticField(f);
```

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -29,6 +29,7 @@
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text;
+using Akka.Util.Internal;
```
Hecate2, line 32:
```suggestion
```

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
Hecate2: array.ForEach is using Akka.Util.Internal

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -243,12 +245,43 @@ private void ProcessFieldInitializer(SemanticModel model, IFieldSymbol field, Ac
...
                     initializer = syntax.Initializer;
                 }
-                if (initializer is null) return;
-                model = model.Compilation.GetSemanticModel(syntaxNode.SyntaxTree);
+
+                if (initializer is null)
+                {
+                    if(_context.ContractFields.Any(f =>
+                           SymbolEqualityComparer.Default.Equals(f.Field, field)) &&
+                       (field.Type.GetStackItemType() == StackItemType.Integer || field.Type.GetStackItemType() == StackItemType.Integer))
```
Hecate2, line 253: f.Field.Type😰?

Hecate2:
If we apply the methodology in https://github.com/neo-project/neo-devpack-dotnet/pull/1233/commits/d552e2e335d1ea247c0c302ce9c0e177e5a73f8f , we try to convert all non-static fields (in a class that directly inherit SmartContract) to static. This can bring trouble when we read the non-static fields, and we will need to change many codes related to IFieldSymbol and IdentifierNameSyntax.

Jim8y:
will be replaced by other prs.

PR #1235, opened 2024-11-08
Compiler source info into `DebugInfo`

shargon:
Alternative to https://github.com/neo-project/neo-devpack-dotnet/pull/1229
Related to https://github.com/neo-project/proposals/issues/184
Require https://github.com/neo-project/neo-devpack-dotnet/pull/1240

This introduce a new property, `sequence-points-v2`, inside the method in debugInfo:

sequence-points remains as before, not changed:

```json
{
  "sequence-points": [
    "20[0]12:20-12:21",
    "21[0]12:24-12:25",
    "22[0]12:20-12:25",
    "23[0]12:20-12:25",
    "25[0]12:28-12:35",
    "32[0]12:20-12:35",
    "33[0]12:20-12:35",
    "35[0]12:13-12:36",
    "37[0]13:9-13:10"
  ]
}
```

`sequence-points-v2` was included in json format, with more information: 

```json
{
  "sequence-points-v2": {
    "20": {
      "source": {
        "document": 0,
        "location": "12:20-12:21"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "21": {
      "source": {
        "document": 0,
        "location": "12:24-12:25"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "22": {
      "source": {
        "document": 0,
        "location": "12:20-12:25"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "23": {
      "source": {
        "document": 0,
        "location": "12:20-12:25"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "25": {
      "source": {
        "document": 0,
        "location": "12:28-12:35"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "32": {
      "source": {
        "document": 0,
        "location": "12:20-12:35"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "33": {
      "source": {
        "document": 0,
        "location": "12:20-12:35"
      },
      "compiler": {
        "file": "Expression.cs",
        "line": 36,
        "method": "ConvertExpression"
      }
    },
    "35": {
      "source": {
        "document": 0,
[...]

PR #1237, opened 2024-11-09
DebugInfoBuilder for sequence point v2

Hecate2:
For https://github.com/neo-project/neo-devpack-dotnet/pull/1235#pullrequestreview-2425173592 . Not used for master, but for PR 1235.


Hecate2:
If seq v2 uses array instead of object, NEED CHANGES

shargon:
> If seq v2 uses array instead of object, NEED CHANGES

I think we can use object, if optimized: false, and if optimized:true it can be an array

Hecate2:
> > If seq v2 uses array instead of object, NEED CHANGES
> 
> I think we can use object, if optimized: false, and if optimized:true it can be an array

OK I can handle it

Hecate2:
The scripts in the artifact of Contract_Instance and Contract_Polymorphism will be automatically deleted by Disassembler, because of a mismatch between the offset in the manifest and the debugInfo

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -304,7 +304,7 @@ public JObject CreateDebugInformation(string folder = "")
                     // Create sequence-points-v2
 
                     var v2 = new JObject();
-                    v2["optimization"] = CompilationOptions.OptimizationType.None.ToString().ToLowerInvariant();
+                    v2["optimization"] = this.Options.Optimize.ToString().ToLowerInvariant();
```
shargon, line 307: If the method was not optimized, it will be good to avoid the change, maybe some methods are not optimized

src/Neo.Disassembler.CSharp/Disassembler.cs
```diff
@@ -40,14 +40,8 @@ public static List<Instruction> ConvertScriptToInstructions(byte[] script)
 
             // Note: Require Debug extended type, we can't relate the abi to the NEP19 and without it, name is compiler dependant
 
-            if (method["abi"] is JObject abi)
-            {
-                var parsedMethod = ContractMethodDescriptor.FromJson(abi);
-                parsedMethod.Offset = abiMethod.Offset; // It can be optimized, we avoid to check the offset
-                if (!parsedMethod.Equals(abiMethod)) continue;
-
+            if (method["abi"] is JObject && int.Parse(method["abi"]!["offset"]!.AsString()) == abiMethod.Offset)
```
shargon, line 43: I prefer to use the parsed version, with offset

Hecate2: Shall I revert this? I do not really understand what to do here.

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
Hecate2: Generally it can be expensive to check whether a method is optimized or not. I need to visit all the Instructions that may be covered by the method.

src/Neo.Disassembler.CSharp/Disassembler.cs
shargon: Yes, only removing the offset = line should work

src/Neo.Compiler.CSharp/Optimizer/AssetBuilder/DebugInfoBuilder.cs
```diff
@@ -69,12 +69,65 @@ static class DebugInfoBuilder
...
+                        }
+                        else
+                            // previousSequencePoint unchanged
+                            writeAddr = previousSequencePoint.ToString();
+                        if (newSequencePointsV2[writeAddr] == null)
+                            newSequencePointsV2[writeAddr] = oldContentArray;
+                        else
+                            newSequencePointsV2[writeAddr] = ((JArray)newSequencePointsV2[writeAddr]!).Concat(oldContentArray).ToArray();
+                    }
+                    method["sequence-points-v2"] = newSequencePointsV2;
```
shargon, line 120: if there is only one entry it should be good to preserve the object

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
shargon: If we check the previous sequencePoints with the current one and are different, it was optimized, isn't it?

src/Neo.Compiler.CSharp/Optimizer/AssetBuilder/DebugInfoBuilder.cs
Hecate2: > if there is only one entry it should be good to preserve the object, also, in what cases will be more (inline methods?)? it was checked that if it's the same information, don't duplicate it?

For example, mutiple RET instructions may be compressed to 1 RET.

Hecate2: > if there is only one entry it should be good to preserve the object, also, in what cases will be more (inline methods?)? it was checked that if it's the same information, don't duplicate it?

Something like `JToken.DeepEquals` comparer is needed for a HashSet, if we need to eliminate the duplicates. Not easy.

PR #1241, opened 2024-11-12
fix default value

Jim8y:
this pr focus on fixing the default value of property and field.


src/Neo.Compiler.CSharp/MethodConvert/PropertyConvert.cs
```diff
@@ -122,21 +122,13 @@ private void ConvertStorageBackedProperty(IPropertySymbol property, AttributeDat
...
-                Jump(OpCode.JMPIFNOT_L, endTarget);
-            }
-            else if (NeedInstanceConstructor(Symbol))
-            {
-                // Check class
-                Jump(OpCode.JMPIF_L, endTarget);
-            }
+            Push("PropertyGet");
+            AddInstruction(OpCode.DROP);
+
```
Jim8y, line 131:
```suggestion

```

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ClassInit.cs
```diff
@@ -31,7 +31,7 @@ public void Test_InitInt()
         public void Test_InitializationExpression()
         {
             Contract.TestInitializationExpression();
-            AssertGasConsumed(1275690);
+            AssertGasConsumed(4032390);
         }
```
shargon, line 35: Wtf?? Thats good?

Jim8y: This is because setitem is very very expensive, but actually this is the right one....

When we adopt the dynamic opcode price solution, the price can drop like 99%.

Jim8y: I'll also do some evaluation to see if there is something wrong here.

Jim8y: Figured out that the gas here is actually 30 times larger than the actual datoshi being calculated, as the gas fee are calculated with a factor, which is 30.

shargon: Could you explain me a little bit the change, to know why it cost now 4 times more?

tests/Neo.SmartContract.Framework.UnitTests/Services/BackedStorageTest.cs
```diff
@@ -73,5 +73,11 @@ public void Test_Kind(Func<BigInteger?> getter, Func<BigInteger?> publicGetter,
 
             Assert.AreEqual(123, publicGetter());
         }
+
+        [TestMethod]
+        public void TestInc()
+        {
+            // Contract.TestInc();
```
shargon, line 80: Remove?

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ClassInit.cs
Jim8y: Previously, when we create an object, it the fileds of that object has no  initializer, it will be null in that object, however, in C#, value types will have a default value even if there is no initializer.

for that test, it has many fields and non of them has initializer, thus they are not initialized, with this pr, those fields will be set a default value. thus having a lot of extra `setitems`,,,,, and its expensive.

shargon: Why not use PACK?

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
```diff
@@ -243,11 +245,22 @@ private void ProcessFieldInitializer(SemanticModel model, IFieldSymbol field, Ac
                     syntaxNode = syntax;
                     initializer = syntax.Initializer;
                 }
-                if (initializer is null) return;
-                model = model.Compilation.GetSemanticModel(syntaxNode.SyntaxTree);
+
+                if (initializer is null)
+                {
+                    if (field.Type.GetStackItemType() == StackItemType.Boolean ||
```
shargon, line 251: If it's nullable this will happend also?

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ClassInit.cs
Jim8y: > Why not use PACK?

i did not implement this~~~logic of using pack or setitem has nothing to do with this pr to be honest.

src/Neo.Compiler.CSharp/MethodConvert/MethodConvert.cs
Jim8y: only if its defined as nullable, otherwise it will be the same.

tests/Neo.SmartContract.Framework.UnitTests/Services/BackedStorageTest.cs
shargon: This is expected to be commented?

Jim8y: I'll remove it.

Jim8y, line 80:
```suggestion
```

tests/Neo.SmartContract.Framework.UnitTests/Services/BackedStorageTest.cs
```diff
@@ -73,5 +73,6 @@ public void Test_Kind(Func<BigInteger?> getter, Func<BigInteger?> publicGetter,
 
             Assert.AreEqual(123, publicGetter());
         }
+        }
```
shargon, line 76:
```suggestion
```

shargon (approved):

shargon (approved):

PR #1242, opened 2024-11-13
fix Biginteger default.

Jim8y:
This pr fix the BigInteger default value.

When we define a BigInteger, it will not be assigned a default value like int, but it will remain as null.

but when it is defined in a class/struct, the defualt constructor of of class/struct will assign it a default value.


Hecate2:
In what case do we utilize BigInteger default value as null? `default(BigInteger)` returns 0.

Jim8y:
> In what case do we utilize BigInteger default value as null? `default(BigInteger)` returns 0.

Well, i might not be clear, default value of BigInteger is 0, but it will not be assigned automatically when we define a BigInteger. I have tried, C# syntax analyer wont allow you to use an uninitialized BigInteger actually, it will show you an error.

image
`Local variable 'a' might not be initialized before accessing`

Hecate2:
https://github.com/neo-project/neo-devpack-dotnet/blob/34926a0f52503f02f7720c99cf82be377670a890/src/Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj#L4
I think enabling nullable is enough (in PR 1214)

Jim8y:
> https://github.com/neo-project/neo-devpack-dotnet/blob/34926a0f52503f02f7720c99cf82be377670a890/src/Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj#L4
> 
> 
> I think enabling nullable is enough (in PR 1214)

I personally dont think we should enable nullable in contract, not really being much help to contract but introduces many complex work in compiler.... but since we already have it, so, lets complete 1214 then.

Hecate2:
> > https://github.com/neo-project/neo-devpack-dotnet/blob/34926a0f52503f02f7720c99cf82be377670a890/src/Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj#L4
> > 
> > I think enabling nullable is enough (in PR 1214)
> 
> I personally dont think we should enable nullable in contract, not really being much help to contract but introduces many complex work in compiler.... but since we already have it, so, lets complete 1214 then.

Nullable simply gives more hints to the user, and does not affect the compiler. If nullable is false, everything can be silently null, being more dangerous.

Jim8y:
> > > https://github.com/neo-project/neo-devpack-dotnet/blob/34926a0f52503f02f7720c99cf82be377670a890/src/Neo.SmartContract.Framework/Neo.SmartContract.Framework.csproj#L4
> > > 
> > > I think enabling nullable is enough (in PR 1214)
> > 
> > 
> > I personally dont think we should enable nullable in contract, not really being much help to contract but introduces many complex work in compiler.... but since we already have it, so, lets complete 1214 then.
> 
> Nullable simply gives more hints to the user, and does not affect the compiler. If nullable is false, everything can be silently null, being more dangerous.

Not in this pr, i mean compiler level nullable support that we already have, nullable value requires extra checks and increases the complexity.

Hecate2:
In this PR we are trying to mark BigInteger declared in method. However, if we apply this methodology, then for all types of variables declared in methods, they are all null, and all have to be checked with many efforts. Thus using <Nullable> in csproj brings more efficient and accurate checks, for free.
And for the potential drawback of <Nullable> making the compiler more complex, the complexity is actually not brought by a <Nullable> property in a csproj. It is always possible to write a `?` after a type, no matter whether <Nullable> is enabled in the csproj.

src/Neo.Compiler.CSharp/MethodConvert/Helpers/StackHelpers.cs
```diff
@@ -232,6 +232,13 @@ private void Push(object? obj)
 
     private Instruction PushDefault(ITypeSymbol type)
     {
+        // Among all integer types, BigInteger has no default value.
+        // To manage this, we will enforce all BigInteger be initialized while defined in Analyzer.
+        if (type.Name == nameof(BigInteger))
+        {
+            return AddInstruction(OpCode.PUSHNULL);
```
shargon, line 239: why here NULL and 0 in `ConvertHelpers`?

Hecate2: This PR seems to give the declaration `BigInteger a` (in a method as a local variable, WITHOUT explicit initial value assignment) a default value NULL. Meanwhile, it keeps declaration in class field a default value 0. I think this PR is not necessary now, because a local variable is initialized with INITSLOT, and originally has NULL. Also we have nullable in contracts everywhere, and the IDE gives hints (green underlines in visual studio) on use of uninitialized local variables.

Jim8y: This pr is before @Hecate2 fixed the property and field issue, now i will reevaluate and have another pr to replace this one.

PR #1243, opened 2024-11-14
Clean remarks and add type information

shargon:
Improve artifact's comments


Jim8y (approved):

Jim8y:
A quick question, is that possible that we also print the called function name?

PR #1244, opened 2024-11-14
only enable static property to have stored attribute

Jim8y:
This pr sets limitation on StoredAttribute, only allow static property to have it. This makes it easier to compile and avoid mistakes.


Hecate2:
https://github.com/neo-project/neo-devpack-dotnet/commit/4623f38fb52fbf128b04981b5ec3b800b7f91a9f
Check out this from PR 1233 and merge it into master should solve the problem.

PR #1247, opened 2024-11-19
fix convert

Jim8y:
this pr fix a bug in the framework where convert to integer is performed on an integer.


src/Neo.SmartContract.Framework/ByteString.cs
```diff
@@ -56,9 +56,10 @@ IEnumerator IEnumerable.GetEnumerator()
 
         [OpCode(OpCode.DUP)]
         [OpCode(OpCode.ISNULL)]
-        [OpCode(OpCode.JMPIFNOT, "0x04")]
+        [OpCode(OpCode.JMPIFNOT, "0x06")]
         [OpCode(OpCode.DROP)]
         [OpCode(OpCode.PUSH0)]
+        [OpCode(OpCode.JMP, "0x04")]
```
Jim8y, line 62: Without this convert will be performed on `0`

Hecate2 (approved):

src/Neo.SmartContract.Framework/ByteString.cs
shargon: We don't have tests for it?

Jim8y: test wont fail, cause convert an integer to an integer is a valid operation,,, wont change any result.

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_NEP11.cs
```diff
@@ -101,11 +101,12 @@ public abstract class Contract_NEP11(Neo.SmartContract.Testing.SmartContractInit
...
     /// 5C : DROP [2 datoshi]
     /// 5D : PUSH0 [1 datoshi]
-    /// 5E : CONVERT 21 'Integer' [8192 datoshi]
-    /// 60 : RET [0 datoshi]
+    /// 5E : RET [0 datoshi]
+    /// 5F : CONVERT 21 'Integer' [8192 datoshi]
+    /// 61 : RET [0 datoshi]
     /// </remarks>
     [DisplayName("balanceOf")]
     public abstract BigInteger? BalanceOf(UInt160? owner);
```
Jim8y, line 112: @shagon, say, test on NEP17.BanalceOf is a test for that,,, before this pr, it was convert integer to integer.

shargon: This info is awesome :)

tests/Neo.SmartContract.Framework.UnitTests/TestingArtifacts/Contract_SupportedStandard17Enum.cs
```diff
@@ -94,11 +94,12 @@ public abstract class Contract_SupportedStandard17Enum(Neo.SmartContract.Testing
...
     /// 59 : ISNULL [2 datoshi]
-    /// 5A : JMPIFNOT 04 [2 datoshi]
+    /// 5A : JMPIFNOT 05 [2 datoshi]
     /// 5C : DROP [2 datoshi]
     /// 5D : PUSH0 [1 datoshi]
-    /// 5E : CONVERT 21 'Integer' [8192 datoshi]
-    /// 60 : RET [0 datoshi]
+    /// 5E : RET [0 datoshi]
+    /// 5F : CONVERT 21 'Integer' [8192 datoshi]
+    /// 61 : RET [0 datoshi]
```
shargon, line 102: Why we have two rets? we should jump only to one isn't it?

Jim8y: I will check it, not sure about where it come from.

Hecate2: > Why we have two rets? we should jump only to one isn't it?

If an unconditional JMP jumps to a RET, my optimizer replaces the JMP with RET.

Jim8y: I see, it should come from the optimizer, the optimizer optimzied the jump to ret to save gas. 

Logic here is if it runs the first ret, it will return, otherwise the jump will skip the first ret, and run the convert. Here ret is just return statement. And it works.

nan01ab (approved):

tests/Neo.SmartContract.Framework.UnitTests/TestingArtifacts/Contract_SupportedStandard17Enum.cs
shargon: @Hecate2 why dont jump to the final ret?

shargon (approved):

tests/Neo.SmartContract.Framework.UnitTests/TestingArtifacts/Contract_SupportedStandard17Enum.cs
Jim8y: > @Hecate2 why dont jump to the final ret?

then there will be an extra jump cost. it is actually unnecessary. test shows that we can have multiple ret in the methods.

PR #1248, opened 2024-11-20
WriteInTryAnalzyer

Hecate2:
Correcting https://github.com/neo-project/neo-devpack-dotnet/pull/1231

### DIFF
--- a/src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
@@ -30,6 +30,20 @@ namespace Neo.Optimizer
     public class BasicBlock
     {
         public readonly int startAddr;
+        private int lastAddrCache = -1;
+        public int lastAddr
+        {
+            get
+            {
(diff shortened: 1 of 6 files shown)


src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
```diff
@@ -30,6 +30,20 @@ namespace Neo.Optimizer
     public class BasicBlock
     {
         public readonly int startAddr;
+        private int lastAddrCache = -1;
+        public int lastAddr
```
shargon, line 34: If it's public the name should start with uppercase, but it seems all are like this

shargon (approved):

src/Neo.Compiler.CSharp/Optimizer/Analysers/BasicBlock.cs
Jim8y: Uppercase only for property?

Jim8y:
Can you brifly explain what this pr's solution? @Hecate2

Jim8y (approved):

PR #1249, opened 2024-11-21
Test optimizer

shargon:
First was good
```
/// 00 : PUSH2 [1 datoshi]
/// 01 : RET [0 datoshi]
```

Second overflow was done good

We don't check the overflow, it's well done when it's safe


Jim8y:
for Biginteger we dont check range, yes.

Jim8y (approved):

Jim8y:
I fixed the conflict for you and please check if it is ok and then merge.

Jim8y (approved):

nan01ab (approved):

PR #1250, opened 2024-11-22
use ConcurrentDictionary for Parallel.ForEach

Hecate2:
May fail in large contracts if not fixed.
```
System.InvalidOperationException
  HResult=0x80131509
  Message=Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
  Source=System.Private.CoreLib
  StackTrace:
   at System.ThrowHelper.ThrowInvalidOperationException_ConcurrentOperationsNotSupported() in System\ThrowHelper.cs:line 450
```


Jim8y (approved):

shargon (commented):
Lgtn

PR #1256, opened 2024-11-28
fix & test WriteInTry analyzer

Hecate2:
If THROW or ABORT is executed in catch, should return THROW or ABORT immediately

### DIFF
--- a/src/Neo.Compiler.CSharp/Optimizer/Analysers/TryCatchFinallyCoverage.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/Analysers/TryCatchFinallyCoverage.cs
@@ -177,22 +177,23 @@ namespace Neo.Optimizer
                         if (tryType == TryType.TRY && allTry[tryBlock].catchBlock != null)
                         {
                             tryStack.Push((tryBlock, null, TryType.CATCH, true));
-                            if (CoverSingleTry(allTry[tryBlock].catchBlock!, tryStack) == BranchType.ABORT)
-                                return BranchType.ABORT;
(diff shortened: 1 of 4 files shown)


Jim8y (approved):

Hecate2:
Wait one. It's still wrong.

Hecate2:
Corrected

src/Neo.Compiler.CSharp/Optimizer/Analysers/TryCatchFinallyCoverage.cs
```diff
@@ -178,20 +178,22 @@ public BranchType CoverSingleTry(BasicBlock currentBlock,
...
                         {
                             tryStack.Push(new(tryBlock, endFinallyBlock, TryType.FINALLY, true));
-                            currentBlock = allTry[tryBlock].finallyBlock!;
+                            nextBlock = allTry[tryBlock].finallyBlock!;
                         }
                         else
                         {
                             allTry[tryBlock].endingBlocks.Add(endFinallyBlock);
-                            currentBlock = endFinallyBlock;
+                            nextBlock = endFinallyBlock;
```
shargon, line 194: We don't need to set also currentBlock?

Hecate2: Better for debugging. If we overwrite currentBlock, we do not know the original currentBlock

shargon (approved):
Well tested

nan01ab (approved):

PR #1258, opened 2024-12-04
test try analyzer in recursion

Hecate2:


Jim8y:
I would rather banning this type of usage... smart contract should not be so complex~~ but anyway.

Jim8y (approved):

Hecate2:
> I would rather banning this type of usage... smart contract should not be so complex~~ but anyway.

Anyway we need to analyze the contract before we know that the contract is writing in try (recursively)

shargon (commented):
LGTM

PR #1259, opened 2024-12-05
optimize IsValid

Hecate2:
A continuation of a cancelled optimization in the past
https://github.com/neo-project/neo-devpack-dotnet/pull/1214/commits/34926a0f52503f02f7720c99cf82be377670a890#diff-c681a7389a2321587cf8377a222c861e641e5a942a1b6aea8c7e522074d92112
because a new standalone PR was requested


shargon (approved):

Hecate2:
GAS updated. A GAS consumption increase in a case because the old codes throw exception, while current codes returns the result.

shargon:
> GAS updated. A GAS consumption increase in a case because the old codes throw exception, while current codes returns the result.

It's better a false

Hecate2:
Approval and merge?

src/Neo.SmartContract.Framework/UInt160.cs
```diff
@@ -29,11 +29,13 @@ public extern bool IsValid
         {
             [OpCode(OpCode.DUP)]
             [OpCode(OpCode.ISTYPE, "0x28")] //ByteString
-            [OpCode(OpCode.SWAP)]
+            [OpCode(OpCode.JMPIF, "06")]  // to SIZE
```
Jim8y, line 32: is this because Null itself has a type thus no need to do extra null check?

Jim8y (approved):

shargon (approved):

PR #1260, opened 2024-12-10
cancel more unnecessary constructor call

Hecate2:
This PR moves codes for static and non-static field initialization to the creation of object


Hecate2:
(The method `ProcessFields(SemanticModel model)` can also be deleted)

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_MemberAccess.cs
```diff
@@ -25,30 +25,28 @@ public abstract class Contract_MemberAccess(Neo.SmartContract.Testing.SmartContr
     /// Unsafe method
     /// </summary>
     /// <remarks>
-    /// Script: VwEACxASwEo0KXBoEM43AABBz+dHlgwDbXNnQc/nR5ZoEc5Bz+dHlmg0GkHP50eWQA==
+    /// Script: VwEADAVoZWxsbxASwHBoEM43AABBz+dHlgwDbXNnQc/nR5ZoEc5Bz+dHlmg0CEHP50eWQA==
```
Hecate2, line 28: The script here seems to be longer. This is because the old script is
```python
00 INITSLOT 01-00 # 1 local variables, 0 arguments
03 PUSHNULL
04 PUSH0
05 PUSH2
06 PACK
```
And the new script is
```python
00 INITSLOT 01-00 # 1 local variables, 0 arguments
03 PUSHDATA1 68-65-6C-6C-6F # as text: "hello"
10 PUSH0
11 PUSH2
12 PACK
```
In the old version we initialize all the fields with default values, and then fill them with the initial values written in contract. In the new version the field is directly initialized by string "hello".

src/Neo.Compiler.CSharp/MethodConvert/Helpers/CallHelpers.cs
```diff
@@ -73,6 +73,8 @@ private void CallInstanceMethod(SemanticModel model, IMethodSymbol symbol, bool
 
         var (convert, methodCallingConvention) = GetMethodConvertAndCallingConvention(model, symbol);
 
+        if (convert != null && convert.Instructions.Count == 1 && convert.Instructions[0].OpCode == OpCode.RET && !symbol.IsExtern)
```
Jim8y, line 76: may i ask the meaning of this checking? if possible, please add some comment here

Hecate2: It means that `convert` contains only a single instruction `RET`

Jim8y: what i mean is why these checking, like why one count, what if there are `nop, ret`, why not `extern`

Jim8y (approved):

src/Neo.Compiler.CSharp/MethodConvert/Helpers/CallHelpers.cs
Hecate2: > what i mean is why these checking, like why one count, what if there are `nop, ret`, why not `extern`

Currently simply checking RET is enough for optimization. For complex cases, read-write analysis is required. `extern` handles special classes from framework, like ByteString construction.

shargon (approved):

PR #1261, opened 2024-12-11
remove artifact opcodes line number

Jim8y:
If you take a close look at other prs, https://github.com/neo-project/neo-devpack-dotnet/pull/1260/files for example, you will find that the git log shows large block of opcodes and actually only very few is really updated, making it hard for us to trace the actual change to the script. The reason of this is because once a single line number is changed in the script, all other opcodes have to be updated accordingly.

Thus, in this pr, i remove the line number from the opcode script, making it abvious and clear where the change to the script is made, omiting the influence of the line number.


Hecate2:
I have no problem in any style. But this PR may cause many conflicts.

Jim8y:
> I have no problem in any style. But this PR may cause many conflicts.

i know, dont worry, i will handle those conflicts.

Hecate2:
Fix the artifacts in Neo.SmartContract.Template?

Hecate2 (approved):

shargon (approved):

PR #1265, opened 2024-12-12
accelerate tests

Hecate2:
Tests that are usually not related to core compiler code changes are grouped in namespace Peripheral, in order to accelerate the core tests. Description in coverage is refactored to avoid the slow execution of `instruction.Operand.Span.TryGetString(out var str)`


Hecate2:
Tests that are usually not related to core compiler code changes are grouped in namespace Peripheral, in order to accelerate the core tests. Description in coverage is refactored to avoid the slow execution of `instruction.Operand.Span.TryGetString(out var str)`

src/Neo.SmartContract.Testing/Coverage/CoverageHit.cs
```diff
@@ -24,7 +26,16 @@ public class CoverageHit(int offset, string description, bool outOfScript = fals
         /// <summary>
         /// The instruction description
         /// </summary>
-        public string Description { get; } = description;
+        private string _description { get; } = description;
+        public string Description
```
shargon, line 30:
```suggestion
        private string _description { get; } = description;

        public string Description
```

shargon (commented):
We lose information in `DescriptionFromInstruction`, maybe it's better to avoid this call during test?

Hecate2:
> We lose information in `DescriptionFromInstruction`, maybe it's better to avoid this call during test?

My changes are exactly doing so

shargon:
I mean the method is public and return less information

Hecate2:
Controlling decoding with `bool tryDecodeOperand`

src/Neo.SmartContract.Testing/Coverage/CoverageHit.cs
```diff
@@ -24,7 +31,15 @@ public class CoverageHit(int offset, string description, bool outOfScript = fals
         /// <summary>
         /// The instruction description
         /// </summary>
-        public string Description { get; } = description;
+        public string Description
```
shargon, line 34: Now we don't need this changes? Isn't it?

Hecate2: > Now we don't need this changes? Isn't it?

I am not sure. `[DebuggerDisplay("Offset:{Offset}, Description:{Description}...` still needs additional description with UTF-8 decoded string

src/Neo.SmartContract.Testing/Coverage/CoveredContract.cs
```diff
@@ -86,7 +86,7 @@ internal void GenerateMethods(MethodDetectionMechanism mechanism, ContractState
 
                 if (!_lines.ContainsKey(ip))
                 {
-                    AddLine(instruction, new CoverageHit(ip, CoverageHit.DescriptionFromInstruction(instruction, state.Nef.Tokens), false));
+                    AddLine(instruction, new CoverageHit(ip, CoverageHit.DescriptionFromInstruction(instruction, tryDecodeOperand: false, state.Nef.Tokens), instruction, false));
```
shargon, line 89: But we are removing the info from coverage, not only for test, isn't it?

Hecate2: Again I am not sure😰

Hecate2: > But we are removing the info from coverage, not only for test, isn't it?

What do we actually want in nef coverage report and in tests? I do not find the operand decoding useful, except in debugging

Hecate2:
Github actions are having weird problems at checkout 😰

Jim8y (dismissed):
I am not familier with the coverage hit thing, others look good to me.

Hecate2:
This version should be good enough

src/Neo.SmartContract.Testing/Coverage/CoverageHit.cs
```diff
@@ -24,7 +30,12 @@ public class CoverageHit(int offset, string description, bool outOfScript = fals
         /// <summary>
         /// The instruction description
         /// </summary>
-        public string Description { get; } = description;
+        public string Description { get => DescriptionFromInstruction(Instruction, MethodTokens); }
```
shargon, line 33:
```suggestion
        public string Description => DescriptionFromInstruction(Instruction, MethodTokens);
```

src/Neo.SmartContract.Testing/Coverage/CoveredContract.cs
```diff
@@ -86,7 +89,7 @@ internal void GenerateMethods(MethodDetectionMechanism mechanism, ContractState
 
                 if (!_lines.ContainsKey(ip))
                 {
-                    AddLine(instruction, new CoverageHit(ip, CoverageHit.DescriptionFromInstruction(instruction, state.Nef.Tokens), false));
+                    AddLine(instruction, new CoverageHit(ip, instruction, false, instruction.OpCode == OpCode.CALLT ? state.Nef.Tokens : null));
```
shargon, line 92: seems that if we send always `Tokens` it will be the same, isn't it?

shargon (approved):

src/Neo.SmartContract.Testing/Coverage/CoveredContract.cs
Hecate2: Yes I keep sending the same tokens

PR #1266, opened 2024-12-13
Disable/Fix some tests warnings

nan01ab:
Fix some  noisy  compiler warnings in /tests. 
Generally just disable these warnings one by one(try not to affect any warnings that may be added and not to affect existing test cases)


tests/Neo.SmartContract.Framework.TestContracts/Contract_Storage.cs
```diff
@@ -65,6 +67,7 @@ public static byte[] TestGetString(byte[] key)
             return (byte[])value;
         }
 
+
```
shargon, line 70:
```suggestion
```

shargon (approved):

Jim8y (approved):

PR #1267, opened 2024-12-16
optimize construtor when initializer exists

Hecate2:
```csharp
new MyClass(A=a, B=b, ...) { PropA = propA, PropB = propB, ... }
```


Jim8y (approved):

Hecate2:
1. If the constructor is not complex (INITSLOT+RET or just RET), continue the optimization. Otherwise give up. This PR is simply adding the case of INITSLOT+RET.
2. Analyze the fields written by the initializer expression, and directly PACK them. (no change in this PR)

shargon (approved):

PR #1272, opened 2025-01-03
test misc expression

Hecate2:
```csharp
int a = 0;
a = a++;  // typically it should be just a++;
a = ++a;
```
The behavior of expressions above are not clearly defined (both in C# and in our devpack), but not a big problem for now.


Jim8y:
This pr is not related to a=a++;?

Hecate2:
> This pr is not related to a=a++;?

I did not change any behavior of the compiler, but I wrote tests for a=a++;

shargon (approved):

nan01ab (approved):

vncoelho (approved):

Issue #1274, opened 2025-01-09
`SyntaxTree is not part of the compilation` exception in compiling assignment on static fields of non-smart-contract class


Hecate2:
Based on test contract `neo-devpack-dotnet/tests/Neo.Compiler.CSharp.TestContracts/Contract_MemberAccess.cs`:
```csharp
using Neo.SmartContract.Framework;
using Neo.SmartContract.Framework.Services;

namespace Neo.Compiler.CSharp.TestContracts
{
    public class Contract_MemberAccess : SmartContract.Framework.SmartContract
    {
        public static void TestMain()
        {
            var my = new MyClass();
            MyClass.Data3 = 1;
            ExecutionEngine.Assert(MyClass.Data3 == 1);
        }

        public class MyClass
        {
            public static int Data3 = 3;
        }
    }
}
```
Assignment on static `Data3` leads to exception at compile time:
```
System.ArgumentException
  HResult=0x80070057
  Message=SyntaxTree is not part of the compilation (Parameter 'syntaxTree')
  Source=Microsoft.CodeAnalysis.CSharp
  StackTrace:
   at Microsoft.CodeAnalysis.CSharp.CSharpCompilation.GetSemanticModel(SyntaxTree syntaxTree, SemanticModelOptions options)
```

Hecate2:
`syntaxNode` is of type `VariableDeclaratorSyntax`. This may be the cause.

Hecate2:
Also assignment on static IPrpertySymbol does not work:
```csharp
using Neo.SmartContract.Framework;
using Neo.SmartContract.Framework.Services;

namespace Neo.Compiler.CSharp.TestContracts
{
    public class Contract_MemberAccess : SmartContract.Framework.SmartContract
    {
        public static void TestMain()
        {
            var my = new MyClass();
            ExecutionEngine.Assert(MyClass.Data5 == 5);
            MyClass.Data5 = 6;
            ExecutionEngine.Assert(MyClass.Data5 == 6);
        }

        public class MyClass
        {
            public static int Data5 { get; set; } = 5;
        }
    }
}

```

Jim8y:
are you working to solve the issue?

Hecate2:
> are you working to solve the issue?

I'm on it, but I have no solution

Jim8y:
i will take a look as well then.

Hecate2:
[...]

PR #1276, opened 2025-01-09
Fix static in non contract

Hecate2:
Close https://github.com/neo-project/neo-devpack-dotnet/issues/1274


Hecate2:
Having trouble displaying coveralls percentage in GitHub. The coverage is below
https://coveralls.io/jobs/158313019

shargon (approved):

PR #1281, opened 2025-01-19
[`Fix`]: `checked(-x)` if x is int or long

### LINKED ISSUE #1175: `checked` does not handle the `-x` operation.
In `inc` operation, Neo-VM script and standard C# program will throw an Exception both if `src` overflowed:
```
public static int Inc(int src)
{
     return checked(++src);
}
```

In `-x` operation,  standard C# program will throw an Exception if `src` overflowed(`src` is `-2147483648`), but Neo-VM script not:
```
public static int Negative(int src)
{
     return checked(-src);
}
```

nan01ab:
Fix #1175:

`checked(-int.MinValue)` -> throw exception.
`unchecked(-int.MinValue)` -> int.MinValue(the same value).

as well as `long`.


shargon (approved):

Hecate2 (approved):
It is OK to check `(typeSymbol.Name != "Int32" && typeSymbol.Name != "Int64")` before handling nullable type with `while (typeSymbol.NullableAnnotation == NullableAnnotation.Annotated)`, because the `.Name` is always `Int32`, no matter whether it is nullable.

Jim8y (approved):

Jim8y:
I hate nullable in contract, if we remove nullable support in contract, it will be much easier.

PR #1286, opened 2025-01-24
Optimizer uses INC DEC instead of PUSH1 ADD and PUSH1 SUB

Hecate2:


tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Lambda.cs
```diff
@@ -315,11 +315,11 @@ public abstract class Contract_Lambda(Neo.SmartContract.Testing.SmartContractIni
     /// Unsafe method
     /// </summary>
     /// <remarks>
-    /// Script: VwACeXgKTQMAADZA
+    /// Script: VwACeXgKTAMAADZA
     /// INITSLOT 0002 [64 datoshi]
     /// LDARG1 [2 datoshi]
     /// LDARG0 [2 datoshi]
-    /// PUSHA 4D030000 [4 datoshi]
+    /// PUSHA 4C030000 [4 datoshi]
```
shargon, line 322: Why pusha changed?

Hecate2: The target address changed. The new PUSHA targets address 913, while the old one targets 914. This is because PUSH1 ADD is replaced by INC, saving 1 byte of instruction.

shargon (approved):
Good improve!

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Lambda.cs
Hecate2: > Why pusha changed?

In this contract there is a replacement from PUSH1 SUB to DEC, near address 230, but it is not represented in this artifact file difference, because the replacement happens in a private method.

PR #1287, opened 2025-01-27
[`Fix`]: check division overflow for int32 and int64

nan01ab:
Division overflow if `int.MinValue / -1` and `long.MinValue /  -1 `.

In C#:
> In a checked context, this causes a System.ArithmeticException (or a subclass thereof) to be thrown. In an unchecked context, it is implementation-defined as to whether a System.ArithmeticException (or a subclass thereof) is thrown or the overflow goes unreported with the resulting value being that of the left operand.

Reference: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12103-division-operator

There is no division overflow check in `unchecked` statement.
But division overflow is checked in current .Net implementation.

Division overflow may cause the result to exceed the range of int32 or int64(in unchecked statement).
This issue will be fixed later.

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
@@ -11,6 +11,7 @@
 
 extern alias scfx;
 
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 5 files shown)


src/Neo.Compiler.CSharp/MethodConvert/Expression/UnaryExpression.PrefixUnary.cs
```diff
@@ -279,19 +279,20 @@ private void EmitIncrementOrDecrement(SyntaxToken operatorToken, ITypeSymbol? ty
...
+        if (typeSymbol is null) return;
         while (typeSymbol.NullableAnnotation == NullableAnnotation.Annotated)
         {
             // Supporting nullable integer like `byte?`
             typeSymbol = ((INamedTypeSymbol)typeSymbol).TypeArguments.First();
         }
 
+        if (typeSymbol.Name != "Int32" && typeSymbol.Name != "Int64")
+        {
+            //  -sbyte, -byte, -short, -ushort, -char -> int, -int, -uint -> long
```
shargon, line 291: unsigned types works with negate?

Jim8y: He is right, - ushort => int

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -99,6 +106,39 @@ private void ConvertBinaryExpression(SemanticModel model, BinaryExpressionSyntax
...
+
+        AddInstruction(OpCode.DUP);
+        Push(-1);
+        Jump(OpCode.JMPNE_L, endTarget);
+
+        AddInstruction(OpCode.OVER);
+        Push(minValue);
+        Jump(OpCode.JMPNE_L, endTarget);
+
+        AddInstruction(OpCode.THROW);
```
Jim8y, line 138: here you can push an exception message before throw, but not necesssary.

nan01ab: > here you can push an exception message before throw, but not necesssary.

No exception message in other similar `THROW`s.
So no exception message here.

Jim8y: > > here you can push an exception message before throw, but not necesssary.
> 
> No exception message before other similar `THROW`s. So no exception message here.

LOL, that is actually a problem in other places, having exception message is the correct behavior, but its ok, no need to update. .

nan01ab: > > > here you can push an exception message before throw, but not necesssary.
> > 
> > 
> > No exception message before other similar `THROW`s. So no exception message here.
> 
> LOL, that is actually a problem in other places, having exception message is the correct behavior, but its ok, no need to update. .

This and others will be fixed by later PRs.

shargon: We should fix it in a different PR

nan01ab: > > > here you can push an exception message before throw, but not necesssary.
> > 
> > 
> > No exception message before other similar `THROW`s. So no exception message here.
> 
> LOL, that is actually a problem in other places, having exception message is the correct behavior, but its ok, no need to update. .

Why LOL?
Many cases were written by you.

Jim8y (approved):

shargon (approved):

PR #1290, opened 2025-02-05
test catching a class of exception; Should catch not all exceptions, but only the required class of exception?

Hecate2:
I am confused whether this is a bug or a feature:
```csharp
namespace Neo.Compiler.CSharp.TestContracts
{
    public class Contract_TryCatch : SmartContract.Framework.SmartContract
    {
        public static string catchExceptionType()
        {
            string result = "NoException";
            try { throw new System.Exception(); }
            catch (UncatchableException e) { result += e; }
            return result;
        }
    }
}
```
```python
# Method Start Neo.Compiler.CSharp.TestContracts.Contract_TryCatch.catchExceptionType()
00 INITSLOT 02-00 # 2 local variables, 0 arguments
# Code test-compiler.cs line 21: ""NoException""
03 PUSHDATA1 4E-6F-45-78-63-65-70-74-69-6F-6E # as text: "NoException"
# Code test-compiler.cs line 21: "result = "NoException""
16 STLOC0
17 TRY 0F-00 # catch pos: 32 (offset: 15), no finally block
# Code test-compiler.cs line 22: "throw new System.Exception();"
20 PUSHDATA1 65-78-63-65-70-74-69-6F-6E # as text: "exception"
# Code test-compiler.cs line 22: "throw new System.Exception();"
31 THROW
# Code test-compiler.cs line 23: "catch"
32 STLOC1
# Code test-compiler.cs line 23: "result += e"
33 LDLOC0
# Code test-compiler.cs line 23: "e"
34 LDLOC1
# Code test-compiler.cs line 23: "result += e"
35 CAT
# Code test-compiler.cs line 23: "result += e"
# Code test-compiler.cs line 23: "result += e"
36 CONVERT 28 # ByteString type
# Code test-compiler.cs line 23: "result += e"
# Code test-compiler.cs line 23: "result += e"
38 STLOC0
39 ENDTRY 02 # pos: 41 (offset: 2)
# Code test-compiler.cs line 24: "result"
# Code test-compiler.cs line 24: "return result;"
41 LDLOC0
# Method End Neo.Compiler.CSharp.TestContracts.Contract_TryCatch.catchExceptionType()
# Code test-compiler.cs line 25: "}"
42 RET
```
It catches any exception and returns `NoExceptionexception`. Maybe we need to judge the exeption type, or simply allow `catch (System.Exception e)` only.


shargon:
`UncatchableException` can't be catched

Hecate2:
> `UncatchableException` can't be catched

When I write `catch (UncatchableException e)` but throw a `System.Exception`, I think the exception should not be caught. But the `System.Exception` is actually caught.

shargon:
We don't have exceptions kinds in the vm... @Jim8y proposed in the past

Jim8y (approved):

nan01ab (approved):

shargon (commented):
LgTM too

PR #1291, opened 2025-02-06
analyze write in nested try

Hecate2:
```python
try:
  try:
    Write()
  finally:
    ...
except:
  ...
```
This is also considered unsafe, because the `Write()` in the internal try may throw and be caught by the external try. Thus it is ambiguous whether the internal `Write()` is reverted.


Hecate2:
It is not checking writes in a catch nested in a try. Fixing.

shargon (approved):

PR #1293, opened 2025-02-08
dumpnef intakes manifest to find method starts

Hecate2:
Useful for deployed contracts without debugInfo

### DIFF
--- a/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
@@ -300,7 +300,7 @@ namespace Neo.Optimizer
                 }
             }
 
-            if (debugInfo == null && manifest != null)
+            if (manifest != null)
                 foreach (ContractMethodDescriptor method in manifest.Abi.Methods)
                     methodStartAddrToName[method.Offset] = method.Name;
 
--- a/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
@@ -11,6 +11,7 @@
 
 using Neo.Json;
 using Neo.SmartContract;
+using Neo.SmartContract.Manifest;
 using Neo.SmartContract.Native;
 using Neo.VM;
 using System;
@@ -253,7 +254,14 @@ namespace Neo.Optimizer
                 ).Select(ai => ai.a).ToList();
         }
 
-        public static string GenerateDumpNef(NefFile nef, JToken? debugInfo)
+        /// <summary>
+        /// Dumps .nef assembly to a text file
+        /// </summary>
+        /// <param name="nef">The NEF file.</param>
+        /// <param name="debugInfo">The debug information. If available, provides source code comments in the dumped assembly.</param>
+        /// <param name="manifest">The contract manifest. Provides method start comments. Not necessary if debugInfo is available</param>
+        /// <returns></returns>
+        public static string GenerateDumpNef(NefFile nef, JToken? debugInfo, ContractManifest? manifest = null)
         {
             Script script = nef.Script;
             string addressPadding = script.GetInstructionAddressPadding();
@@ -291,6 +299,10 @@ namespace Neo.Optimizer
                     }
                 }
             }
+
+            if (debugInfo == null && manifest != null)
+                foreach (ContractMethodDescriptor method in manifest.Abi.Methods)
+                    methodStartAddrToName[method.Offset] = method.Name;
 
             Dictionary<string, string[]> docPathToContent = new();
             StringBuilder dumpnef = new();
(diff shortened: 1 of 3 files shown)


src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
```diff
@@ -292,6 +300,10 @@ public static string GenerateDumpNef(NefFile nef, JToken? debugInfo)
                 }
             }
 
+            if (debugInfo == null && manifest != null)
```
shargon, line 303:
```suggestion
            if (manifest != null)
```
Manifest is more trustable, isn't it?

Hecate2: > Manifest is more trustable, isn't it?

Yes it is.

Jim8y (approved):

src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
```diff
@@ -253,7 +254,14 @@ public static List<int> OpCodeAddressesInMethod(NefFile nef, JToken DebugInfo, s
...
 
-        public static string GenerateDumpNef(NefFile nef, JToken? debugInfo)
+        /// <summary>
+        /// Dumps .nef assembly to a text file
+        /// </summary>
+        /// <param name="nef">The NEF file.</param>
+        /// <param name="debugInfo">The debug information. If available, provides source code comments in the dumped assembly.</param>
+        /// <param name="manifest">The contract manifest. Provides method start comments. Not necessary if debugInfo is available</param>
+        /// <returns></returns>
+        public static string GenerateDumpNef(NefFile nef, JToken? debugInfo, ContractManifest? manifest = null)
```
shargon, line 264: We should have a unit test for this method, I don't know well the final output, because maybe we need to makes different between if the method comes from manifest or from debugInfo

Hecate2: > We should have a unit test for this method, I don't know well the final output, because maybe we need to makes different between if the method comes from manifest or from debugInfo

`nccs YourContract.csproj --assembly` or `nccs YourContract.nef --assembly` generates the dumpnef text file. The content is just assembly instructions with comments. The method name from debugInfo is much longer and more accurate, because it contains the namespace and class names from the source code. For example, `# Method Start Neo.Compiler.CSharp.TestContracts.Contract_WriteInTry.Write()`. Method names from manifest is much more concise, like `# Method Start totalSupply`

shargon: Maybe we add a new entry for the manifest?
# Public Method: xxxx

Hecate2: > Maybe we add a new entry for the manifest? # Public Method: xxxx

I am still thinking about this. Do we need to include both `# Method Start` and `# Public Method`, or only one of them? 
I think my original codes are the most reliable😂😭. People dumpnef because they need an assembly document commented well by source codes, and the debugInfo has to be correct for this. If the debugInfo is wrong, a user can try again without the debugInfo.
Also there exists a compilation pattern moving the public method entry point to somewhere else, in order to initialize the non-static variables in the contract class. With this pattern the manifest is less reliable than the debugInfo, because the user has more difficulty finding the actual method implementation.

Hecate2: > Maybe we add a new entry for the manifest? # Public Method: xxxx

I think it is good if we write entry address keys from manifest, if it is not written by the debugInfo.

src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
```diff
@@ -292,6 +300,10 @@ public static string GenerateDumpNef(NefFile nef, JToken? debugInfo)
                 }
             }
 
+            if (manifest != null)
+                foreach (ContractMethodDescriptor method in manifest.Abi.Methods)
+                    methodStartAddrToName[method.Offset] = method.Name;
```
Hecate2, line 305:
```suggestion
                    methodStartAddrToName.TryAdd(method.Offset, method.Name);
```

nan01ab (approved):

src/Neo.Compiler.CSharp/Optimizer/DumpNef.cs
shargon: @Hecate2 make the changes and I will approve it, but it would be nice to have the expected result somewhere

Hecate2: Will try to have some nice tests.

Hecate2:
Still struggling to balance the restriction and correctness of tests

Hecate2:
Tests ready. Difficult to include more strict tests.

shargon (approved):

PR #1303, opened 2025-03-21
fix examples project issues and upgrade dependency

Jim8y:
This pr made the following changes:

1. update the example project dependency
2. fix the compiling errors of those examples to match latest devpack setting.
3. fix compling errors in the compiler that makes the examples unable to compile.
4. add readme files to exsisting examples.


src/Neo.Compiler.CSharp/MethodConvert/Expression/CastExpression.cs
```diff
@@ -52,7 +52,19 @@ private void ConvertCastExpression(SemanticModel model, CastExpressionSyntax exp
             CallMethodWithInstanceExpression(model, method, null, expression.Expression);
             return;
         }
+
+        // Handle null literal expressions: (T)null
```
Jim8y, line 56: examples faild for parsing `(ECPoint)null` with exception: reference to a null object.

src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
```diff
@@ -46,20 +47,28 @@ private void ConvertExpression(SemanticModel model, ExpressionSyntax syntax, Syn
 
     private bool TryConvertConstant(SemanticModel model, ExpressionSyntax syntax, SyntaxNode? syntaxNode)
     {
-        var constant = model.GetConstantValue(syntax);
-        var value = constant.Value;
-        if (value == null)
-            return false;
+        try
+        {
+            var constant = model.GetConstantValue(syntax);
```
Jim8y, line 52: exception of `model not exsit in the currrent syntax tree` thrown

Wi1l-B0t:
This PR has several different changes.

Why are these changes submitted in one PR.

Jim8y:
> This PR has several different changes.
> 
> Why are these changes submitted in one PR.

the purpose of this pr is to make examples to work, without these changes, can't make it work~~


and my original intention was to update examples only, then i discovered those issues.

And this is for examples, not worth too much efforts. but can break into smaller prs if its considered a `must`

vncoelho (commented):
Looks good, but why the debug change in type?

Also, the fix could come in different pr

Jim8y:
> Looks good, but why the debug change in type?
> 
> Also, the fix could come in different pr

cause type was changed long time ago, we just did not update the examples

shargon (approved):

PR #1316, opened 2025-04-24
Improve contract syntax analyzer

Jim8y:
# Fix CheckWitnessUsageAnalyzer to check if result is used rather than specific patterns

## Changes
- Updated the CheckWitnessUsageAnalyzer to check if the result of Runtime.CheckWitness is being used in any way, rather than requiring specific patterns
- The analyzer now only reports an issue if the CheckWitness call is used directly as a statement without capturing the result
- Updated error messages to be more accurate
- GasOptimizationAnalyzer already skips checking string literals in contract attributes

## Reason for changes
The previous implementation of CheckWitnessUsageAnalyzer was too restrictive, requiring specific patterns like if statements or Assert calls. The updated version recognizes that the result of CheckWitness can be used in many valid ways, such as being assigned to a variable, used in a condition, passed as a parameter, etc.


Wi1l-B0t:
UT failed?

tests/Neo.Compiler.CSharp.TestContracts/Contract_Inline.cs
```diff
@@ -105,3 +111,5 @@ private static int CallMethodThatReturnsInt(int a, int b)
         }
     }
 }
+
```
shargon, line 114: Why all ends in double empty lines?

PR #1330, opened 2025-06-16
SafeAttribute supports in properties

erikzhang:


Jim8y:
should a safe property mean it should not have a set method?

Jim8y (approved):

shargon (approved):

PR #1331, opened 2025-06-16
feat: Add automatic 'using System;'

Jim8y:
## Overview

This PR implements automatic injection of 'using System;' directives into C# source files during compilation when the directive is not already present.

## Problem Solved

Developers often forget to include 'using System;' in their Neo smart contracts, leading to compilation errors when using basic System types like Exception, Convert, Math, etc. This feature eliminates those errors by automatically ensuring the directive is present.

### Files Modified:
- **Helper.cs**: Added 3 new methods for detection and injection logic
  - `EnsureUsingSystemExists()` - Main entry point
  - `HasUsingSystem()` - Intelligent detection of existing using System directives  
  - `InjectUsingSystem()` - Smart injection with proper placement logic

- **CompilationEngine.cs**: Integrated injection at all source processing points
  - Direct source file compilation
  - Project-based compilation  
  - Package source file processing

- **UsingSystemInjectionTests.cs**: Comprehensive test suite with 13 test cases

### Key Features:
- ✅ **Automatic Detection**: Intelligently detects existing using System statements
- ✅ **Smart Placement**: Inserts at the correct location (after comments, before other usings)
- ✅ **Format Support**: Handles 'using System;', 'global using System;', case variations
- ✅ **Backward Compatible**: Existing files with using System unchanged
- ✅ **No Duplicates**: Never creates duplicate using statements

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
+++ b/src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
@@ -116,7 +116,12 @@ namespace Neo.Compiler
 
         public List<CompilationContext> Compile(IEnumerable<string> sourceFiles, IEnumerable<MetadataReference> references)
         {
-            IEnumerable<SyntaxTree> syntaxTrees = sourceFiles.OrderBy(p => p).Select(p => CSharpSyntaxTree.ParseText(File.ReadAllText(p), options: Options.GetParseOptions(), path: p));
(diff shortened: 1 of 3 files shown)


vncoelho (commented):
@Jim8y,

I understand the point, but is it a good practice to keep adding all these automatically?

Jim8y:
> Why not avoid ImplicitUsings in csproj?

Can be a way if it work, go ahead of fixing in another pr.

PR #1334, opened 2025-06-17
Disable implicit usings for smart contract projects 

Jim8y:
Disable implicit usings for smart contract projects - Add ImplicitUsings=disable to examples/Directory.Build.props and all smart contract templates for deterministic compilation


erikzhang:
What's the difference?

shargon (approved):

shargon (approved):

PR #1341, opened 2025-06-24
[`Feature`] Increase/decrease

shargon:
Increase/Decrease is very used in smart contracts


src/Neo.SmartContract.Framework/Services/StorageMap.cs
```diff
@@ -304,6 +304,33 @@ public extern ByteString? this[byte[] key]
             return StdLib.Deserialize(value);
         }
 
+        public BigInteger Increase(byte[] key) => Increase((ByteString)key, BigInteger.One);
+        public BigInteger Increase(ByteString key) => Increase(key, BigInteger.One);
+        public BigInteger Increase(byte[] key, BigInteger amount) => Increase((ByteString)key, amount);
+        public BigInteger Increase(ByteString key, BigInteger amount)
+        {
+            var i = GetIntegerOrZero(key) + amount;
+            Put(key, i);
```
Wi1l-B0t, line 313: -1 + 1 = 0?
Delete key if value is zero? Same as Decrease.

vncoelho: maybe add a flag if that is the case (deleteIfZero)

Wi1l-B0t (approved):

PR #1381, opened 2025-08-19
feat: Add GuardHelpers utility class for smart contract validation

Jim8y:
Introduces a comprehensive validation helper library to standardize precondition and postcondition checks in Neo smart contracts.

Features:
- Basic validation: Require(), Ensure(), Revert()
- Null checks: RequireNotNull()
- Numeric validation: RequireNonNegative(), RequirePositive(), RequireInRange()
- Address validation: RequireValidAddress(), RequireWitness()
- Security checks: RequireCaller(), RequireEquals()
- String validation: RequireNotEmpty()

Benefits:
- Reduces boilerplate validation code
- Standardizes error messages with short codes for gas efficiency
- Improves contract security and reliability
- Makes smart contracts more readable and maintainable

PR #1386, opened 2025-09-08
Change guard helpers to use Abort/Assert

shargon:
Related to https://github.com/neo-project/neo-devpack-dotnet/pull/1381/files#r2293140778


shargon:
I didn't see https://github.com/neo-project/neo-devpack-dotnet/pull/1384/files

PR #1388, opened 2025-09-09
Implement Local Storage

shargon:
Implementation of https://github.com/neo-project/neo/pull/4118

### DIFF
--- a/src/Neo.SmartContract.Framework/Services/Storage.cs
+++ b/src/Neo.SmartContract.Framework/Services/Storage.cs
@@ -34,6 +34,8 @@ namespace Neo.SmartContract.Framework.Services
             get;
         }
 
+        #region Get
+
         /// <summary>
         /// Returns the value corresponding to the given key for Storage context (faster: generates opcode directly)
         /// </summary>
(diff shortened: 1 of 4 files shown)


Jim8y:
its merged already?

shargon:
> its merged already?

no, is waiting

shargon:
Merge?

src/Neo.SmartContract.Framework/Services/Storage.cs
```diff
@@ -76,6 +94,40 @@ public static extern StorageContext CurrentReadOnlyContext
...
+        /// Writes the key/value pair for the current contract (faster: generates opcode directly)
+        /// </summary>
+        [Syscall("System.Storage.Local.Put")]
+        public static extern void Put(ByteString key, ByteString value);
+
+        /// <summary>
+        /// Writes the key/value pair for the current contract (faster: generates opcode directly)
+        /// </summary>
+        [Syscall("System.Storage.Local.Put")]
+        public static extern void Put(byte[] key, ByteString value);
```
vncoelho, line 107: @shargon, in the future, would it be better if we do something like?
```cs
public static void PutKey<T>(T key, ByteString value)
{
    switch (key)
    {
        case ByteString bs:
            Put(bs, value);
            break;
        case byte[] ba:
            Put(ba, value);
            break;
        default:
            xxxxxxxxxxx
    }
}
``

vncoelho (approved):
It looks good to me

src/Neo.SmartContract.Framework/Services/Storage.cs
shargon: it cost more gas because the type check

PR #1392, opened 2025-09-22
Guard oracle template callback against unauthorized invocations

Jim8y:
## Summary
- ensure the oracle template verifies the native oracle caller before accepting responses
- avoid null invocation of the response event when no subscribers are registered

## Testing
- not run (not requested)


tests/Neo.SmartContract.Template.UnitTests/templates/neocontractoracle/OracleRequestTests.cs
```diff
@@ -68,7 +72,7 @@ public void TestDoRequestAndFinish()
             Engine.Transaction.Attributes = new[]{
                 new OracleResponse()
                 {
-                     Code = OracleResponseCode.Error,
+                     Code = (Neo.Network.P2P.Payloads.OracleResponseCode)OracleResponseCode.Error,
```
Wi1l-B0t, line 75: Why not `OracleResponseCode`?

shargon (approved):

PR #1397, opened 2025-10-19
Optimize: use explicit foreach DesignationSyntax instead to avoid warning

Wi1l-B0t:
Use explicit foreach DesignationSyntax instead to avoid warning.

Now, there is only one warning when `dotnet build`.


src/Neo.Compiler.CSharp/MethodConvert/Statement/CommonForEachStatement.cs
```diff
@@ -178,7 +185,7 @@ private void ConvertIteratorForEachVariableStatement(SemanticModel model, ForEac
                     }
                     else
                     {
-                        byte variableIndex = AddLocalVariable(symbols[i]);
+                        byte variableIndex = AddLocalVariable(symbols[i]!);
```
shargon: but symbols[i] now can be null

shargon (approved):

PR #1401, opened 2025-10-22
Simplify system method control flow with DSL

Jim8y:
## Summary
- extend the control-flow helper DSL with conditional jump wrappers and loop scopes
- refactor system method implementations to use DSL guards instead of manual jump wiring
- ensure opcode order is preserved for parse/create, nullable, enum, string, math, and numeric handlers

## Testing
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj -c Release --filter "FullyQualifiedName~WriteInTry|Name~Parse|Name~CreateChecked"
- dotnet test neo-devpack-dotnet.sln -c Release

### DIFF
--- a/docs/CSharpSyntaxCheckList.md
+++ b/docs/CSharpSyntaxCheckList.md
@@ -408,7 +408,7 @@ C# Language Specification :  https://ecma-international.org/publications-and-sta
 - [x] `ushort`: Represents a 16-bit unsigned integer. Example: `ushort smallPositiveValue = 65535;`
 - [x] `uint`: Represents a 32-bit unsigned integer. Example: `uint positiveValue = 123;`
 - [x] `ulong`: Represents a 64-bit unsigned integer. Example: `ulong bigPositiveValue = 1234567890UL;`
-- [ ] (Will Support) `enum`: Represents an enumeration. Example: `enum Days { Sunday, Monday, Tuesday, Wednesday, Thursday, Friday, Saturday };`
+- [x] `enum`: Represents an enumeration. Example: `enum Days { Sunday, Monday, Tuesday, Wednesday, Thursday, Friday, Saturday };`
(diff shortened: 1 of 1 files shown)


src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Short.cs
```diff
@@ -132,14 +135,17 @@ private static void HandleShortCreateChecked(MethodConvert methodConvert, Semant
...
-        endTarget.Instruction = methodConvert.Nop();               // End target
+        methodConvert.EmitIf(
+            () =>
+            {
+                methodConvert.Dup();
+                methodConvert.Within(short.MinValue, short.MaxValue); // Check if value is within short range
+            },
+            () => { },
+            () => methodConvert.Throw(),
+            fallThroughElse: true);
```
shargon: seems that almost always it's true

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ControlFlowDsl.cs
```diff
@@ -0,0 +1,345 @@
...
+    private void EmitSwitch(Action valueEmitter, IReadOnlyList<(BigInteger value, Action body)> cases, Action? defaultBody = null)
+    {
+        ArgumentNullException.ThrowIfNull(valueEmitter);
+        ArgumentNullException.ThrowIfNull(cases);
+
+        valueEmitter();
+        if (cases.Count == 0 && defaultBody is null)
+        {
+            Drop();
+            return;
```
shargon: does we need to emit first?

Jim8y:
> Code looks good to me, but we should ensure that all the changes are covered by UT because it's easy to have a mistake with all of these changes

As you can see no artifacts or ut changes. but if you think more ut is required, i will add them

PR #1402, opened 2025-10-23
Expose policy governance APIs

Jim8y:
## Summary
- expose missing Policy governance setters/getters and role designation helpers to smart contracts
- add managed transaction attribute wrappers so contracts can inspect native attributes
- update testing harness artifacts and unit tests to cover the new governance APIs

## Testing
- dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj --filter NativeTest


src/Neo.Compiler.CSharp/SecurityAnalyzer/ReEntrancyAnalyzer.cs
```diff
@@ -243,17 +242,45 @@ private class SourceLocation
...
+            {
+                path = Path.Combine(Environment.CurrentDirectory, path);
+            }
+
+            try
+            {
+                if (path != null && File.Exists(path))
+                {
+                    string[] lines = File.ReadAllLines(path);
+                    foreach (int candidate in new[] { zeroBasedLineNumber, zeroBasedLineNumber - 1 })
```
shargon: Why we need to check two lines?

erikzhang:
I don't see any point in integrating these APIs into the SDK. No contract will need to call these APIs.

PR #1414, opened 2025-10-27
Enhance string Split/Trim system call support

Jim8y:
## Summary
- extend compiler string.Split support (char/string and single-element array separators)
- add trim handlers for char array overloads and update analyzer/docs
- expand contract/tests and regenerate artifacts, keeping syntax checks aligned

## Testing
- dotnet test tests/Neo.Compiler.CSharp.UnitTests --filter UnitTest_String


src/Neo.SmartContract.Analyzer/StringMethodUsageAnalyzer.cs
```diff
@@ -26,14 +26,12 @@ public class StringMethodUsageAnalyzer : DiagnosticAnalyzer
         // Add string method names to this array as needed
         private readonly string[] _unsupportedStringMethods =
         {
-            "Clone", "Compare", "CompareOrdinal", "CompareTo",
+            "Clone", "CompareOrdinal", "CompareTo",
             "Copy", "CopyTo", "Format", "GetEnumerator",
             "GetHashCode", "GetType", "GetTypeCode",
-            "IndexOfAny", "Insert", "Intern", "IsInterned",
```
shargon: `Insert` and others that was removed are already done?

PR #1415, opened 2025-10-27
Fix forward-referenced contract dependencies

Jim8y:
- fix the compiler dropping forward-referenced smart contracts by collecting
    all smart contract symbols before building the dependency graph
  - rebuild the dependency list using symbol equality so members referencing
    contracts declared later are preserved and _nonDependencies cannot trim
    required types
  - guard against duplicate and abstract dependency entries while keeping the
    rest of the compilation flow unchanged

  Issue:
  CompilationEngine only added a smart contract to classDependencies after it
  had already been visited, so a field or property that pointed to a contract
  declared later looked like a plain class reference. When _nonDependencies
  ran, those untracked contracts were considered foreign and stripped from the
  compilation context, leading to partial NEF output in multi-contract projects.

  Fix Details:
  The change executes two passes over the syntax trees:

  1. Record every public, non-abstract type that derives from SmartContract.
  2. For each of those contracts, walk its members and add any referenced smart
     contract symbol to the dependency list, deduplicating and skipping abstract
     types.

  This ensures the subsequent topological sort sees the full dependency graph
  regardless of declaration order, so every required contract is compiled and
  emitted.


Jim8y:
> Any test for this ? How to test this ?

this is hard to be tested.... cuase it will be related to multiple contract situations. This is not a serious issue as well.

Jim8y (commented):
Thanks for the patch – the fix in `MethodConvert.cs` looks reasonable, but Shargon’s earlier question (“Any test for this? How to test this?”) is still outstanding.

Because this change affects contract dependency resolution, we need at least one regression test so we don’t break this path again. Could you add a unit/integration test (even a small multi-contract scenario) that proves the new flow actually compiles forward-dependent contracts? Without that we can’t safely merge.

PR #1445, opened 2025-12-24
Add: System.Runtime.MintGas syscall for GAS-only contract fees

Jim8y:
## Summary

Implements the NEP for GAS-only contract fees via system fee minting. This is a minimal protocol extension that supports contract fees without modifying the ABI.

### New Syscall

```csharp
void System.Runtime.MintGas(BigInteger amount)
```

**Semantics:**
- `amount` must be non-negative
- If `SystemFee + amount` exceeds `MaxSystemFee`, VM will FAULT
- On success, `SystemFee` increases by `amount` and equivalent GAS is credited to `Runtime.ExecutingScriptHash`
- Can only be used in Application trigger

### Usage Example

```csharp
public class MyContract : SmartContract
{
    // Fixed fee of 1 GAS
    public static void PaidService()
    {
        Runtime.MintGas(100000000); // 1 GAS
        // ... service logic
    }

    // Dynamic fee based on data size
    public static void StoreData(byte[] data)
    {
        BigInteger fee = data.Length * 1000; // 0.00001 GAS per byte
        Runtime.MintGas(fee);
        // ... storage logic
    }
}
```

### Design Rationale

- No ABI/manifest changes required
- Fee logic stays inside contract code
- Authorization is clearly scoped at transaction level via `MaxSystemFee`
- Simpler than attribute-based approach

### Dependencies

Requires neo-vm implementation of `System.Runtime.MintGas` syscall.

## Test Plan

- [x] Test contract demonstrating MintGas usage patterns
- [ ] Integration tests (pending neo-vm implementation)


Wi1l-B0t:
UT ?

Wi1l-B0t:
This is a part of NEO4 ?

Jim8y:
this will not be considered

PR #1466, opened 2026-01-20
Inline NuGet release steps in workflow

Jim8y:
## Summary
- inline ordered NuGet publishing in the Release workflow (no external script)
- keep dependency order, package availability polling, and Testing package swap logic
- restore release solution handling and cleanup
- fix Neo package version detection in Release job (VersionPrefix/Version)

## Testing
- not run (workflow change only)


Jim8y:
too old and not appliable

PR #1486, opened 2026-01-27
fix(compiler): add left shift overflow check in checked context

Jim8y:
## Summary
Add runtime validation for left shift operations in checked context to prevent unexpectedly large values.

## Changes
- Add CheckShiftOverflow method to validate shift amount
- Runtime check throws if shift amount exceeds type bit width
- Supports all integer types (sbyte, byte, short, ushort, int, uint, long, ulong)

## Security Audit
Fixes HIGH-001 from compiler security audit

## Testing
- Build verified successfully

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
@@ -139,6 +139,54 @@ internal partial class MethodConvert
         endTarget.Instruction = AddInstruction(OpCode.NOP);
     }
 
+    /// <summary>
+    /// Checks for left shift overflow in checked context.
+    /// Validates that the shift amount is non-negative and within the bit width of the left operand type.
+    /// </summary>
+    /// <param name="model">The semantic model.</param>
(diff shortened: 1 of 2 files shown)


shargon (approved):
Please, port to N4

Wi1l-B0t (approved):

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -137,35 +124,75 @@ private void CheckDivideOverflow(ITypeSymbol? type)
...
+        if (leftType is null) return;
+
+        while (leftType.NullableAnnotation == NullableAnnotation.Annotated)
+        {
+            leftType = ((INamedTypeSymbol)leftType).TypeArguments.First();
+        }
+
+        // Determine the bit width based on the type
+        // Note: In NEO, BigInteger is Int256 (256-bit integer)
+        var maxShift = leftType.Name switch
```
Wi1l-B0t, line 168: It's different from C-sharp, why add this?

PR #1509, opened 2026-02-07
Fix: no contract method `GetExecPicoFeeFactor` in N4

Wi1l-B0t:
The `PolicyContract` doesn't have contract method `GetExecPicoFeeFactor` in N4.

So adding an implementation rather than definition.


Jim8y (dismissed):

shargon (commented):
Why don't remove it?

shargon (approved):

PR #1510, opened 2026-02-07
fix(compiler): evaluate switch cases before default fallback

Jim8y:
## Summary
- fix switch statement lowering so `default` no longer short-circuits case checks when declared before `case` labels
- keep `default` as the fallback jump emitted after all case/pattern checks
- add a regression unit test that inspects generated assembly for `switch` with `default` first

## Testing
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_SwitchStatementSemantics" -v minimal`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Switch" -v minimal`


shargon (approved):

Wi1l-B0t:
Port to N4?

PR #1511, opened 2026-02-07
fix(compiler): correct recursive pattern multi-property semantics

Jim8y:
## Summary
- fix recursive property-pattern lowering to evaluate each subpattern against the original matched value
- add proper short-circuit AND behavior across multiple subpatterns
- add guards for missing property names and getter-less properties in recursive patterns
- add a regression case in `Contract_Pattern`/`UnitTest_Pattern` to verify multi-property recursive pattern semantics

## Root cause
The recursive pattern converter loaded the candidate value once before iterating subpatterns. After the first property check, the stack top became the comparison result (`bool`), and subsequent property checks operated on that boolean instead of the original value.

## Validation
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Pattern.RecursivePattern_Test" -v minimal`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Pattern" -v minimal`


src/Neo.Compiler.CSharp/MethodConvert/Pattern/RecursivePattern.cs
```diff
@@ -21,40 +21,69 @@ internal partial class MethodConvert
...
+            {
+                throw CompilationException.UnsupportedSyntax(subpattern, $"Recursive patterns currently only support constant pattern matching. Found: {subpattern.Pattern?.GetType().Name ?? "null"}. Use syntax like '{{ PropertyName: constantValue }}'.");
+            }
+
+            if (hasSubpattern)
+            {
+                JumpTarget evaluateNextPattern = new();
+                JumpTarget endPattern = new();
+                Jump(OpCode.JMPIF_L, evaluateNextPattern);
+                Push(false);
```
Wi1l-B0t, line 42: Why `Push(false)` ?

ajara87: I think it’s because JMPIF_L consumes the boolean on the stack so Push(false) is needed to restore the correct result for the path

Wi1l-B0t:
Conflicts.

shargon (approved):

ajara87 (approved):

Wi1l-B0t (approved):

PR #1512, opened 2026-02-07
fix(compiler): add int support for type patterns

Jim8y:
## Summary
- add missing `int` mapping in `Helper.GetPatternType` so `is int` type patterns compile
- add `UnitTest_PatternTypeSupport.TypePattern_Int_Compiles` regression test

## Root cause
`GetPatternType` used string-based type matching and omitted the `"int"` case, even though the error message listed it as supported.

## Validation
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_PatternTypeSupport.TypePattern_Int_Compiles" -v minimal`

### DIFF
--- a/docs/csharp-syntax/UnsupportedFeatures.md
+++ b/docs/csharp-syntax/UnsupportedFeatures.md
@@ -29,7 +29,6 @@ The versioned syntax checklists flag every feature the Neo compiler currently re
   - Exception filters (`exception_filter`)
 
 - **C# 7 Syntax Checklist**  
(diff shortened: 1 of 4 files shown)


shargon (commented):
Ut fault

shargon (approved):

Wi1l-B0t:
Port to N4?

shargon:
> Port to N4?

I think we should start master n4 from master n3 branch when we have ready the vm, otherwise we will duplicate the work

PR #1513, opened 2026-02-07
fix(compiler): fail fast on local function declarations

Jim8y:
## Summary
- reject `LocalFunctionStatementSyntax` explicitly in statement conversion
- add regression test ensuring local-function declarations fail compilation with an explicit compiler diagnostic

## Root cause
`ConvertStatement` silently ignored local function declarations (`break` branch), which let contracts compile while dropping local-function code unexpectedly.

## Validation
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_LocalFunctionSyntax.LocalFunctionDeclaration_CompilationFails" -v minimal`


Wi1l-B0t (approved):

shargon (approved):

Wi1l-B0t:
Port to N4 ?

PR #1517, opened 2026-02-22
Cherry-Pick: LocalStorageMap (#1506)

Wi1l-B0t:
Cherry-Pick #1506 to N4.


shargon:
Master n4 have a different VM, when we finish it, we can copy master n3 branch and do the changes from n4,  otherwise we will have the same braches, and duplicate the work

Wi1l-B0t:
> Master n4 have a different VM, when we finish it, we can copy master n3 branch and do the changes from n4, otherwise we will have the same braches, and duplicate the work

OK

Issue #1298, opened 2025-03-03
Optimize null checks when use `IsValid`


shargon:
I saw a lot of smart contracts that use this code

```csharp
UInt160 hash = xxxxx;
Assert(hash != null  && hash.IsValid)
```

this should be optimized to remove the null check because is already done in `IsValid`

Jim8y:
from the syntax perspective, if hash is null, it should not be able to call IsValid.

shargon:
The problem is that we already did a null check in IsValid, so or it's redudant, or we optimize it

shargon:
From a c# developer is like @Qiao-Jin said, but I prefer to change to an extension, as @Wi1l-B0t said and deprecate the old one, all contracts use this way now

Wi1l-B0t:
Done

PR #1518, opened 2026-02-22
Remove: redundant null checking for UInt160.IsValid and UInt256.IsValid

Wi1l-B0t:
Remove redundant null checking for UInt160.IsValid and UInt256.IsValid

Solve #1298


Jim8y (approved):

shargon (commented):
Lgtm

PR #1530, opened 2026-03-15
Fix: Oracle.Request definition

Wi1l-B0t:
Fix Oracle.Request definition: the `filter` and `userdata` can be null.


src/Neo.SmartContract.Framework/Native/Oracle.cs
```diff
@@ -47,8 +48,8 @@ public class Oracle
         /// <param name="url">The URL to request an Oracle response from.</param>
         /// <param name="filter">The filter(json path expression) to apply to the response.</param>
         /// <param name="callback">The method name of calling contract to call when the response is received.</param>
-        /// <param name="userData">The user data to pass to the callback.</param>
+        /// <param name="userData">The user data to pass to the callback</param>
```
shargon:
```suggestion
        /// <param name="userData">The user data to pass to the callback.</param>
```

shargon (approved):

shargon (approved):

PR #1540, opened 2026-03-22
perf(compiler): prefer short-form jumps even with Optimize=None

Jim8y:
## Summary

- Run the jump compression pass during compilation regardless of optimization flags, so short-form jumps are emitted when the target is in range.
- Add regression test covering Optimize=None compilation for a small conditional branch.

## Test Plan

- [x] dotnet format
- [x] dotnet build neo-devpack-dotnet.sln (run twice)
- [x] dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj (run twice)
- [x] dotnet test --filter UnitTest_JumpEncoding.SmallConditional_UsesShortJumpEncoding_EvenWithoutOptimizations


Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

vncoelho (commented):
I think that,  in such scenario, maybe we should introduce a parameter with zero/no optimization

PR #1541, opened 2026-03-22
perf(compiler): optimize dense integer switch dispatch

Jim8y:
## Summary
- Lower large integer-only switch statements to a balanced branch tree under Basic optimization instead of a linear equality chain.
- Keep the existing switch lowering as the fallback for pattern switches and other non-constant cases.
- Add a regression test that asserts large integer switches compile without the linear equality chain.

## Test Plan
- [x] dotnet format
- [x] dotnet build neo-devpack-dotnet.sln (run twice)
- [x] dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj (run twice)
- [x] dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_Switch
- [x] dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_SwitchStatementSemantics
- [x] dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_SwitchJumpTable.IntegerSwitch_UsesBranchTreeInsteadOfLinearEqualityChain


shargon:
Conflicts

Wi1l-B0t:
Is it faster in neo-vm ?

Jim8y:
For this PR, the intended win is at the generated-script level: dense integer switches are lowered to a balanced branch tree instead of a linear equality chain, so the VM executes fewer sequential comparisons on those cases. I have not added a dedicated neo-vm microbenchmark in this PR, so the claim here is reduced dispatch work in the emitted script rather than a separately measured vm benchmark.

shargon (approved):

ajara87 (approved):

Wi1l-B0t (approved):

PR #1543, opened 2026-03-22
feat(framework): add Pausable base

Jim8y:
Adds a minimal Pausable pattern to the framework.

- Pausable base exposes a safe Paused flag and protected Pause/Unpause helpers.
- WhenNotPaused / WhenPaused modifier attributes guard methods based on the paused flag.
- Includes unit tests compiled on-the-fly to validate pause/unpause behavior and modifiers.

Verification:
- dotnet format
- dotnet build neo-devpack-dotnet.sln (x2)
- dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj (x2)

### DIFF
--- a/src/Neo.SmartContract.Framework/Pausable.cs
+++ b/src/Neo.SmartContract.Framework/Pausable.cs
@@ -1,0 +1,36 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// Pausable.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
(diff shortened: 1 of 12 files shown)


shargon (approved):

ajara87 (approved):

src/Neo.SmartContract.Framework/Pausable.cs
```diff
@@ -0,0 +1,36 @@
...
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.SmartContract.Framework.Attributes;
+using Neo.SmartContract.Framework.Services;
+
+namespace Neo.SmartContract.Framework
+{
+    public abstract class Pausable : SmartContract
```
Wi1l-B0t, line 17: The default inteface method is supported ?

`Pausable`  is an abstract class. If any contract extends this class, it cannot extend others.
So it's better to define `Pausable` as an interface.

PR #1547, opened 2026-03-23
feat(template): add packaged nep11 template

Jim8y:
## Summary
- add the missing packaged `neocontractnep11` dotnet template under `src/Neo.SmartContract.Template/templates/`
- add template unit tests and generated testing artifact coverage for the packaged NEP-11 template
- update template artifact cleanup so the packaged NEP-11 template is compiled and validated with the rest of the template package

## Context
This is the validated prerequisite for the remaining P2 roadmap work around NFT templates. Current `master-n3` advertises `neocontractnep11` in the template package README, but the packaged template directory was missing entirely.

## Testing
- dotnet test tests/Neo.SmartContract.Template.UnitTests/Neo.SmartContract.Template.UnitTests.csproj


src/Neo.SmartContract.Template/templates/neocontractnep11/.config/dotnet-tools.json
Wi1l-B0t, line 1: Why is there a hidden folder(starts with `.`)?

Jim8y: Removed the hidden `.config/dotnet-tools.json` in `3632bee8`. It was an accidental tool manifest, not required for the packaged template.

ajara87 (approved):

src/Neo.SmartContract.Template/templates/neocontractnep11/Nep11Contract.cs
```diff
@@ -0,0 +1,133 @@
...
+                Image = image
+            });
+
+            return tokenId;
+        }
+
+        [Safe]
+        public override Map<string, object> Properties(ByteString tokenId)
+        {
+            if (tokenId.Length > 64) throw new Exception("The argument \"tokenId\" should be 64 or less bytes long.");
```
shargon: I think 64 can't be also, beacuse the max key length is 64, and with the prefix it's more

Wi1l-B0t: It's tokenId?

Wi1l-B0t (approved):

src/Neo.SmartContract.Template/templates/neocontractnep11/Nep11Contract.cs
shargon: We have a max length for storage keys, so I think is bot possible to have a 64 bytes token Id

Jim8y:
```suggestion
            if (tokenId.Length >= 64) throw new Exception("The argument \"tokenId\" should be 64 or less bytes long.");
```

ajara87 (approved):

shargon:
TokenId length issue was not solved

Jim8y:
> TokenId length issue was not solved

its upadted according to your comment to "tokenId.Length >= 64"

PR #1550, opened 2026-03-24
feat(compiler): support default interface implementations

Jim8y:
## Summary
- allow interface instance methods with bodies instead of rejecting them with analyzer `NC4052`
- surface inherited default interface methods on implementing classes/contracts during compiler member collection
- compile/export those inherited methods so contracts can rely on interface-provided default implementations
- add focused analyzer and compiler tests for default interface implementation and override behavior

## Scope
This PR supports inherited default interface method implementations on implementing classes/contracts, including contract ABI export and invocation of the inherited method body. It does not broaden general interface-typed runtime dispatch beyond what is required for that inherited method surface.

## Testing
- dotnet test tests/Neo.SmartContract.Analyzer.UnitTests/Neo.SmartContract.Analyzer.UnitTests.csproj
- dotnet build src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj


docs/plans/2026-03-24-interface-default-implementation.md
```diff
@@ -0,0 +1,122 @@
+# Interface Default Implementation Plan
```
shargon: This file is not requred, it's the claude plan

Jim8y: Removed in `0302960b`. The PR no longer includes the internal plan file.

src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -316,6 +320,32 @@ private static IEnumerable<ISymbol> GetAllMembersInternal(ITypeSymbol type)
             }
         }
 
+        private static void AddDefaultInterfaceMethods(INamedTypeSymbol type, List<ISymbol> myMembers)
```
Wi1l-B0t, line 323: What would happen if different interfaces have the same method?

ajara87: Both methods would be added silently (which would be a bug)

src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -316,6 +320,32 @@ private static IEnumerable<ISymbol> GetAllMembersInternal(ITypeSymbol type)
...
+                    continue;
+
+                // Roslyn returns the interface member itself when the type relies on the default body.
+                if (!SymbolEqualityComparer.Default.Equals(implementation, interfaceMethod))
+                    continue;
+
+                if (type.BaseType is INamedTypeSymbol baseType &&
+                    baseType.FindImplementationForInterfaceMember(interfaceMethod) is not null)
+                    continue;
+
```
ajara87:
```suggestion
                var existingConflict = myMembers.OfType<IMethodSymbol>().FirstOrDefault(m => m.Name == interfaceMethod.Name && m.Parameters.Length == interfaceMethod.Parameters.Length && !SymbolEqualityComparer.Default.Equals(m.ContainingType, interfaceMethod.ContainingType));
                if (existingConflict is not null)
                {
                    throw new CompilationException(interfaceMethod, DiagnosticId.MethodNameConflict,
                        $"Ambiguous default interface method '{interfaceMethod.Name}': " +
                        $"both '{existingConflict.ContainingType.Name}' and '{interfaceMethod.ContainingType.Name}' " +
                        $"provide a default implementation. The class '{type.Name}' must provide its own implementation.");
                }
```

ajara87: Detect ambiguous default interface methods: a method with the same name and parameter signature from a different interface is already present

src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -316,6 +320,39 @@ private static IEnumerable<ISymbol> GetAllMembersInternal(ITypeSymbol type)
...
+                // Roslyn returns the interface member itself when the type relies on the default body.
+                if (!SymbolEqualityComparer.Default.Equals(implementation, interfaceMethod))
+                    continue;
+
+                if (type.BaseType is INamedTypeSymbol baseType &&
+                    baseType.FindImplementationForInterfaceMember(interfaceMethod) is not null)
+                    continue;
+                var existingConflict = myMembers.OfType<IMethodSymbol>().FirstOrDefault(m => m.Name == interfaceMethod.Name && m.Parameters.Length == interfaceMethod.Parameters.Length && !SymbolEqualityComparer.Default.Equals(m.ContainingType, interfaceMethod.ContainingType));
```
Wi1l-B0t:
```suggestion
                var existingConflict = myMembers.OfType<IMethodSymbol>()
                    .FirstOrDefault(m => m.Name == interfaceMethod.Name
                            && m.Parameters.Length == interfaceMethod.Parameters.Length
                            && !SymbolEqualityComparer.Default.Equals(m.ContainingType, interfaceMethod.ContainingType));
```

Wi1l-B0t (approved):

Wi1l-B0t:
What behavior of C# if two interfaces have the same method?

Jim8y:
> What behavior of C# if two interfaces have the same method?

it will throw DiagnosticId.MethodNameConflict: NC3003

ajara87 (approved):

src/Neo.Compiler.CSharp/Helper.cs
shargon: We should emit a compiler warning

shargon: This was not solved

Issue #1514, opened 2026-02-10
Devpack review report.


Jim8y:
# Neo DevPack .NET — Comprehensive Review Report

> **Date**: 2026-02-10
> **Branch**: `master-n3`
> **Commit**: `8d49ffd8`
> **Review Team**: 7 specialized AI agents (compiler, framework, security, testing, architecture, devops, documentation)
> **Total Findings**: ~262

---

## Table of Contents

- [Executive Summary](#executive-summary)
- [Priority Action Matrix](#priority-action-matrix)
- [1. Compiler Review](#1-compiler-review)
- [2. Framework Review](#2-framework-review)
- [3. Security Review](#3-security-review)
- [4. Testing Review](#4-testing-review)
- [5. Architecture Review](#5-architecture-review)
- [6. Build / CI / DevEx Review](#6-build--ci--devex-review)
- [7. Documentation Review](#7-documentation-review)
- [Overall Assessment](#overall-assessment)

---

## Executive Summary

| Expert Agent          | Review Domain                                | Findings |
| --------------------- | -------------------------------------------- | -------- |
| `compiler-expert`     | Compiler syntax coverage, bugs, optimization | 29       |
| `framework-expert`    | Framework API design, completeness           | ~30      |
| `security-expert`     | Security analyzers, vulnerability detection  | 12       |
| `testing-expert`      | Test coverage, quality, infrastructure       | 25       |
| `architecture-expert` | Architecture, SOLID, code quality            | 28       |
| `devops-expert`       | Build, CI/CD, Docker, developer experience   | 31       |
| `docs-expert`         | Documentation completeness, accuracy         | 107      |
| **Total**             |                                              | **~262** |

### Severity Distribution

| Severity    | Count | Description                                                     |
| ----------- | ----- | --------------------------------------------------------------- |
| 🔴 CRITICAL | 4     | Broken now — Docker, DevContainer, Makefile, UpdateAnalyzer bug |
| 🟠 HIGH     | ~20   | Compiler bugs, security gaps, framework null derefs             |
| 🟡 MEDIUM   | ~60   | Missing features, test gaps, architecture issues                |
| 🔵 LOW      | ~40   | Code quality, typos, style inconsistencies                      |
| ℹ️ INFO     | ~138  | Documentation improvements, minor suggestions                   |

---

## Priority Action Matrix

> Status updated: 2026-03-30 (based on current `master-n3` at `fa73e3a9`).

[...]

PR #1555, opened 2026-03-29
feat(analyzer): detect duplicate storage map prefixes

Jim8y:
## Summary
- add a Roslyn analyzer that flags duplicate constant `StorageMap` / `LocalStorageMap` prefixes declared in the same contract type
- normalize byte, string, and byte-array prefixes so obvious namespace collisions are surfaced early
- add focused analyzer unit tests for duplicate byte and string prefixes plus non-collision cases

## Context
- first implementation slice for `Security: Storage key collision detection` from #1514
- based on the latest `master-n3`

## Test Plan
- dotnet test tests/Neo.SmartContract.Analyzer.UnitTests/Neo.SmartContract.Analyzer.UnitTests.csproj --filter FullyQualifiedName~StorageKeyCollisionAnalyzerUnitTests

### DIFF
--- a/src/Neo.SmartContract.Analyzer/StorageKeyCollisionAnalyzer.cs
+++ b/src/Neo.SmartContract.Analyzer/StorageKeyCollisionAnalyzer.cs
@@ -59,18 +59,22 @@ namespace Neo.SmartContract.Analyzer
                 if (field.DeclaringSyntaxReferences[0].GetSyntax(context.CancellationToken) is not VariableDeclaratorSyntax declarator)
                     continue;
 
-                if (declarator.Initializer?.Value is not BaseObjectCreationExpressionSyntax creation)
-                    continue;
-
-                SemanticModel semanticModel = context.Compilation.GetSemanticModel(creation.SyntaxTree);
                 if (!IsStorageNamespaceType(field.Type))
                     continue;
 
-                ExpressionSyntax? prefixExpression = GetPrefixExpression(creation);
-                if (prefixExpression is null)
+                if (!TryGetPrefixExpression(
+                        declarator.Initializer?.Value,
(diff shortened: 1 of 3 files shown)


src/Neo.SmartContract.Analyzer/StorageKeyCollisionAnalyzer.cs
```diff
@@ -0,0 +1,257 @@
...
+            Dictionary<string, PrefixUsage> seenPrefixes = new(StringComparer.Ordinal);
+            foreach (IFieldSymbol field in typeSymbol.GetMembers().OfType<IFieldSymbol>())
+            {
+                if (field.DeclaringSyntaxReferences.Length == 0)
+                    continue;
+
+                if (field.DeclaringSyntaxReferences[0].GetSyntax(context.CancellationToken) is not VariableDeclaratorSyntax declarator)
+                    continue;
+
+                if (declarator.Initializer?.Value is not BaseObjectCreationExpressionSyntax creation)
```
ajara87: I think this currently only analyzes direct object creation (new ...) but not assignments like XXX = CreateXXXX();

Jim8y: Fixed in ac37f2f8. The analyzer now follows parameterless factory initializers like `Owners = CreateOwners();` and resolves the returned `StorageMap`/`LocalStorageMap` prefix before checking for collisions. Added a regression test covering both block-bodied and expression-bodied factory methods.

shargon:
Love it!!! Thabks @Jim8y

shargon (approved):

ajara87 (approved):

Issue #1556, opened 2026-03-30
master-n3: validated audit follow-up checklist


Jim8y:
## Summary

This issue tracks the findings that were validated as real on `master-n3` and is updated against the current branch head `421e9db79bb37c10d5088f94fe283f49c52483b7`.

Scope rules for this checklist:
- Includes only findings confirmed during code review and targeted verification
- Excludes claims that were not substantiated
- Separates direct defects from analyzer coverage gaps and framework/design risks

Priority legend:
- `P0` = fix first, high-risk correctness or compiler defect
- `P1` = important bug or analyzer gap
- `P2` = meaningful framework or semantic risk
- `P3` = low-risk cleanup

## Direct Defects

- [x] `Issue 1 / P0` Fix `BigInteger.TryParse(string, out result)` out-handler stack and return semantics
  - File: `src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Out.cs`
  - Close when: success and failure paths match expected `TryParse` behavior, the out capture is written correctly, and regression tests cover valid input, invalid input, and stack-safe control flow.
  - Status: merged in [#1557](https://github.com/neo-project/neo-devpack-dotnet/pull/1557)

- [x] `Issue 2 / P1` Guard `StorageMap.Increase/Decrease` against negative-balance footguns or document permissive semantics explicitly
  - Files: `src/Neo.SmartContract.Framework/Services/StorageMap.cs`, `src/Neo.SmartContract.Framework/Services/LocalStorageMap.cs`
  - Close when: either non-negative invariants are enforced for balance-style helpers or the API is clearly narrowed, renamed, or documented so callers do not mistake it for a safe balance primitive, with tests covering negative amounts and underflow cases.
  - Status: merged in [#1558](https://github.com/neo-project/neo-devpack-dotnet/pull/1558)

- [x] `Issue 3 / P0` Reject duplicate token IDs in `Nep11Token.Mint`
  - File: `src/Neo.SmartContract.Framework/Nep11Token.cs`
  - Close when: minting an existing token ID fails deterministically and tests prove `TotalSupply`, owner balances, and token state cannot be inflated by overwrite.
  - Status: merged in [#1559](https://github.com/neo-project/neo-devpack-dotnet/pull/1559)

- [x] `Issue 4 / P1` Add overflow checks for parameter and local slot counts
[...]

PR #1571, opened 2026-03-31
fix(analyzer): detect CALLT reentrancy edges

Jim8y:
## Summary
- teach `ReEntrancyAnalyzer` to treat `CALLT` opcodes as external contract calls
- add a focused analyzer regression using a native-contract `CALLT` followed by `Storage.Put`
- keep the existing reentrancy analyzer behavior unchanged for syscall-based external calls

## Testing
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~Neo.Compiler.CSharp.UnitTests.SecurityAnalyzer.ReentrancyTests.Test_ReentrancyAnalyzer_Detects_CALLT_Based_External_Call|FullyQualifiedName~Neo.Compiler.CSharp.UnitTests.SecurityAnalyzer.ReentrancyTests.Test_HasReentrancy|FullyQualifiedName~Neo.Compiler.CSharp.UnitTests.SecurityAnalyzer.ReentrancyTests.Test_ReentrancyWithEnhancedDiagnostics"`

Related checklist item: issue #1556, Issue 9.

### DIFF
--- a/src/Neo.Compiler.CSharp/SecurityAnalyzer/ReEntrancyAnalyzer.cs
+++ b/src/Neo.Compiler.CSharp/SecurityAnalyzer/ReEntrancyAnalyzer.cs
@@ -13,6 +13,7 @@ using Neo.Json;
 using Neo.Optimizer;
 using Neo.SmartContract;
 using Neo.SmartContract.Manifest;
+using Neo.SmartContract.Native;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -170,7 +171,9 @@ namespace Neo.Compiler.SecurityAnalyzer
                 {
                     if (instruction.OpCode == VM.OpCode.CALLT)
                     {
-                        callOtherContractInstructions[b].Add(addr);
+                        uint tokenId = instruction.TokenU16;
+                        if (tokenId < nef.Tokens.Length && !IsKnownSafeNativeCallt(nef.Tokens[tokenId]))
+                            callOtherContractInstructions[b].Add(addr);
                     }
                     else if (instruction.OpCode == VM.OpCode.SYSCALL)
                     {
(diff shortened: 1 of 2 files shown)


src/Neo.Compiler.CSharp/SecurityAnalyzer/ReEntrancyAnalyzer.cs
```diff
@@ -208,6 +215,11 @@ public static ReEntrancyVulnerabilityPair AnalyzeSingleContractReEntrancy
             return new(vulnerabilityPairs, callOtherContractInstructions, writeStorageInstructions, debugInfo);
         }
 
+        private static bool IsKnownSafeNativeCallt(MethodToken token)
+        {
+            return token.Hash == NativeContract.StdLib.Hash;
+        }
+
```
ajara87: It seems too broad. Could we restrict it to an allowlist of specific methods that are clearly non-reentrant  without relevant side effects for this analysis instead of exempting the entire StdLib?

Jim8y: Narrowed this from a blanket StdLib hash check to an explicit method allowlist and added a regression test showing unknown StdLib tokens are no longer auto-exempt. Safe known methods like `itoa` still stay ignored for this analysis.

shargon (approved):

src/Neo.Compiler.CSharp/SecurityAnalyzer/ReEntrancyAnalyzer.cs
```diff
@@ -208,6 +239,12 @@ public static ReEntrancyVulnerabilityPair AnalyzeSingleContractReEntrancy
             return new(vulnerabilityPairs, callOtherContractInstructions, writeStorageInstructions, debugInfo);
         }
 
+        private static bool IsKnownSafeNativeCallt(MethodToken token)
```
Wi1l-B0t:
```suggestion
        private static bool IsKnownSafeNativeCallT(MethodToken token)
```

Jim8y: Applied the rename and merged the branch onto latest `master-n3`. The current PR head is `d907621c`.

PR #1574, opened 2026-04-05
Align devpack surfaces with latest master-n3

Jim8y:
## Summary
This PR now only contains changes needed to keep `neo-devpack-dotnet` aligned with current Neo `master-n3` behavior since `v3.9.1`.

Devpack-only fixes and the reverted Policy surface change are intentionally excluded for follow-up PRs.

## Checklist
- [x] Align consumed Neo package versions to the current `master-n3` package line (`3.9.3-CI02036`) so the devpack runtime and tests execute against the same upstream assemblies.
- [x] Match `ApplicationEngine` hardfork jump-table selection in testing to upstream Neo changes:
  - `neo-project/neo#4381` (`JumpTable for not Gorgon`)
  - `neo-project/neo#4516` (`enable SHL/SHR fix starting from Gorgon hardfork`)
- [x] Match `RoleManagement.DesignateAsRole` duplicate-key rejection behavior to `neo-project/neo#4486`.
- [x] Update `StaticStorageMap` byte-prefix expectations to the runtime behavior exposed by the updated Neo/VM package line above.

## Explicitly Out Of Scope
- Policy contract wrapper changes
- RpcStore / injected `HttpClient` / range lookup cleanup
- Iterator wrapper regression coverage
- Wrapper backlog for `NEO.GetAccountState` / `NeoAccountState` / older `RoleManagement` surface parity
- Example project package-plumbing changes

## Verification
- `dotnet format ./neo-devpack-dotnet.sln`
- `dotnet test tests/Neo.SmartContract.Testing.UnitTests/Neo.SmartContract.Testing.UnitTests.csproj --filter "FullyQualifiedName~Neo.SmartContract.Testing.UnitTests.TestEngineTests|FullyQualifiedName~Neo.SmartContract.Testing.UnitTests.TestRoleManagement"`
- `dotnet test neo-devpack-dotnet.sln --filter "FullyQualifiedName!=Neo.SmartContract.Testing.UnitTests.Storage.RpcStoreTests.TestRpcStore"`
- `dotnet test tests/Neo.SmartContract.Testing.UnitTests/Neo.SmartContract.Testing.UnitTests.csproj` still fails from this environment on `RpcStoreTests.TestRpcStore` because the live RPC endpoint resets the connection; this PR does not change that path.
- Local coverage on the verified scope: line `65.04%`, branch `50.39%`.


src/Neo.SmartContract.Testing/TestingApplicationEngine.cs
```diff
@@ -88,11 +88,41 @@ public override UInt160? EntryScriptHash
...
 
         public TestingApplicationEngine(TestEngine engine, TriggerType trigger, IVerifiable container, DataCache snapshot, Block persistingBlock)
-            : base(trigger, container, snapshot, persistingBlock, engine.ProtocolSettings, engine.Fee, null)
+            : base(trigger, container, snapshot, persistingBlock, engine.ProtocolSettings, engine.Fee, null,
+                ResolveJumpTable(engine.ProtocolSettings, snapshot, persistingBlock))
         {
             Engine = engine;
         }
 
+        private static JumpTable ResolveJumpTable(ProtocolSettings settings, DataCache snapshot, Block? persistingBlock)
```
shargon: Is not set the jump table already in base class?

Wi1l-B0t (approved):

ajara87 (approved):

src/Neo.SmartContract.Testing/TestingApplicationEngine.cs
Jim8y: Verified against the same PR package set by reverting only `TestingApplicationEngine` in a throwaway worktree and rerunning the new jump-table tests. With the old `base(..., null)` constructor path, all three regressions fail: pre-Echidna still uses `SubStr`, pre-Gorgon still uses `HasKey`, and the old fault messages are lost. So `null` here is not preserving the historical jump-table behavior we need on this branch.

I pushed a follow-up simplification on top of this PR: direct `HF_Gorgon` / `HF_Echidna` checks, non-null `persistingBlock.Index`, and the inherited `ComposeNotGorgonJumpTable()` instead of the extra defensive helper. That keeps the parity fix and removes the uncovered fallback branches that were tripping Codecov.

shargon (approved):

PR #1576, opened 2026-04-07
fix(compiler): guard unchecked signed division overflow

Jim8y:
## Summary
- apply the existing signed `MinValue / -1` overflow guard for `int` and `long` division in unchecked/default contexts too
- update division overflow tests to assert the runtime exception behavior for unchecked `int` and `long`
- refresh affected compiler artifacts and gas expectations for contracts that now emit the guard

## Test Plan
- [x] `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --nologo`


shargon (approved):

ajara87 (approved):

Wi1l-B0t:
Why is this needed when there is `EnsureIntegerInRange`?

PR #1579, opened 2026-04-08
Fix EndsWith when suffix matches entire string

Jim8y:
## Summary
- fix `string.EndsWith()` lowering to accept a start position of `0` when the suffix length matches the full string length
- add a regression assertion for `"world".EndsWith("world")` and update the expected gas and generated artifact output
- refresh the generated `Contract_String` testing artifact to match the compiler output

## Test Plan
- [x] `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter Test_TestEndWith --nologo`
- [x] `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter FullyQualifiedName~Neo.Compiler.CSharp.UnitTests.UnitTest_String --nologo`
- [x] `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --nologo`


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_String.cs
```diff
@@ -115,11 +115,14 @@ public void Test_TestEndWith()
...
 
+            Assert.IsTrue(Contract.TestEndWith("world"));
+            AssertGasConsumed(1357650);
+
             Assert.IsFalse(Contract.TestEndWith("hel"));
             AssertGasConsumed(1049190);
 
             Assert.IsFalse(Contract.TestEndWith("hello"));
-            AssertGasConsumed(1049190);
+            AssertGasConsumed(1357650);
```
shargon: The price is the same but gas changed?

Jim8y: Previously, when `str.Length == suffix.Length`, `startPosition` was `0` and
  the buggy `> 0` check returned `false` early. After changing it to `>= 0`,
  equal-length inputs now go through the normal `SUBSTR + EQUAL` path.

  So `"hello".EndsWith("world")` is still false, but it now consumes the same
  gas as other equal-length comparisons because it no longer exits early on the
  bugged branch.

Wi1l-B0t:
Do you tests it with `'Hello'.EndsWith('Hello')`?

ajara87:
I would add a test for the empty suffix case Assert.IsTrue(Contract.TestEndWith(""))

Wi1l-B0t (dismissed):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_String.cs
shargon: It doesn't work with "" to me

shargon:
```suggestion
            AssertGasConsumed(1357650);

            Assert.IsTrue(Contract.TestEndWith(""));
            AssertGasConsumed(1357650); // TODO: Fix when fix ''
```

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
ajara87, line 419:
```suggestion
        var validCountTarget = new JumpTarget();
    	var suffixNotEmptyTarget = new JumpTarget();
        methodConvert.Dup();                         // Duplicate suffix
        methodConvert.Size();                        // Get suffix length
        methodConvert.Push(0);
        methodConvert.JumpIfNotEqual(suffixNotEmptyTarget); // if len(suffix) != 0 skip
        methodConvert.Drop();                        // Drop suffix (empty string)
        methodConvert.Drop();                        // Drop original string
        methodConvert.PushT();                       // Push true any string ends with ""
        methodConvert.JumpAlways(endTarget);
        suffixNotEmptyTarget.Instruction = methodConvert.Nop();        
```
to fix ""

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
```diff
@@ -417,6 +417,16 @@ private static void HandleStringEndsWith(MethodConvert methodConvert, SemanticMo
...
+    	var suffixNotEmptyTarget = new JumpTarget();
+        methodConvert.Dup();                         // Duplicate suffix
+        methodConvert.Size();                        // Get suffix length
+        methodConvert.Push(0);
+        methodConvert.JumpIfNotEqual(suffixNotEmptyTarget); // if len(suffix) != 0 skip
+        methodConvert.Drop();                        // Drop suffix (empty string)
+        methodConvert.Drop();                        // Drop original string
+        methodConvert.PushT();                       // Push true any string ends with ""
+        methodConvert.JumpAlways(endTarget);
+        suffixNotEmptyTarget.Instruction = methodConvert.Nop();
```
ajara87:
```suggestion
        suffixNotEmptyTarget.Instruction = methodConvert.Nop();
```

Wi1l-B0t:
UT failed.

shargon (dismissed):

shargon (approved):

PR #1583, opened 2026-04-12
Document: remove StorageContext if it can be removed

Wi1l-B0t:
Update document.
Remove StorageContext usage if it can be removed.

### DIFF
--- a/examples/Example.SmartContract.NFT/Loot.Admin.cs
+++ b/examples/Example.SmartContract.NFT/Loot.Admin.cs
@@ -37,7 +37,7 @@ namespace NFT
         /// Security requirement:
         /// The prefix should be unique in the contract: checked globally.
         /// </summary>
-        private static readonly StorageMap OwnerMap = new(Storage.CurrentContext, (byte)StoragePrefix.Owner);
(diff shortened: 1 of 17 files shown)


examples/Example.SmartContract.NFT/Loot.Admin.cs
```diff
@@ -37,7 +37,7 @@ public partial class Loot
         /// Security requirement:
         /// The prefix should be unique in the contract: checked globally.
         /// </summary>
-        private static readonly StorageMap OwnerMap = new(Storage.CurrentContext, (byte)StoragePrefix.Owner);
+        private static readonly LocalStorageMap OwnerMap = new("owner");
```
shargon: We don't have byte prefix?

Wi1l-B0t: > We don't have byte prefix?

It's just an example.

shargon: People use examples in his code later, and it's more optimized if we use only one byte as a prefix

Wi1l-B0t: Done

shargon (approved):

PR #1586, opened 2026-04-18
Reject unsupported ABI types

Jim8y:
## Summary
- reject floating-point types when contract ABI parameter or return types are mapped
- reject task-like async types from contract ABI mapping
- add compiler tests for unsupported floating-point and task ABI signatures

## Validation
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "ClassName=Neo.Compiler.CSharp.UnitTests.Syntax.UnitTest_UnsupportedAbiTypes" --no-restore


ajara87 (dismissed):

src/Neo.Compiler.CSharp/Helper.cs
```diff
@@ -105,6 +108,25 @@ public static ContractParameterType GetContractParameterType(this ITypeSymbol ty
             return ContractParameterType.Any;
         }
 
+        private static void ThrowIfUnsupportedContractType(ITypeSymbol type)
```
shargon, line 125: Create method, IsValidType? It can be reused in other places

Jim8y: Applied. Added a reusable `IsValidType` predicate and kept the throwing helper as a thin guard so later validation paths can reuse the same unsupported ABI-type rule without duplicating it.

shargon (approved):

ajara87 (approved):

PR #1587, opened 2026-04-18
Analyze floating-point method signatures

Jim8y:
## Summary
- report float, double, and decimal usage in method return types
- report float, double, and decimal usage in method parameters
- add analyzer tests for floating-point method signatures

## Validation
- dotnet test tests/Neo.SmartContract.Analyzer.UnitTests/Neo.SmartContract.Analyzer.UnitTests.csproj --filter "MethodSignature_ShouldReportDiagnostic" --no-restore
- dotnet test tests/Neo.SmartContract.Analyzer.UnitTests/Neo.SmartContract.Analyzer.UnitTests.csproj --no-restore


shargon:
Conflicts, and can reuse a IsValidType method (from other pr)

Wi1l-B0t (approved):

ajara87 (approved):

src/Neo.SmartContract.Analyzer/UnsupportedTypeUsageAnalyzerHelpers.cs
```diff
@@ -0,0 +1,61 @@
...
+        DiagnosticDescriptor rule,
+        Func<ITypeSymbol, object?[]> getMessageArgs)
+    {
+        if (context.Node is not ParameterSyntax parameter || parameter.Type is null) return;
+
+        var type = (context.SemanticModel.GetDeclaredSymbol(parameter, context.CancellationToken) as IParameterSymbol)?.Type;
+        ReportIfValidType(context, parameter.Type.GetLocation(), type, specialType, rule, getMessageArgs);
+    }
+
+    private static void ReportIfValidType(
```
shargon, line 48: Report if valid instead is not valid?

shargon (approved):

src/Neo.SmartContract.Analyzer/UnsupportedTypeUsageAnalyzerHelpers.cs
shargon: This was not resolved

Jim8y:
You're right. I missed this before merging. The helper currently reports when the type matches the analyzer's unsupported `SpecialType`, so the behavior is intentional, but the `IsValidType` / `ReportIfValidType` names are misleading in this context. I’ll clean this up in a follow-up by renaming/reusing the helper with clearer matching semantics.

PR #1600, opened 2026-04-25
Update remaining 2024 copyright headers

Jim8y:
## Summary
- Updates two remaining `2015-2024` source headers to `2015-2026`.
- Leaves file contents otherwise unchanged.

## Tested
- `git diff --check`
- `dotnet build src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj -v minimal`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,7 +17,7 @@ charset = utf-8
 end_of_line = lf
 #indent_style = tab  # TODO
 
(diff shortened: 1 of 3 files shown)


shargon (dismissed):

shargon:
Why it doesn't fault  previously during format?

Jim8y:
It did not fail in the format step because the current formatter configuration does not make the file-header analyzer a failing diagnostic. The `.editorconfig` `file_header_template` value is only a template unless IDE0073/header validation is enabled with a severity that `dotnet format --verify-no-changes` enforces.

ajara87 (approved):

PR #1603, opened 2026-04-25
Add NeoDebugInfo root object error message

Jim8y:
## Summary
- Gives `NeoDebugInfo.Load` a specific `FormatException` message when the debug-info root is not a JSON object.
- Preserves the existing exception type.

## Tested
- `git diff --check`
- `dotnet build src/Neo.SmartContract.Testing/Neo.SmartContract.Testing.csproj -v minimal`


ajara87 (approved):

shargon (approved):
All of these exception messages should be done in the same pr

Jim8y:
Agreed. I split the exception-message fixes too aggressively. For small message-only changes across the same area, one PR would have been easier to review. I’ll batch those next time unless there is a real behavior boundary.

PR #1609, opened 2026-04-25
Avoid duplicate optimizer strategy registration

Jim8y:
## Summary
- track optimizer strategy methods by module and metadata token so discovery cannot register the same strategy twice
- keep strategy delegate registration and priority ordering intact
- run an explicit final jump cleanup pass for patterns introduced by late strategies
- add unit coverage for duplicate registration, invalid registration, named strategies, method token remapping, and final jump cleanup

## Validation
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Break|FullyQualifiedName~UnitTest_GoTo|FullyQualifiedName~UnitTest_Pattern|FullyQualifiedName~UnitTest_OptimizerAutomation" --no-restore


shargon (commented):
Maybe it's better using a dictionary

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Break.cs
```diff
@@ -123,29 +124,31 @@ public abstract class Contract_Break(Neo.SmartContract.Testing.SmartContractInit
     /// PUSHDATA1 FF '?' [8 datoshi]
     /// SYSCALL 077652F3 'System.Storage.Local.Find' [32768 datoshi]
     /// STLOC0 [2 datoshi]
-    /// JMP 7B [2 datoshi]
+    /// JMP_L 83000000 [2 datoshi]
```
Wi1l-B0t: JMP_L 83000000?

Why wasn't it optimized to a Jump instruction?

ajara87:
there are some conflicts. @Jim8y  could you fix it?

Wi1l-B0t:
Conflicts.

src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
```diff
@@ -95,9 +98,23 @@ private static void DiscoverAndOrderStrategies(Assembly assembly)
...
             // Order by priority (higher priority first)
-            orderedStrategies.AddRange(strategyMethods.OrderByDescending(s => s.attribute.Priority));
+            foreach (var (method, attribute) in strategyMethods.OrderByDescending(s => s.attribute.Priority))
+                RegisterStrategyMethod(method, attribute);
+        }
+
+        private static bool RegisterStrategyMethod(MethodInfo method, StrategyAttribute attribute)
+        {
+            var methodId = GetStrategyMethodId(method);
+            if (!registeredStrategyMethods.TryAdd(methodId, method))
```
shargon, line 108: It works with this key?

Jim8y: Yes. `MetadataToken` is only unique within a module, so the key also includes `ModuleVersionId`. Re-discovering the same strategy method produces the same pair and is skipped, while methods from different modules do not collide.

shargon: But Dictionary key works with a tuple?

Jim8y: Yes. This is a `ValueTuple<Guid, int>`, and `Dictionary` uses value equality through `EqualityComparer<T>.Default`. `Guid` and `int` both compare by value, so the pair is a stable dictionary key.

shargon (approved):

src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
```diff
@@ -27,6 +27,7 @@ public static class Optimizer
         public static readonly int[] OperandSizeTable = new int[256];
         public static readonly Dictionary<string, Func<NefFile, ContractManifest, JObject, (NefFile nef, ContractManifest manifest, JObject debugInfo)>> strategies = new();
         private static readonly List<(MethodInfo method, StrategyAttribute attribute)> orderedStrategies = new();
+        private static readonly Dictionary<(Guid moduleVersionId, int metadataToken), MethodInfo> registeredStrategyMethods = new();
```
ajara87, line 30: use ConcurrentDictionary ???

ajara87 (approved):

Wi1l-B0t (approved):

src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
shargon: Other dictionaries are not thread safe, I think this process is not multi-tread

Wi1l-B0t: No concurrent operation.

ajara87:
@shargon merging is blocked. Could you merge it?

src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
Jim8y: Agreed. This path is not expected to run concurrently, and the rest of the optimizer registration code uses regular dictionaries as well. I kept this as a Dictionary so it matches the existing single-threaded registration flow; the new key only prevents duplicate strategy registration when discovery runs more than once.

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Break.cs
Jim8y: This was from an older artifact diff and is now outdated. The current head no longer changes `Contract_Break.cs`; it remains `JMP 7B`.

For the old `JMP_L 83000000` case, `0x83` is a forward offset of 131. A short `JMP` uses a signed byte operand, so it can only encode -128 through 127. That offset could not be represented as a short jump at that point.

PR #1611, opened 2026-04-25
Fail visibly on optimizer strategy errors

Jim8y:
## Summary
- stop swallowing experimental optimizer strategy exceptions
- report the failing strategy name and preserve the underlying exception
- add a regression test that injects and restores a failing strategy

## Validation
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_OptimizerAutomation -v minimal --no-restore


shargon (commented):
I think that is better using generic exception and showing rhe error

Wi1l-B0t (approved):

shargon (approved):

ajara87:
lgtm

PR #1612, opened 2026-04-25
Return compiled contracts deterministically

Jim8y:
## Summary
- keep parallel contract compilation
- return compiled contexts in prepared topological class order instead of ConcurrentDictionary enumeration order
- add a regression test for two-contract source ordering

## Validation
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_CompilationEngineOrdering -v minimal --no-restore


src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
```diff
@@ -339,7 +339,14 @@ private List<CompilationContext> CompileProjectContracts(Compilation compilation
                 Contexts.TryAdd(c, context);
             });
 
-            return Contexts.Select(p => p.Value).ToList();
+            return GetContextsInClassOrder(sortedClasses);
+        }
+
+        private List<CompilationContext> GetContextsInClassOrder(IEnumerable<INamedTypeSymbol> sortedClasses)
+        {
+            return sortedClasses.Select(c => Contexts.TryGetValue(c, out var context)
```
shargon: Sorted classes don't contains all the classes, and context the smart contracts?

Wi1l-B0t:
Why does a source file contain multiple contracts?

Wi1l-B0t:
Can a source file only contain one contract?

src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
ajara87: agree

src/Neo.Compiler.CSharp/CompilationEngine/CompilationEngine.cs
```diff
@@ -339,7 +339,17 @@ private List<CompilationContext> CompileProjectContracts(Compilation compilation
...
+        }
+
+        /// <summary>
+        /// Returns contexts only for sorted smart-contract classes; non-contract helper classes are compiled as supporting symbols and are not returned.
+        /// </summary>
+        private List<CompilationContext> GetContextsInContractOrder(IEnumerable<INamedTypeSymbol> sortedContractClasses)
+        {
+            return sortedContractClasses.Select(c => Contexts.TryGetValue(c, out var context)
+                ? context
+                : throw new InvalidOperationException($"Compilation context for contract '{c.Name}' was not created.")).ToList();
```
shargon, line 352: Context ordered by indexOf in sortedClasses will preserve the previous behavior

shargon (changes requested):

Jim8y:
Closing because deterministic multi-contract result ordering is not considered an issue for this change set.

PR #1621, opened 2026-04-26
Use specific persisting block exceptions

Jim8y:
Summary:
- replace generic PersistingBlock exceptions with more specific exception types
- preserve VM context on OnPersist/PostPersist failures through TestException
- cover the new persisting failure helpers and transaction-state lookup helper

Tests:
- dotnet build src/Neo.SmartContract.Testing/Neo.SmartContract.Testing.csproj
- dotnet test tests/Neo.SmartContract.Testing.UnitTests/Neo.SmartContract.Testing.UnitTests.csproj --filter "TestPersistFailureHelperReportsTriggerName|TestPersistTransactionStateHelperReportsMissingState|TestNextBlock"
- dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj --filter Test_GetTxVMState


ajara87 (approved):

src/Neo.SmartContract.Testing/PersistingBlock.cs
```diff
@@ -206,5 +203,19 @@ private static Header CreateNextHeader(Header previous, TimeSpan elapsed, ulong
                 }
             };
         }
+
+        private static void ThrowIfPersistFailed(string triggerName, TestingApplicationEngine engine, VMState state)
+        {
+            if (state != VMState.HALT)
+                throw new InvalidOperationException($"Error executing {triggerName}.", new TestException(engine));
+        }
+
```
shargon, line 212: This method is only used once, I don't see any improvement with the change

shargon (changes requested):

src/Neo.SmartContract.Testing/PersistingBlock.cs
Wi1l-B0t: > This method is only used once, I don't see any improvement with the change

`GetTransactionState` ?

shargon: Yes

Jim8y:
Closing after review. The requested-change thread is valid: the PR adds a one-use helper and reflection tests against private helpers, while the current code already reports the relevant persisting phase or transaction hash. If this diagnostic path needs work later, it should come back as a narrower behavior-focused PR without the single-use helper.

PR #1625, opened 2026-04-26
Support C# 14 null-conditional assignments

Jim8y:
## Summary
- support C# 14 null-conditional assignments for member, array element, and indexer-property targets
- guard every receiver in a `?.` or `?[]` chain before evaluating indexes or right-hand operands
- support simple assignments, compound assignments, and `??=` while returning `null` when a receiver is skipped
- add runtime tests for null receivers, index and RHS side effects, nested receivers, arrays, multidimensional arrays, and indexers

## Validation
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter FullyQualifiedName~UnitTest_NullConditionalAssignment --no-restore`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --no-restore`


src/Neo.Compiler.CSharp/MethodConvert/Expression/AssignmentExpression.SimpleAssignment.cs
```diff
@@ -407,4 +456,160 @@ private void ConvertConditionalMemberBindingAssignment(SemanticModel model, Memb
...
+    private byte[] StoreConditionalElementIndices(SemanticModel model, ElementBindingExpressionSyntax elementBinding, ITypeSymbol receiverType)
+    {
+        IArrayTypeSymbol arrayType = GetConditionalElementArrayType(elementBinding, receiverType);
+        if (arrayType.Rank > 1)
+        {
+            EnsureMultiDimensionalArguments(elementBinding.ArgumentList.Arguments, arrayType.Rank, elementBinding.ArgumentList);
+        }
+        else if (elementBinding.ArgumentList.Arguments.Count != 1)
+        {
+            throw new CompilationException(elementBinding.ArgumentList, DiagnosticId.MultidimensionalArray, $"Unsupported array rank: {elementBinding.ArgumentList.Arguments}");
```
ajara87: should not be `elementBinding.ArgumentList.Arguments.Count` ?

shargon: At least `elementBinding.ArgumentList.Arguments`

Wi1l-B0t:
Good.
But maybe complicated.

tests/Neo.Compiler.CSharp.TestContracts/Contract_NullConditional.cs
```diff
@@ -9,6 +9,61 @@ private class Node
...
+
+    private static Node CreateNodeWithSideEffect()
+    {
+        SideEffects += 1;
+        return new Node();
+    }
+
+    private static int CreateValueWithSideEffect()
+    {
+        SideEffects += 1;
```
shargon, line 59: How did you check SideEffects?

Jim8y: It is checked through the contract return values in `UnitTest_NullConditionalAssignment`. Each side-effect contract method resets `SideEffects`, runs the null-conditional assignment, then returns the counter or an encoded value that includes it. The tests cover both branches, for example `AssignChildSideEffects(true) == 1` and `AssignChildSideEffects(false) == 0`, plus element/index/value/coalesce/indexer cases such as `AssignElementIndexAndValueSideEffects` and `CoalesceElementIndexAndValueSideEffects`.

shargon: But if is not stored, every call reset the value

Jim8y: Yes, static fields are reset between contract invocations. The tests do not rely on persistence across calls. `SideEffects` is only a per-invocation probe: each method resets it, runs one null-conditional assignment, and returns the counter in the same invocation. The encoded assertions such as `AssignElementIndexAndValueSideEffects(true) == 27` and `CoalesceElementIndexAndValueSideEffects(true, true) == 13` verify how many side-effect expressions ran within that single call.

shargon (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/Expression/AssignmentExpression.SimpleAssignment.cs
```diff
@@ -409,4 +463,262 @@ private void ConvertConditionalMemberBindingAssignment(SemanticModel model, Memb
...
+                if (!property.IsStatic) AccessSlot(OpCode.LDLOC, receiverSlot);
+                CallMethodWithConvention(model, property.GetMethod);
+                ConvertExpression(model, right);
+                EmitComplexAssignmentOperator(type, operatorToken);
+                byte propertyValueSlot = AddAnonymousVariable();
+                AccessSlot(OpCode.STLOC, propertyValueSlot);
+                AccessSlot(OpCode.LDLOC, propertyValueSlot);
+                if (!property.IsStatic) AccessSlot(OpCode.LDLOC, receiverSlot);
+                CallMethodWithConvention(model, property.SetMethod, CallingConvention.Cdecl);
+                AccessSlot(OpCode.LDLOC, propertyValueSlot);
```
ajara87: Could replacing STLOC + 2× LDLOC with a single DUP work here? One copy stays on the stack as the return value and the setter consumes the other

Jim8y: Applied. The conditional property compound assignment now keeps the assigned value on the stack with `DUP` before invoking the setter, so it no longer needs the temporary local.

ajara87 (approved):

shargon (approved):

PR #1632, opened 2026-04-27
Fix string(char, int) constructor lowering

Jim8y:
## Summary
- lower System.String(char, int) constructors before the generic extern constructor path
- emit a ByteString result for constant and dynamic counts, including NUL constants
- add a runtime regression test for empty, NUL, repeated, dynamic-count, dynamic-char, and negative-count cases

## Validation
- pre-fix regression failed with: Return value count mismatch: expected 1, but got 0 items on the evaluation stack
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_StringConstructor --no-restore
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_String" --no-restore
- dotnet build neo-devpack-dotnet.sln
- dotnet test neo-devpack-dotnet.sln --no-build --no-restore --filter "FullyQualifiedName!~RpcStoreTests"

## Note
The unfiltered solution test run was also attempted. It failed only in the external-network RpcStoreTests.TestRpcStore path because http://seed2t5.neo.org:20332 returned a non-JSON response.


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_StringConstructor.cs
```diff
@@ -0,0 +1,121 @@
...
+            """;
+
+        var context = CompileSingleContract(source);
+        Assert.IsTrue(context.Success, string.Join(Environment.NewLine, context.Diagnostics.Select(p => p.ToString())));
+
+        var engine = new TestEngine(true);
+        return engine.Deploy<StringConstructorContract>(context.CreateExecutable(), context.CreateManifest());
+    }
+
+    private static CompilationContext CompileSingleContract(string sourceCode)
```
shargon: I think this method is duplicated along multiple uts

Wi1l-B0t: > I think this method is duplicated along multiple uts

Fixed https://github.com/neo-project/neo-devpack-dotnet/pull/1634

ajara87 (dismissed):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_StringConstructor.cs
shargon: @Wi1l-B0t this should be merged first?

shargon (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
```diff
@@ -30,6 +30,107 @@ private enum SplitSeparatorKind
...
+        var charExpression = GetConstructorArgument(arguments, symbol.Parameters[0], symbol);
+        var countExpression = GetConstructorArgument(arguments, symbol.Parameters[1], symbol);
+        var charConstant = model.GetConstantValue(charExpression);
+        var countConstant = model.GetConstantValue(countExpression);
+
+        if (charConstant.HasValue && charConstant.Value is char character &&
+            countConstant.HasValue && countConstant.Value is int repeatCount &&
+            repeatCount >= 0)
+        {
+            Push(new string(character, repeatCount));
```
Wi1l-B0t, line 53: https://github.com/neo-project/neo-devpack-dotnet/pull/1638
It should be rejected if the string length exceeds the max item size.

shargon: How can be set the limits? Default can be different than the target network limits

Jim8y: Added a compile-time guard for constant `new string(char, int)` sizes. The compiler currently does not expose target-network execution limits in `CompilationOptions`, so this uses `ExecutionEngineLimits.Default.MaxItemSize`, matching the byte-array guard in #1638 and avoiding a silent oversized constant emission without adding a new configuration surface.

shargon (dismissed):

Wi1l-B0t (dismissed):

ajara87 (dismissed):

Wi1l-B0t:
Conflicts.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1634, opened 2026-04-29
Fix: remove duplicate `CompileSingleContract`

Wi1l-B0t:
Remove some duplicate `CompileSingleContract` implementations.


ajara87:
UT fails

shargon:
Require any pr?

Wi1l-B0t:
> UT fails

Fixed.

ajara87 (dismissed):

tests/Neo.Compiler.CSharp.UnitTests/TestHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+using Neo.Compiler.CSharp.UnitTests.Syntax;
+using Neo.Compiler.SecurityAnalyzer;
+using System;
+using System.IO;
+
+namespace Neo.Compiler.CSharp.UnitTests;
+
+public static class TestHelper
+{
+
```
shargon:
```suggestion
```

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_DefaultInterfaceImplementation.cs
```diff
@@ -337,19 +337,12 @@ public class Contract : SmartContract, ILeftGreeting, IRightGreeting
...
 
-        var context = CompileSingleContract(source);
+        var context = TestHelper.CompileSingleContract(source);
         Assert.IsFalse(context.Success, "Competing unrelated default interface methods should require an explicit class implementation.");
         Assert.IsTrue(context.Diagnostics.Any(d => d.Id == "NC3003"),
             string.Join(Environment.NewLine, context.Diagnostics.Select(p => p.ToString())));
     }
 
-    private static CompilationContext CompileSingleContract(string sourceCode)
-    {
```
shargon, line 347: CompileSource is the same, isn't it?

ajara87:
UT Fails

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_DefaultInterfaceImplementation.cs
Wi1l-B0t: > CompileSource is the same, isn't it?

No

Wi1l-B0t:
> UT Fails

OK now.

Wi1l-B0t:
@Jim8y

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_DefaultInterfaceImplementation.cs
Jim8y: The removed helper was only the local wrapper that asserts exactly one compilation context and returns `contexts[0]`. `CompileSource` still returns the engine and all contexts, so callers that need those keep using it. These tests only need the single-contract context, so using `TestHelper.CompileSingleContract` keeps that assertion in one shared place.

shargon (approved):

ajara87 (approved):

ajara87:
ut fails

Jim8y:
This failed during package restore from MyGet before unit tests ran. The feed returned 500/timeouts while downloading `Neo.VM.3.9.3-CI00366`. The failed workflow is already running again.

PR #1639, opened 2026-04-29
Restore GuardHelpers error codes

Jim8y:
## Summary
- Restore the public GuardHelpers default error codes to their established uppercase short-code form.
- Align the framework helper implementation with the existing guard helper test contract semantics.
- Exclude GuardHelpers from CLR line coverage because these methods wrap smart-contract runtime opcodes; behavior is validated through contract VM tests.

## Validation
- `dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj --filter GuardHelpersTest --no-restore`
- `git diff --check`


Wi1l-B0t:
Why ?

Why is the exception information in uppercase letters?

Jim8y:
These were the old public GuardHelpers error codes. This PR only restores them after they were changed to PascalCase.

Wi1l-B0t:
> These uppercase strings are the existing public GuardHelpers abort codes. This PR is restoring compatibility after the docs PR accidentally changed them to PascalCase. Keeping values like `NOT_POSITIVE`, `INVALID_ADDR`, and `NO_WITNESS` avoids changing observable contract behavior for callers that key off abort messages.

What?

AI reply?

Jim8y:
No. Short version: those were the old public error codes. This PR only restores them after they were changed to PascalCase.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Are you sure that can't be covered? I think it should work

Jim8y:
I checked the direct coverage path by switching the test contract to call Neo.SmartContract.Framework.Helpers.GuardHelpers and removing the coverage exclusion. It does not compile today: the compiler tries to inline the referenced helper methods and fails with `ArgumentException: Syntax node is not within syntax tree` from `TryProcessInlineMethods`.

The existing runtime tests cover the same public error strings through the contract-local guard helper test contract, but direct source coverage for this framework helper needs a compiler-side inline fix first. I kept this PR scoped to restoring the public error codes.

shargon (approved):

PR #1640, opened 2026-04-29
Fix security docs tables of contents

Jim8y:
## Summary
- Fix stale table-of-contents entries in the security testing guide.
- Fix stale table-of-contents entries in the safe arithmetic guide.
- Keep only links that resolve to headings still present in each document.

## Validation
- Verified every table-of-contents anchor in `docs/security/security-testing.md` resolves to an existing heading.
- Verified every table-of-contents anchor in `docs/security/safe-arithmetic.md` resolves to an existing heading.
- `git diff --check`
- `git diff --cached --check`

### DIFF
--- a/docs/security/safe-arithmetic.md
+++ b/docs/security/safe-arithmetic.md
@@ -6,15 +6,13 @@ This guide provides comprehensive patterns for safe arithmetic in Neo N3 smart c
 ## Table of Contents
 
 - [Arithmetic Security Fundamentals](#arithmetic-security-fundamentals)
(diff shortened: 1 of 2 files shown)


shargon (changes requested):
Join with #1641

Jim8y:
Done. I combined the safe arithmetic ToC fix from #1641 into this PR and closed #1641, so this PR now contains both documentation ToC fixes.

Wi1l-B0t (approved):

Jim8y:
#1641 has been combined into this PR. Could you re-review this one now?

shargon (approved):

PR #1641, opened 2026-04-29
Fix safe arithmetic table of contents

Jim8y:
## Summary
- Fix the safe arithmetic guide table of contents so every anchor resolves.
- Remove stale entries for sections that are no longer present.
- Point the BigInteger entry at the current heading name.

## Validation
- Verified every table-of-contents anchor in `docs/security/safe-arithmetic.md` resolves to an existing heading.
- `git diff --check`


shargon (changes requested):
Please join with #1640

Jim8y:
Combined into #1640.

PR #1642, opened 2026-04-29
Fix named arguments with omitted defaults

Jim8y:
## Summary
- Bind regular call arguments by parameter symbol instead of indexing the raw argument list by parameter ordinal.
- Preserve default parameter values when earlier parameters are omitted and later parameters are supplied by name.
- Add regression coverage for `Add(b: 5)` and mixed positional/named arguments with omitted defaults using the existing syntax probe helper.

## Validation
- Confirmed the new `UnitTest_NamedArguments` tests failed before the compiler fix with NC2001.
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_NamedArguments`
- `git diff --check`


Wi1l-B0t (approved):
Good fix.

shargon (approved):

ajara87 (approved):

PR #1643, opened 2026-04-29
Return contract contexts in deterministic order

Jim8y:
## Summary
- Return compiled contract contexts in the deterministic topological order already computed by the compiler.
- Avoid exposing `ConcurrentDictionary` enumeration order to callers.
- Add regression coverage for multi-contract source ordering.

## Validation
- Confirmed the new ordering test failed before the fix due to concurrent dictionary result order.
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_CompilationEngineReference`
- `git diff --check`


Jim8y:
Closing this draft because deterministic multi-contract result ordering is not considered an issue for this change set.

PR #1645, opened 2026-04-29
Clarify unsupported type diagnostics and semantics

Jim8y:
## Summary
- Remove decimal from floating-point diagnostic suggestions because decimal is unsupported in contracts.
- Document Neo contract semantics for direct typeof expressions.
- Rename unsupported-type analyzer helper methods so their names match the behavior.

## Validation
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_CompilationException"
- dotnet test tests/Neo.SmartContract.Analyzer.UnitTests/Neo.SmartContract.Analyzer.UnitTests.csproj --filter "FullyQualifiedName~DecimalUsageAnalyzerUnitTests|FullyQualifiedName~FloatUsageAnalyzerUnitTests|FullyQualifiedName~DoubleUsageAnalyzerUnitTests"

Supersedes #1646 and #1647.

PR #1646, opened 2026-04-29
Document typeof contract semantics

Jim8y:
## Summary
- Document Neo contract semantics for direct `typeof(T)` expressions.
- Clarify that `typeof` lowers to a simple type name string rather than a .NET `System.Type` object.
- Note that this support primarily exists for enum helper compilation.

## Validation
- Compiled a temporary contract returning `typeof(int)` with assembly output and verified it lowers to `Int32`.
- `git diff --check`


Jim8y:
Combined into #1645.

PR #1648, opened 2026-04-30
Add source to workflow

ajara87:
Add source to avoid error 500.

Related to #1645


Wi1l-B0t (dismissed):

shargon (commented):
Why don't take https://github.com/neo-project/neo-devpack-dotnet/blob/master-n3/NuGet.Config?

ajara87:
> Why don't take https://github.com/neo-project/neo-devpack-dotnet/blob/master-n3/NuGet.Config?

let me check

PR #1650, opened 2026-05-02
Fix: InstructionToString with PUSHDATA2, PUSHDATA4

Wi1l-B0t:
Only OpCode.PUSHDATA1, No PUSHDATA2 and PUSHDATA4


ajara87 (approved):

shargon (approved):
Should be good if you can add a test for it

Wi1l-B0t:
> Should be good if you can add a test for it

No test project for disassembler now.

shargon:
> > Should be good if you can add a test for it
> 
> No test project for disassembler now.

That's why I said that should be good

PR #1651, opened 2026-05-02
Optimize: use one opcode instead of comparison-with-jump if numeric comparison

Wi1l-B0t:
Use one opcode instead of comparison-with-jump if numeric comparison for string methods.


vncoelho:
Looks good to me

ajara87 (approved):
lgtm

shargon (commented):
technically it's different, `NumEqual` only accept BigInteger types, and now accept any kind

Wi1l-B0t:
> technically it's different, `NumEqual` only accept BigInteger types, and now accept any kind

only numeric comparison changed.

Wi1l-B0t:
> technically it's different, `NumEqual` only accept BigInteger types, and now accept any kind

`JMPEQ`, `JUMPGE`, etc. only accept BigInteger too.

shargon:
For example:
```
public virtual void JmpEq(ExecutionEngine engine, Instruction instruction)
{
    var x2 = engine.Pop().GetInteger();
    var x1 = engine.Pop().GetInteger();
    if (x1 == x2)
        ExecuteJumpOffset(engine, instruction.TokenI8);
}
```
vs
```
 public virtual void Equal(ExecutionEngine engine, Instruction instruction)
 {
     var x2 = engine.Pop();
     var x1 = engine.Pop();
     engine.Push(x1.Equals(x2, engine.Limits));
 }
```
With this change we doesn't check that the type it's a number, now it will be converted from ByteString for example to BigInteger

Wi1l-B0t:
> For example:
> 
> ```
> public virtual void JmpEq(ExecutionEngine engine, Instruction instruction)
> {
>     var x2 = engine.Pop().GetInteger();
>     var x1 = engine.Pop().GetInteger();
>     if (x1 == x2)
>         ExecuteJumpOffset(engine, instruction.TokenI8);
> }
> ```
> 
> vs
> 
> ```
>  public virtual void Equal(ExecutionEngine engine, Instruction instruction)
>  {
>      var x2 = engine.Pop();
>      var x1 = engine.Pop();
>      engine.Push(x1.Equals(x2, engine.Limits));
>  }
> ```
> 
> With this change we doesn't check that the type it's a number, now it will be converted from ByteString for example to BigInteger

Only the case of NumEqual(and LE, GE etc.) was modified, the case of Equal was not modified.
```
    public virtual void NumEqual(ExecutionEngine engine, Instruction instruction)
    {
        var x2 = engine.Pop().GetInteger();
        var x1 = engine.Pop().GetInteger();
        engine.Push(x1 == x2);
    }
```

Jim8y (commented):
I reviewed this pass against the changed string helpers. The substitutions keep numeric operands (lengths, indices, memorySearch results, and char codes) on the stack before using the comparison jump opcodes, so I do not see a correctness issue in the diff. Leaving this as a comment rather than approval because codecov/project is still failing.

Wi1l-B0t:
> I reviewed this pass against the changed string helpers. The substitutions keep numeric operands (lengths, indices, memorySearch results, and char codes) on the stack before using the comparison jump opcodes, so I do not see a correctness issue in the diff. Leaving this as a comment rather than approval because codecov/project is still failing.

Only `-0.02%`.

Jim8y (approved):
Re-reviewed the numeric comparison jump substitutions and ran the affected string/enum tests locally: `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_String|FullyQualifiedName~UnitTest_Enum" --no-restore` (57/57). No blocking code findings from my side. This is a code-review approval only; the branch is still behind `master-n3` and `codecov/project` is still failing, so those need to be resolved before merge.

shargon (approved):

PR #1655, opened 2026-05-06
Fix master-n3 test project classification

Jim8y:
## Summary
- Mark the two TestContracts projects as non-test projects so solution-level `dotnet test` does not invoke VSTest on helper contract assemblies.
- Limit the workflow's `coverlet.msbuild` setup to real UnitTests projects instead of every project under `tests`.
- Remove an unused ApplicationInsights import from `Contract_PostfixUnary` that was only compiling because test package references leaked into the helper contract project.
- Add a regression test covering the TestContracts project classification.

## Root cause
`Neo.Compiler.CSharp.TestContracts` and `Neo.SmartContract.Framework.TestContracts` are helper contract assemblies, but they were marked with `IsTestProject=true`. That caused solution-level test execution and the CI coverage setup step to treat them as test projects. After the helper projects were corrected, the workflow still used `find tests -name *.csproj`, so it attempted to add `coverlet.msbuild` to the helper projects and remained exposed to feed failures during package mutation.

## Validation
- `find tests -path '*UnitTests/*.csproj' | sort`
- `git diff --check`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj -c Release --filter "Test_Abort|TestContractsAreNotMarkedAsTestProjects"`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj -c Release --no-build --filter TestContractsAreNotMarkedAsTestProjects`
- `dotnet build tests/Neo.Compiler.CSharp.TestContracts/Neo.Compiler.CSharp.TestContracts.csproj -c Release`
- `dotnet build tests/Neo.SmartContract.Framework.TestContracts/Neo.SmartContract.Framework.TestContracts.csproj -c Release`
- `dotnet test neo-devpack-dotnet.sln -c Release --logger "console;verbosity=minimal"`
- `dotnet test neo-devpack-dotnet.sln -c Release --no-build --logger "console;verbosity=minimal"`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,7 +74,7 @@ jobs:
     - name: Build Solution
       run: dotnet build ./neo-devpack-dotnet.sln
     - name: Add package coverlet.msbuild
-      run: find tests -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
+      run: find tests -path '*UnitTests/*.csproj' | xargs -I % dotnet add % package coverlet.msbuild
     - name: Test Neo.Compiler.CSharp.UnitTests
       run: |
         dotnet test ./tests/Neo.Compiler.CSharp.UnitTests \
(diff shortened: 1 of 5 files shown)


shargon (approved):

ajara87 (commented):
lgtm

PR #1658, opened 2026-05-07
Map NuGet package sources

Jim8y:
## Summary
- Map public packages to NuGet.org and Neo CI packages to the Neo MyGet feed.
- Limit the CI `coverlet.msbuild` package install to NuGet.org.
- Reduce failures caused by the Neo MyGet feed being queried for packages that are available from NuGet.org.

## Validation
- `ruby -r rexml/document -e "REXML::Document.new(File.read('NuGet.Config')); puts 'nuget config xml ok'"`
- `git diff --check`
- `dotnet restore ./src/Neo.SmartContract.Analyzer/Neo.SmartContract.Analyzer.csproj --configfile NuGet.Config --no-cache -v minimal`

## Notes
- Packages such as `Neo`, `Neo.Extensions`, and `Neo.IO` at `3.9.3-CI02036` are still served by the Neo MyGet feed because they are not available on NuGet.org.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,7 +74,7 @@ jobs:
     - name: Build Solution
       run: dotnet build ./neo-devpack-dotnet.sln
     - name: Add package coverlet.msbuild
-      run: find tests -path '*UnitTests/*.csproj' | xargs -I % dotnet add % package coverlet.msbuild
+      run: find tests -path '*UnitTests/*.csproj' | xargs -I % dotnet add % package coverlet.msbuild --source https://api.nuget.org/v3/index.json
     - name: Test Neo.Compiler.CSharp.UnitTests
       run: |
         dotnet test ./tests/Neo.Compiler.CSharp.UnitTests \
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

Wi1l-B0t (approved):

Wi1l-B0t:
Good job.

.github/workflows/main.yml
```diff
@@ -74,7 +74,7 @@ jobs:
     - name: Build Solution
       run: dotnet build ./neo-devpack-dotnet.sln
     - name: Add package coverlet.msbuild
-      run: find tests -path '*UnitTests/*.csproj' | xargs -I % dotnet add % package coverlet.msbuild
+      run: find tests -path '*UnitTests/*.csproj' | xargs -I % dotnet add % package coverlet.msbuild --source https://api.nuget.org/v3/index.json
```
shargon, line 77: This is required?

shargon (approved):

.github/workflows/main.yml
Jim8y: Yes. This is required because the workflow restores many packages, and CI has been frequently hitting 500 responses from MyGet when public packages are resolved through that feed. Limiting this install to NuGet.org avoids one of the recurring MyGet failure paths while keeping MyGet available for the Neo CI packages that still need it.

PR #1661, opened 2026-05-08
Cache restored packages in CI

Jim8y:
## Summary
- Store the NuGet package cache inside the workflow workspace.
- Cache restored packages for the Test job with a dependency-specific key.
- Use an exact cache key without restore fallbacks so dependency version changes create a fresh cache instead of restoring older packages.

## Why
CI still needs Neo CI packages from the MyGet feed. MyGet has been returning intermittent 500 responses during restore, which can fail PR checks before the code is built. Caching the restored package folder reduces repeated requests to MyGet for the same dependency set while keeping MyGet available when a new Neo package version is required.

## Testing
- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/main.yml'); puts 'workflow yaml ok'"`
- `git diff --check`
- GitHub Actions Test job passed

## Notes
GitHub Actions cache entries are immutable, so this workflow does not use restore fallbacks. When dependency versions change, the key changes and CI creates a new cache instead of restoring the previous dependency set. Older caches will expire under the normal cache retention policy.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   DOTNET_VERSION: 10.0.x
+  NUGET_PACKAGES: ${{ github.workspace }}/.nuget/packages
   DOTNET_TEST_PARAMETERS: --no-build /p:CollectCoverage=true -l "console;verbosity=detailed"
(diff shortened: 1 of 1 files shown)


shargon (approved):
It waa removed in the past, I remember already this cache

Jim8y:
Thanks. This was reintroduced with a narrower cache policy: the key is tied to the current dependency set and the workflow does not use restore fallbacks, so older package sets are not restored after dependency version changes. The goal is to reduce repeated MyGet downloads for the same dependencies without keeping stale packages active.

Wi1l-B0t (approved):
Good.

ajara87 (approved):

PR #1669, opened 2026-05-09
Optimize: less checks for EnsureIntegerInRange

Wi1l-B0t:
Not all operations need to check the upper and lower bounds.
For INC, only upper bound should be checked, and for DEC, only lower should should be checked.


Jim8y (changes requested):
Requesting changes because the range-check pruning is not safe for externally supplied contract parameters.

`EmitIncrementOrDecrement` now skips the lower-bound check for `++` and skips the upper-bound check for `--`. That is safe only if the current value is already within the CLR type range. Contract ABI integer parameters can be invoked with values outside the declared C# range; the existing tests already exercise this by calling `UnitTest_Param_Dec_Checked(-1)`. With the new logic, `UnitTest_Param_Inc_Checked(-2)` can advance through -1/0 without the checked lower-bound failure, and `UnitTest_Param_Dec_Checked(uint.MaxValue + 2)` can advance through max+1/max without the checked upper-bound failure. Please either preserve both bounds for parameter-derived values or add coverage proving those cases still throw/wrap as intended.

PR #1670, opened 2026-05-10
Fix: underflow if value is Int256.MinValue

Wi1l-B0t:
Fix: `BigInteger.IsPowerOfTwo`:

If the `n` is Int256.MinValue, the  `n-1` will underflow and execution will be failed.

`(n & (n-1) == 0) and (n != 0)` -> ` (n > 0) and (n & (n-1) == 0`


Jim8y (dismissed):
The positive-value guard avoids the DEC/AND path for zero and negative BigInteger values, which matches IsPowerOfTwo semantics and avoids the Int256.MinValue underflow. Existing signed IsPow2 tests cover negative inputs, and CI is green.

src/Neo.Compiler.CSharp/MethodConvert/Statement/SwitchStatement.cs
```diff
@@ -66,13 +66,18 @@ private void ConvertSwitchStatement(SemanticModel model, SwitchStatementSyntax s
 
             // handle possible normal labels in all sections of this switch
             foreach (SwitchSectionSyntax section in syntax.Sections)
+            {
                 foreach (StatementSyntax label in section.Statements)
+                {
+
```
shargon:
```suggestion
```

shargon (approved):

ajara87 (approved):

shargon (approved):

PR #1671, opened 2026-05-10
Add: unit tests for Disassembler

Wi1l-B0t:
Add unit tests for Disassembler


Jim8y (changes requested):
Requesting changes for a public API compatibility issue.

`InstructionToString` is a public extension method. Renaming the optional parameter from `addPrice` to `withPrice` breaks source compatibility for callers using `InstructionToString(addPrice: false)`. Please keep the existing parameter name and update the new tests to call `addPrice: false`, or omit the named argument.

The solution file also appears to gain a UTF-8 BOM. Please remove that incidental change if it was not intentional.

Wi1l-B0t:
> Requesting changes for a public API compatibility issue.
> 
> `InstructionToString` is a public extension method. Renaming the optional parameter from `addPrice` to `withPrice` breaks source compatibility for callers using `InstructionToString(addPrice: false)`. Please keep the existing parameter name and update the new tests to call `addPrice: false`, or omit the named argument.
> 
> The solution file also appears to gain a UTF-8 BOM. Please remove that incidental change if it was not intentional.

Fixed.

ajara87 (dismissed):

shargon (dismissed):

neo-devpack-dotnet.sln
```diff
@@ -1,3 +1,4 @@
+﻿
```
shargon:
```suggestion
```

shargon (dismissed):

src/Neo.Disassembler.CSharp/Disassembler.cs
```diff
@@ -29,15 +29,14 @@ public static class Disassembler
 
     public static List<Instruction> ConvertScriptToInstructions(byte[] script)
     {
-        var res = EnumerateInstructions(script);
-
-        return res.Select(x => x.instruction).ToList();
+        return EnumerateInstructions(script)
+            .Select(x => x.instruction)
+            .ToList() ?? [];
```
ajara87:
```suggestion
        return EnumerateInstructions(script)
            .Select(x => x.instruction)
            .ToList();
```
ToList never return null so I think `?? []` is not needed

shargon (approved):

ajara87 (approved):

Jim8y (approved):
Re-reviewed after the updates. The solution BOM change is gone and the public `InstructionToString(..., bool addPrice = true)` parameter name remains source-compatible for named callers. The new test project passed locally with restore: `dotnet test tests/Neo.Disassembler.CSharp.Tests/Neo.Disassembler.CSharp.Tests.csproj` (12/12).

PR #1672, opened 2026-05-10
Fix: Drop the result if exception thrown for Math.BigMul

Wi1l-B0t:
Drop the result if exception thrown for Math.BigMul.
Avoid that the stack has the invalid result if exception.


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Math.cs
```diff
@@ -156,7 +156,7 @@ public void divRemUInt_test()
         }
 
         [TestMethod]
-        public void divRemULong_test()
+        public void TestDivRemUlong()
```
ajara87:
```suggestion
        public void TestDivRemULong()
```

shargon (approved):

ajara87 (approved):

PR #1675, opened 2026-05-10
Fix: BigInteger.Compare may fail because overflow

Wi1l-B0t:
BigInteger.Compare gets result with `Sub`, but `Sub` operation may overflow, and execution wil be fail.

So `left - right` cannot be used.


shargon:
Overflow in BigInteger?

ajara87:
It might be good to add tests for `long.MaxValue vs -1` and `long.MinValue vs 1`

Wi1l-B0t:
> Overflow in BigInteger?

BigInteger is Int256 in Neo.VM.

Not real BigInteger without size limit.

Wi1l-B0t:
> It might be good to add tests for `long.MaxValue vs -1` and `long.MinValue vs 1`

Added.

ajara87 (approved):

shargon (approved):

PR #1676, opened 2026-05-10
Fix: wrong behavior if argument is zero for  `BigInteger.GreatestCommnDivisor`

Wi1l-B0t:
From https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.greatestcommondivisor?view=net-10.0
> If the left and right parameters are non-zero numbers, the method always returns at least a value of 1 because all numbers can be divided by 1. If either parameter is zero, the method returns the absolute value of the non-zero parameter. If both values are zero, the method returns zero.

The current implementation is wrong if argument is zero.


ajara87:
@Wi1l-B0t  some conflicts. Could you check it?

Wi1l-B0t:
> @Wi1l-B0t some conflicts. Could you check it?

Done.

shargon (approved):

ajara87 (approved):

Jim8y (approved):
Reviewed the GCD stack flow and zero-argument cases. The updated loop preserves the Euclidean algorithm and returns the absolute remaining value for both initial and computed zero remainders. Targeted test passed locally: `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter Test_GreatestCommonDivisor --no-restore`.

PR #1683, opened 2026-05-14
Fix inline method parameter scope handling

Jim8y:
## Summary
- store aggressive-inline arguments in temporary local slots before converting the inline body
- route return statements inside inline bodies to an inline-local end target
- honor NoInline for MethodImplOptions.AggressiveInlining
- detect AggressiveInlining when it is combined with other MethodImplOptions flags
- support aggressive-inline extension methods invoked with instance syntax by including the receiver as the first inline argument
- add regression coverage for caller parameter reuse, repeated inline parameter reads, NoInline behavior, combined MethodImplOptions, and extension-method receivers

## Testing
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_Inline
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "UnitTest_Inline|UnitTest_Extensions"
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Inline|FullyQualifiedName~UnitTest_InitSlot|FullyQualifiedName~UnitTest_SwitchJumpTable"
- git diff --check


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Inline.cs
```diff
@@ -58,5 +161,40 @@ public void Test_ArrowMethodNoReturn()
...
+            return (int.Parse(range[..dashIndex]), int.Parse(range[(dashIndex + 1)..]));
+        }
+
+        public abstract class InlineMainContract(SmartContractInitialize initialize)
+            : Neo.SmartContract.Testing.SmartContract(initialize)
+        {
+            [DisplayName("main")]
+            public abstract BigInteger? Main(BigInteger? value);
+        }
+
```
shargon:
```suggestion
```

shargon:
Maybe we can remove `NoInline` from options

ajara87 (approved):

Wi1l-B0t (approved):

PR #1684, opened 2026-05-14
Preserve method argument evaluation order

Jim8y:
## Summary
- evaluate observable regular arguments in source order before arranging the call stack for Cdecl
- preserve named/default argument binding while keeping the existing stack order expected by callees
- add regression coverage for positional and named calls with side effects
- update generated artifacts affected by the corrected argument order

## Testing
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "UnitTest_ArgumentEvaluationOrder|UnitTest_Linq|UnitTest_ByteString|UnitTest_Reentrancy|UnitTest_NameofUnboundGeneric"
- git diff --check


shargon (dismissed):

ajara87 (dismissed):

Wi1l-B0t:
Why more LDLOC/STLOC opcodes used?

Jim8y:
Good question. The extra local-slot traffic is intentional for this fix. For cdecl calls, the final NeoVM stack order is reversed, but C# argument expressions must still be evaluated left-to-right. Emitting directly in VM stack order can make observable side effects happen in the wrong order.\n\nThis change evaluates the source arguments in C# order, stores the resulting values in temporary locals, and then reloads them in the order required by the call. Even simple-looking operands may need to be captured when another argument can mutate state before the final reload; otherwise cases like reading a value before a later mutating argument can compile with different C# semantics.\n\nThe preservation path is limited to multi-argument cdecl calls with observable argument evaluation, and by-ref/params cases are intentionally left out of this PR. So the additional LDLOC/STLOC instructions are the cost of preserving C# evaluation order for this specific path, not accidental output churn.

Wi1l-B0t (dismissed):

Jim8y:
Updated this path to reduce the extra local-slot traffic.

For Cdecl calls where the argument order already matches the parameter order, the compiler now evaluates the arguments left-to-right and reverses the top argument segment once before the call. That preserves C# evaluation order while replacing the previous STLOC/LDLOC pairs with SWAP/REVERSE3/REVERSE4/REVERSEN where possible.

The local-slot fallback is still kept for cases that cannot be represented by a simple stack reversal, such as out-of-order named arguments. The refreshed artifacts show fewer local slots and smaller scripts in the affected call paths.

Jim8y:
Synced with the latest `master-n3`; CI is green again. Ready for review.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_InitSlot.cs
```diff
@@ -103,6 +103,55 @@ public static int Probe(string value)
...
+                    }
+                }
+                """;
+
+            var context = TestHelper.CompileSingleContract(source);
+            var nef = context.CreateExecutable();
+            var manifest = context.CreateManifest();
+            var initialize = manifest.Abi.GetMethod("_initialize", 0);
+
+            Assert.IsNotNull(initialize, "Static field initialization should emit an _initialize method.");
```
shargon, line 144: We should use test engine to test devpack

Jim8y: Updated the test to deploy the compiled contract with TestEngine and assert the static initializer result from VM execution. The INITSLOT assertion is kept as the narrow regression guard for the slot allocation fix.

Jim8y:
Synced with the latest `master-n3`; CI and Codecov are green again. Ready for review.

Wi1l-B0t (approved):

shargon (approved):

PR #1686, opened 2026-05-14
Optimize: EQUAL is null-safe in Neo.VM

Wi1l-B0t:
The opcode `EQUAL` is null-safe in Neo.VM.

So null-checking is not necessary for `Nullable<T>.Equals`.


shargon (commented):
It's covered by tests?

Wi1l-B0t:
> It's covered by tests?

Yes.

shargon (approved):

Jim8y (approved):
Reviewed the nullable equality optimization. Using NeoVM EQUAL for nullable equality preserves the expected null-safe behavior while removing the previous manual null branches.

Verified locally:
- `dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj --filter NullableTest` passed: 44/44

The remaining blocker I see is the failed Codecov patch check.

PR #1688, opened 2026-05-15
Fix nullable ToString receiver emission

Jim8y:
## Summary
- Remove the duplicate receiver conversion from `Nullable<T>.ToString()` code generation.
- Add a compiler regression test that verifies direct nullable `ToString()` calls load the receiver once.

## Motivation
Direct `Nullable<T>.ToString()` calls were emitting the receiver expression twice. The extra value was left below the string result on the VM stack, which increased script cost and could pollute callers that compose the generated expression.

## Validation
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_NullableToString.NullableIntegerToString_LoadsReceiverOnce" --no-restore`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_NullableToString|FullyQualifiedName~UnitTest_NullableType" --no-restore`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_NullableToString|FullyQualifiedName~UnitTest_NullableType|FullyQualifiedName~UnitTest_NullConditional" --no-restore`
- `dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj --filter "FullyQualifiedName~NullableTest" --no-restore`


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_NullableToString.cs
```diff
@@ -0,0 +1,57 @@
...
+
+public class Contract : SmartContract
+{
+    public static string Main(int? value) => value.ToString();
+}
+""");
+
+        Assert.IsTrue(context.Success, string.Join(Environment.NewLine, context.Diagnostics.Select(p => p.ToString())));
+
+        var methodBlock = ExtractMethodBlock(context.CreateAssembly(), "Contract.Main(int?)");
```
shargon: The test didn't execute the code in VM so it's not possible no ensure that it works as expected

shargon (changes requested):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Nullable.cs
```diff
@@ -437,14 +437,12 @@ private static void HandleNullableULongGetValueOrDefault(MethodConvert methodCon
     /// </remarks>
     private static void HandleNullableToString(MethodConvert methodConvert, SemanticModel model, IMethodSymbol symbol, ExpressionSyntax? instanceExpression, IReadOnlyList<SyntaxNode>? arguments)
     {
-        if (instanceExpression is not null)
```
Wi1l-B0t, line 440: Why removed?

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_NullableToString.cs
Jim8y: Added VM execution coverage for nullable ToString with both non-null and null values, including a composed expression case. I kept the assembly assertion as a narrow guard for the receiver duplication regression.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Nullable.cs
Jim8y: The removed conversion loaded the nullable receiver before this handler loaded it again for the null check. That left the original receiver under the string result on the VM stack. The handler now evaluates the receiver once, then prepares arguments, matching the nullable bool ToString flow.

Jim8y:
Addressed the review feedback, synced with the latest `master-n3`, and CI is green. Ready for re-review.

shargon (approved):

ajara87 (approved):

Wi1l-B0t (approved):

PR #1694, opened 2026-05-17
Fix enum parse ignoreCase expression handling

Jim8y:
## Summary
- Store the evaluated Enum.Parse ignoreCase argument before it is consumed.
- Reuse that stored value when choosing case-sensitive or uppercase enum member names.
- Add regression coverage where ignoreCase is supplied as an expression.

## Root cause
The Enum.Parse(type, value, ignoreCase) handler used the evaluated ignoreCase argument for normalizing the input string, but later reloaded LDARG1 while comparing enum member names. That only worked when the contract method parameter at index 1 was exactly the ignoreCase value passed to Enum.Parse.

## Validation
- NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Enum.TestEnumParseIgnoreCaseFromExpression" --no-restore
- NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Enum" --no-restore
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_SystemCallDsl" --no-restore


tests/Neo.Compiler.CSharp.TestContracts/Contract_Enum.cs
```diff
@@ -35,6 +35,11 @@ public static object TestEnumParseIgnoreCase(string value, bool ignoreCase)
             return System.Enum.Parse(typeof(TestEnum), value, ignoreCase);
         }
 
+        public static object TestEnumParseIgnoreCaseFromExpression(string value, bool ignoreCase)
+        {
+            return System.Enum.Parse(typeof(TestEnum), Normalize(value), !ignoreCase);
```
shargon: Why test it's inverted with `!ignoreCase`?

Jim8y: Good point. I renamed the test parameter to `caseSensitive`, so the inversion now reads as converting the caller-facing flag to the `ignoreCase` argument expected by `Enum.Parse`. The expression is intentional because this regression needs the evaluated argument to differ from the original method parameter.

shargon (approved):

Wi1l-B0t (approved):

ajara87 (approved):

PR #1696, opened 2026-05-17
Fix: char.IsSymbol

Wi1l-B0t:
The `char.IsSymbol` in C# standard library  returns `true` if the char value is one of 
```
<$+<=>^`|~>  (for ASCII)
```
But the current implementation returns `true` if the char value is one of:
```
$%&'()*+ <=> >?@ [\]^_` {|}~ (for ASCII).
```

The fixed implementation has less opcodes, and the same behavior as  C# standard library (for ASCII).

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Char.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Char.cs
@@ -10,6 +10,8 @@
 // modifications are permitted.
 
 using System.Collections.Generic;
+using System.Globalization;
+using System.Numerics;
 using System.Runtime.InteropServices;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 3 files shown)


src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Char.cs
```diff
@@ -277,27 +279,24 @@ private static void HandleCharIsPunctuation(MethodConvert methodConvert, Semanti
...
-    /// Algorithm: Checks if character is within ASCII symbol ranges: $%&'()*+ <=> >?@ [\]^_` {|}~
+    /// Algorithm: Checks if character is within ASCII symbol ranges: "<$+<=>^`|~>"
     /// </remarks>
     private static void HandleCharIsSymbol(MethodConvert methodConvert, SemanticModel model, IMethodSymbol symbol,
         ExpressionSyntax? instanceExpression, IReadOnlyList<SyntaxNode>? arguments)
     {
         if (arguments is not null)
             methodConvert.PrepareArgumentsForMethod(model, symbol, arguments);
         var endTarget = new JumpTarget();
+        var bits = BigInteger.Parse("50000001400000007000081000000000", NumberStyles.AllowHexSpecifier);
```
ajara87, line 290: how is calculated this value?

ajara87 (approved):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Char.cs
Wi1l-B0t: > how is calculated this value?

It's a bit map. 
The bit is 1 if `char.IsSymbol` returns true for this char value.

shargon (approved):

PR #1698, opened 2026-05-18
Fix string case conversion receiver expressions

Jim8y:
## Summary
- Evaluate string `ToLower` and `ToUpper` receivers before running the case conversion loops.
- Keep direct parameter and local receivers on the efficient load path.
- Add regression coverage for expression receivers.

## Testing
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_String"`


Wi1l-B0t:
What kind of bug is it?

Give an example?

Jim8y:
This is an observable receiver-evaluation bug. For example, in a call such as NextValue().ToUpper(), NextValue() must run first and the uppercase conversion must use that returned string. The old conversion path was written around the already-loaded argument shape, so expression receivers could be handled incorrectly when the receiver had side effects or needed to be evaluated before conversion.\n\n#1707 supersedes this PR with the same receiver fix, enum ignore-case compatibility, and additional regression coverage, so I am closing this one to avoid duplicate review on the same code path.

PR #1707, opened 2026-05-18
Fix string case conversion receiver handling

Jim8y:
## Summary
- Evaluate the receiver expression for string.ToLower() and string.ToUpper() before case conversion.
- Reuse the evaluated receiver from a local slot during the conversion loop instead of reading arg0 directly.
- Keep enum ignore-case conversions compatible with the shared uppercase helper by preserving the original input where those handlers require it.
- Add coverage for local and expression receivers so transformed strings keep receiver-side changes.

## Testing
- NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_StringCaseReceiver|FullyQualifiedName~UnitTest_String|FullyQualifiedName~UnitTest_Enum" --no-restore


shargon:
@Jim8y please check your auto-summary, \n was shown

ajara87:
conflicts

Jim8y:
Updated the PR description so the line breaks render correctly. I also rebased the branch on the latest master-n3 and reran the string/enum regression tests.

shargon (approved):

ajara87 (approved):

vncoelho (approved):

PR #1708, opened 2026-05-18
Fix char IsSymbol non-ASCII result

Jim8y:
## Summary
- Return false explicitly for non-ASCII input in the shared ASCII character bitmap helper.
- Prevent the original character value from being left as the method result when the ASCII bitmap check is skipped.
- Add coverage for non-ASCII characters in the existing char method tests.

## Testing
- NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_Char" --no-restore


ajara87 (dismissed):

shargon:
Conflicts (and \n in description)

Jim8y:
Resolved the conflicts, fixed the PR description formatting, and rebased the branch on the latest master-n3. I also kept the fix in the shared ASCII character bitmap helper so non-ASCII inputs return false instead of leaking the original character value as the result.

shargon (approved):

ajara87 (approved):

PR #1711, opened 2026-05-19
Reject unsupported ToString conversions

### LINKED ISSUE #860: Map.toString() throwing Exception: An unhandled exception was thrown. Key not found in Map
Problem:
When I call the toString() method in a map, the VM will throw an exception. The problems happen in runtime and not compile time. I'm not sure if this is a compiler error or not (probably is). 

Compile and run the following code:
```
Map<String, object> myNft = new Map<String, object>();
myNft["name"] = "My NFT";
myNft["description"] = "My NFT Description";
var tokenString = myNft.ToString(); // The problem is here
```

Result:
 ```
 Invoke results:
Gas Consumed: 0.0295854 GAS
Exception: An unhandled exception was thrown. Key not found in Map
Result: [  ]
```

However, if I return the map and print it locally, it works.

### PR DESCRIPTION
**Pull request by Jim8y**
## Problem
`Map.ToString()` and other unsupported object conversions compile today, but can fail at runtime with confusing VM errors such as `Key not found in Map`. Developers should get a clear compile-time diagnostic when a conversion cannot be represented safely.

## Solution
- Reject unsupported `object.ToString()` conversions during compilation.
- Keep supported string and VM-backed object conversions compiling.
- Add regression coverage for `Map.ToString()` diagnostics and `string.ToString()` compatibility.

Fixes #860

## Testing
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_ToStringDiagnostics --no-restore`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_String|FullyQualifiedName~UnitTest_Types_ECPoint|FullyQualifiedName~UnitTest_UIntTypes|FullyQualifiedName~UnitTest_ToStringDiagnostics" --no-restore`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_ToStringDiagnostics|FullyQualifiedName~UnitTest_Foreach" --no-restore`
- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --no-restore`
- `dotnet format --no-restore --verify-no-changes --verbosity minimal`
- `git diff --check`

### CODE REVIEW

File: src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs [hunk 1]
```diff
@@ -776,11 +776,39 @@ private static void HandleCharToString(MethodConvert methodConvert, SemanticMode
...
+    {
+        if (type is null)
+            return false;
+
+        if (type.SpecialType is SpecialType.System_String or SpecialType.System_Object)
+            return true;
+
+        return type.ToString() switch
+        {
+            "char" or
```
Wi1l-B0t, line 801: `int` `long` is not supported?

shargon (changes requested):
Description must be fixed

ajara87:
> Description must be fixed

agree. Ping @Jim8y

Jim8y:
Updated the PR description with a clearer runtime problem statement, compile-time diagnostic behavior, issue reference, and validation steps.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
Jim8y: Direct numeric `ToString()` overloads are still supported through the numeric handlers registered for `int`, `long`, and the other integer types. This guard only applies to the `object.ToString()` fallback path. I added a focused regression test for `int`, `long`, and `BigInteger` so that behavior is explicit in this PR.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ToStringDiagnostics.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        var exception = Assert.ThrowsException<TargetInvocationException>(() =>
+            handler.Invoke(null, [null, null, objectToStringSymbol, null, null]));
+
+        Assert.IsInstanceOfType(exception.InnerException, typeof(CompilationException));
+        Assert.AreEqual(DiagnosticId.InvalidToStringType, ((CompilationException)exception.InnerException!).Diagnostic.Id);
+
+        var isSupported = methodConvertType.GetMethod("IsSupportedObjectToStringType", BindingFlags.NonPublic | BindingFlags.Static)!;
+        Assert.IsFalse((bool)isSupported.Invoke(null, [null])!);
+        Assert.IsTrue((bool)isSupported.Invoke(null, [CreateCompilation("").GetSpecialType(SpecialType.System_Char)])!);
```
shargon: Why test with reflection?

Jim8y: Removed the reflection-based test. The coverage now goes through normal contract compilation paths, and the object `ToString()` handler keeps the unsupported-type guard without testing private methods directly.

shargon (commented):
Please test with TestEngine

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_ToStringDiagnostics.cs
Jim8y: Added TestEngine coverage for the supported `ToString()` paths. The new test compiles a contract, deploys it with `TestEngine`, and executes numeric (`int`, `long`, `BigInteger`), string, and char `ToString()` conversions in the VM.

Validation passed:

- `NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_ToStringDiagnostics --no-restore`
- `NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~UnitTest_ToStringDiagnostics|FullyQualifiedName~UnitTest_String|FullyQualifiedName~UnitTest_Types_ECPoint|FullyQualifiedName~UnitTest_UIntTypes" --no-restore`
- `dotnet format --no-restore --verify-no-changes --verbosity minimal`
- `git diff --check`

ajara87 (approved):

shargon (approved):

PR #1714, opened 2026-05-20
Fix: Div operation shouldn't throw exception in unchecked context

Wi1l-B0t:
Div operation shouldn't throw exception in unchecked context,
and overflow already checked in `EnsureIntegerInRange` for `/=` in checked context.

### DIFF
--- a/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_DivisionOverflow.cs
+++ b/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_DivisionOverflow.cs
@@ -72,16 +72,20 @@ namespace Neo.Compiler.CSharp.UnitTests
         }
 
         /// <summary>
-        /// Test: int.MinValue /= -1 in unchecked context should throw.
+        /// Test: int.MinValue /= -1 in unchecked context shouldn't throw.
         /// Compound division should align with binary division.
         /// </summary>
         [TestMethod]
-        public void Test_DivideAssignUncheckedInt32_Overflow_ShouldThrow()
+        public void Test_DivideAssignUncheckedInt32_ShouldWork()
         {
-            Assert.ThrowsException<TestException>(() =>
+            var r1 = Contract.DivideAssignUncheckedInt32(int.MinValue, -1);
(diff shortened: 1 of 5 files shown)


ajara87 (dismissed):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_DivisionOverflow.cs
```diff
@@ -72,16 +72,20 @@ public void Test_DivideUncheckedInt32_Overflow_ShouldThrow()
         }
```
ajara87, line 72: should not be updated this test too?

Jim8y:
Pushed a small cleanup to keep the compiler diff focused on the unchecked compound division change. The extra `opText`/unsupported-operator formatting change was unrelated to this PR and was likely what made Codecov patch coverage fail.

Validation:
- `NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_DivisionOverflow --no-restore`
- `NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_MemberAccess --no-restore`

Jim8y (changes requested):
I think this changes observable C# semantics for variable division. In .NET, both of these throw `OverflowException` even in an unchecked context when the operands are variables:

```csharp
unchecked { var c = int.MinValue / -1; }
unchecked { var a = int.MinValue; a /= -1; }
```

The existing binary division tests already assert that unchecked `int.MinValue / -1` and `long.MinValue / -1` throw. Compound assignment should remain equivalent to `x = (T)(x / y)` except that `x` is evaluated once, so `/=` should not wrap when the underlying division overflows.

With this change, compound division is handled differently from binary division because the explicit divide-overflow check is removed and `EnsureIntegerInRange` wraps the result afterward in unchecked context. I think the test expectation should stay as throwing, and the compiler should preserve the division overflow check for `/=`.

PR #1715, opened 2026-05-21
Fix: right shift does not require checking for overflow

Wi1l-B0t:
Right shift does not require checking for overflow.

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Expression/AssignmentExpression.ComplexAssignment.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Expression/AssignmentExpression.ComplexAssignment.cs
@@ -284,8 +284,7 @@ internal partial class MethodConvert
         bool isBoolean = itemType == VM.Types.StackItemType.Boolean;
         bool isString = itemType == VM.Types.StackItemType.ByteString;
 
-        var opText = operatorToken.ValueText;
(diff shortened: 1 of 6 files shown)


shargon (dismissed):

ajara87 (dismissed):

Jim8y:
Pushed a small cleanup to keep the compiler diff focused on `>>=` avoiding the result range check. The unrelated `opText` refactor/default-branch formatting was removed, which should address the Codecov patch failure without changing the PR behavior.

Validation:
- `NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_Shift --no-restore`
- `NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter "FullyQualifiedName~ComplexAssign" --no-restore`

shargon (approved):

Jim8y (approved):
Looks good to me. The checked and unchecked right-shift assignment paths now compile to the same `SHR` sequence, and the targeted shift and complex-assignment tests pass.

ajara87 (approved):

PR #1716, opened 2026-05-22
Fix: remove  unnecessary overflow checking for AND, OR, XOR

Wi1l-B0t:
Remove unnecessary overflow checking for AND, OR, XOR for less GAS consumption.

Because no overflow for AND, OR and XOR.


ajara87:
could you add UTs for codecov?

Wi1l-B0t:
> could you add UTs for codecov?

Done.

ajara87 (approved):

shargon (approved):

PR #1719, opened 2026-05-30
CreateValueType throw exception

ajara87:
If CreateInstance then an exception should be throwed.

### DIFF
--- a/src/Neo.SmartContract.Testing/Extensions/TestExtensions.cs
+++ b/src/Neo.SmartContract.Testing/Extensions/TestExtensions.cs
@@ -214,7 +214,7 @@ namespace Neo.SmartContract.Testing.Extensions
         private static object CreateValueType(IEnumerable<StackItem> objects, Type valueType, IStringInterpreter stringInterpreter)
         {
             var arr = objects.ToArray();
-            var value = Activator.CreateInstance(valueType) ?? new NoNullAllowedException("Impossible create value type");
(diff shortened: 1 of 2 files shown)


shargon (commented):
The test is not related to the change, it doesn't receive the exception

ajara87:
> The test is not related to the change, it doesn't receive the exception

You're right the test is not covering the fix itself since `Activator.CreateInstance` never returns null for value types in .NET. I'll rename it to better reflect what it actually tests. The ?? throw fix is defensive code but still semantically correct

Wi1l-B0t:
UT?

shargon:
> > The test is not related to the change, it doesn't receive the exception
> 
> You're right the test is not covering the fix itself since `Activator.CreateInstance` never returns null for value types in .NET. I'll rename it to better reflect what it actually tests. The ?? throw fix is defensive code but still semantically correct

The exception received in the test is different than the ones was thrown

shargon:
Why was closed?

ajara87:
> Why was closed?

by mistake

ajara87:
> > > The test is not related to the change, it doesn't receive the exception
> > 
> > 
> > You're right the test is not covering the fix itself since `Activator.CreateInstance` never returns null for value types in .NET. I'll rename it to better reflect what it actually tests. The ?? throw fix is defensive code but still semantically correct
> 
> The exception received in the test is different than the ones was thrown

UT removed

shargon (approved):

PR #1720, opened 2026-06-02
Optimize: use IsContract to check contract existence

Wi1l-B0t:
Optimzie for Nep17Token and Nep11Token: use `IsContract` to check contract existence.


shargon (approved):
Good!

ajara87 (approved):

PR #1721, opened 2026-06-03
Fix: Check null at first when converting  `byte[]` to `UInt160&UInt256`

Wi1l-B0t:
Check null or not at first when converting  `byte[]` to `UInt160&UInt256`,
Because the `null` cannot be converted.


shargon (approved):

ajara87 (approved):

shargon:
@Wi1l-B0t please check ECPoint also

Wi1l-B0t:
> @Wi1l-B0t please check ECPoint also

OK

PR #1726, opened 2026-06-05
Sanitize generated ABI source names

Jim8y:
Summary
- Escape ABI names before writing them into generated C# string literals.
- Sanitize generated class, method, event, and parameter identifiers.
- Add coverage for artifact and contract interface source generation.

Tests
- dotnet test neo-devpack-dotnet.sln --filter "FullyQualifiedName~TestGetArtifactsSourceEscapesAbiNames|FullyQualifiedName~TestGenerateInterfaceEscapesAbiNames" --verbosity minimal


src/Neo.Compiler.CSharp/CSharpSourceName.cs
```diff
@@ -0,0 +1,85 @@
...
+            return identifier.StartsWith('@') ? identifier[1..] : identifier;
+        }
+
+        public static string StringLiteralValue(string value)
+        {
+            var builder = new StringBuilder(value.Length);
+
+            foreach (var ch in value)
+            {
+                builder.Append(ch switch
```
Wi1l-B0t, line 65: How about `\xXX`, \uUUUU`?

Wi1l-B0t (dismissed):

src/Neo.Compiler.CSharp/CSharpSourceName.cs
Jim8y: Added coverage for hex and unicode escape sequences; they stay escaped in generated source.

src/Neo.Compiler.CSharp/Extensions/ArtifactExtensions.cs
```diff
@@ -357,20 +278,21 @@ private static string CreateSourceEventFromManifest(ContractEventDescriptor ev,
 
             // Add On prefix
 
-            var evName = TongleLowercase(EscapeName(ev.Name));
+            var evName = CSharpSourceName.Identifier(TongleLowercase(ev.Name), "Event");
             if (!evName.StartsWith("On")) evName = "On" + evName;
+            var delegateName = "del" + CSharpSourceName.IdentifierSuffix(TongleLowercase(ev.Name), "Event");
```
shargon, line 283: Delegate instead of event?

ajara87 (dismissed):

src/Neo.Compiler.CSharp/ContractInterfaceGenerator.cs
```diff
@@ -119,9 +126,9 @@ private static string GetPropertyName(string methodName)
             // Extract property name from get/set_X method names
             if (methodName.StartsWith("get_") || methodName.StartsWith("set_"))
             {
-                return methodName.Substring(4);
+                return CSharpSourceName.Identifier(methodName.Substring(4), "Property");
             }
-            return methodName;
+            return CSharpSourceName.Identifier(methodName, "Property");
```
shargon:
```suggestion
                return CSharpSourceName.Identifier(methodName.Substring(4), "Method");
}
            return CSharpSourceName.Identifier(methodName, "Method");
```

shargon: This fallback should be method

shargon: Ping @Jim8y

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1728, opened 2026-06-05
Escape temporary project XML values

Jim8y:
Summary
- Escape temporary project XML attribute values for package references, project references, and source files.
- Add coverage that parses the generated temporary project content.

Tests
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter FullyQualifiedName~UnitTest_TempProjectGeneration --verbosity minimal


src/Neo.Compiler.CSharp/CompilationEngine/TemporaryProjectWorkspace.cs
```diff
@@ -137,6 +137,28 @@ internal static string BuildTempProjectContent(CompilationSourceReferences refer
 </Project>";
         }
 
+        private static string EscapeXmlAttributeValue(string value)
```
Wi1l-B0t, line 141: use `SecurityElement.Escape` instead?

src/Neo.Compiler.CSharp/CompilationEngine/TemporaryProjectWorkspace.cs
```diff
@@ -137,6 +137,28 @@ internal static string BuildTempProjectContent(CompilationSourceReferences refer
...
+        {
+            var builder = new StringBuilder(value.Length);
+
+            foreach (var ch in value)
+            {
+                builder.Append(ch switch
```
Wi1l-B0t: It's not enough.

SecurityElement.Escape is better?

src/Neo.Compiler.CSharp/CompilationEngine/TemporaryProjectWorkspace.cs
Jim8y: Switched to SecurityElement.Escape and kept attribute whitespace round-tripping.

shargon (approved):

ajara87 (approved):

PR #1731, opened 2026-06-05
Fix: make string printable when write asm to file

Wi1l-B0t:
There is unexpected behaviour when print unprintable string to `.asm` file.


src/Neo.Compiler.CSharp/MethodConvert/Instruction.cs
```diff
@@ -392,7 +392,8 @@ private static bool TryGetString(Span<byte> span, [NotNullWhen(true)] out string
         {
             try
             {
-                s = Utility.StrictUTF8.GetString(span);
+                var raw = Utility.StrictUTF8.GetString(span);
+                s = System.Text.Json.JsonSerializer.Serialize(raw); // use json serializer to escape the string
```
shargon: This will return "XXX"

Wi1l-B0t: > This will return "XXX"

Yes. It's expected.

ajara87:
```suggestion
                s = raw.Any(c => char.IsControl(c)) ? System.Text.Json.JsonSerializer.Serialize(raw) : raw; // use json serializer to escape the string
```
Using `JsonSerializer.Serialize` unconditionally wraps every string in quotes which changes the `.asm` format for all strings. Consider only applying the serialization when the string actually contains control characters

Wi1l-B0t: This operations just for `PUSHDATA`, and some unprintable chars(for example  '\x00') is not control symbol.

ajara87: `\x00` (`U+0000`, the NULL character) is a control character according to Unicode. In .NET `char.IsControl('\x00')` returns `true` because `U+0000` belongs to the C0 Controls range (`U+0000-U+001F`) which is exactly what `char.IsControl` detects

Wi1l-B0t: '\x00' Just an example.

Wi1l-B0t: > '\x00' Just an example.

Another example: '\n'.

shargon: Was not called when artifacts are created?

ajara87: `\n` is covered by `char.IsControl` too. Using `char.IsControl` makes the change smaller, safer, and more compatible because it only escapes strings that actually need it whereas the proposed version changes the output of absolutely every string

Wi1l-B0t: > Was not called when artifacts are created?

Only for `.asm` file(with option `--assembly`).

shargon (approved):

PR #1742, opened 2026-06-16
Fix: keep the behavior of the `/=, %=, <<=` consistent with the `/, %, <<`.

Wi1l-B0t:
`/=, %=, <<=` have some different behaviors with `/' %, <<`.
This PR fixes these inconsistencies.

### DIFF
--- a/tests/Neo.Compiler.CSharp.TestContracts/Contract_ComplexAssign.cs
+++ b/tests/Neo.Compiler.CSharp.TestContracts/Contract_ComplexAssign.cs
@@ -101,6 +101,11 @@ namespace Neo.Compiler.CSharp.TestContracts
             return (value1, value2);
         }
 
+        public static int LeftShiftAssignChecked(int value, int shift)
+        {
+            return checked(value <<= shift);
+        }
+
         public static (uint, int) UnitTest_Left_Shift_Assign_UnChecked()
         {
             uint value1 = uint.MaxValue;
--- a/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Contract_ComplexAssign.cs
+++ b/tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Contract_ComplexAssign.cs
@@ -73,7 +73,10 @@ namespace Neo.Compiler.CSharp.UnitTests
         public void Test_LeftShiftAssign_Checked()
         {
             Assert.ThrowsException<TestException>(Contract.UnitTest_Left_Shift_Assign_Checked);
-            AssertGasConsumed(1002120);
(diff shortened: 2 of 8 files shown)


tests/Neo.Compiler.CSharp.TestContracts/Contract_ComplexAssign.cs
```diff
@@ -101,6 +101,11 @@ public static (uint, int) UnitTest_Left_Shift_Assign_Checked()
             return (value1, value2);
         }
 
+        public static int LeftShiftAssignChecked(int value, int shift)
+        {
+            return checked(value << shift);
```
ajara87:
```suggestion
            return checked(value <<= shift);
```
should not be <<= instead of << to test the case of this pr?

shargon: I think so

Wi1l-B0t: Fixed.

ajara87 (approved):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Contract_ComplexAssign.cs
```diff
@@ -73,7 +73,10 @@ public void Test_MulAssign_UnChecked()
         public void Test_LeftShiftAssign_Checked()
         {
             Assert.ThrowsException<TestException>(Contract.UnitTest_Left_Shift_Assign_Checked);
-            AssertGasConsumed(1002120);
+            AssertGasConsumed(1002420);
+
+            Assert.ThrowsException<TestException>(() => Contract.LeftShiftAssignChecked(1, 33));
+            AssertGasConsumed(1062780);
```
shargon, line 79: No test without the exception?

Wi1l-B0t: Tested by other cases.

Jim8y (approved):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Contract_ComplexAssign.cs
shargon: It's new, how It's tested by other cases?

Wi1l-B0t: LeftShiftAssign that not overflow has been tested.

PR #1745, opened 2026-06-18
Avoid ByteString enumeration in character helpers

Jim8y:
## Summary
- Rewrite ByteString character classification helpers to use `Length` and index access.
- Add a regression check to keep these helpers away from unsupported enumeration.
- Update the generated Regex artifact and gas expectations.

## Test Plan
- `dotnet test tests/Neo.SmartContract.Framework.UnitTests/Neo.SmartContract.Framework.UnitTests.csproj --filter "CharacterClassHelpers_DoNotUseByteStringEnumeration|Regex" --no-restore -v minimal`

### DIFF
--- a/src/Neo.SmartContract.Framework/ByteString.Extension.cs
+++ b/src/Neo.SmartContract.Framework/ByteString.Extension.cs
@@ -22,8 +22,9 @@ public static class ByteStringExtension
     /// <returns>True if is number</returns>
     public static bool IsNumber(this ByteString byteString)
     {
-        foreach (var value in byteString)
+        for (int i = 0; i < byteString.Length; i++)
         {
+            byte value = byteString[i];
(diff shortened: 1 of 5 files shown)


ajara87 (approved):

shargon:
@Jim8y why it's better to avoid the foreach? What's the problem?

ajara87:
> @Jim8y why it's better to avoid the foreach? What's the problem?

In Neo VM `ByteString` is a primitive stack type. Index access `byteString[i]` is translated to PICKITEM which is a native and wellsupported instruction. A foreach over `ByteString` would require an enumerator which does not exist in the VM

shargon (approved):
We should fix foreach in byte string at compiler level

Wi1l-B0t (approved):

PR #1748, opened 2026-06-21
Fix: NEP-11 transfer must not be marked Safe

Jim8y:
## Problem

The NEP-11 manifest compliance check (`ContractManifestExtension.Nep11.cs`) validated the `transfer` method with an inverted condition and an incorrect message:

```csharp
if (transferMethod1 is { Safe: false } &&
    transferMethod1.ReturnType != ContractParameterType.Boolean)
    errors.Add(... "transfer, it is not safe, you should add a 'Safe' attribute to the transfer method");
```

This is wrong in two ways:

1. **Inverted rule.** A NEP-11 `transfer` changes state and emits the `Transfer` event, so it **must not** be marked `Safe`. The check only fired when `transfer` was *not* Safe (combined with a wrong return type) and never flagged the actual violation — a `transfer` that *is* marked `Safe`.
2. **Contradictory guidance.** The message told developers to *add* a `[Safe]` attribute, the exact opposite of the rule, and contradicts the guidance `CheckStandards` itself prints ("Do not write `[Safe]` for `Transfer` method!").

The sibling NEP-17 check (`ContractManifestExtensions.Nep17.cs`) already implements the correct rule.

## Fix

Mirror the NEP-17 logic for both the 3-parameter (indivisible) and 5-parameter (divisible) `transfer` overloads:

- Report when `transfer` **is** marked `Safe`, with the message `transfer, it should not be marked as Safe`.
- Keep the return-type (`Boolean`) check independent, so a wrong return type is reported on its own.

## Tests

Added two regression tests to `UnitTest_ManifestStandards`:

- `Nep11_TransferMarkedSafe_ProducesExpectedDiagnostic` — a NEP-11 contract whose `transfer` is marked `Safe` now produces the diagnostic.
- `Nep11_CompliantTransfer_ProducesNoSafeDiagnostic` — a compliant NEP-11 contract produces neither the new diagnostic nor the old, inverted message.

Full `Neo.Compiler.CSharp.UnitTests` suite passes locally (1285 tests).


Wi1l-B0t (commented):
Minor Fix.

ajara87 (approved):
`Nep11_DivisibleTransferMarkedSafe_ProducesExpectedDiagnostic` constructs the divisible manifest by patching an indivisible contract which is functional but slightly fragile. A dedicated divisible NEP11 reference contract would be more robust long-term but not a blocker.

lgtm

shargon (approved):

PR #1749, opened 2026-06-21
Fix: guard Transfer event parameter access in NEP standard checks

Jim8y:
## Problem

The NEP-11 and NEP-17 manifest compliance checks look up the `Transfer` event **by name only**:

```csharp
var transferEvent = manifest.Abi.Events.FirstOrDefault(a => a.Name == "Transfer");
```

They then validate each parameter by index. The per-parameter checks ran even when the parameter count was wrong:

```csharp
if (transferEvent is not null && transferEvent.Parameters.Length != 4)
    errors.Add(... "parameters length is not 4");          // reports, but does not stop
if (transferEvent is not null && transferEvent.Parameters[1].Type != ...) // crashes on a short event
```

A contract declaring a malformed `Transfer` event (fewer parameters than the standard requires) therefore crashed the compiler with `IndexOutOfRangeException` instead of producing a diagnostic. NEP-17 has the same defect: its per-parameter checks live inside the `else` branch but still index unconditionally after the length check.

## Fix

Guard the per-parameter checks behind the expected length (`4` for NEP-11, `3` for NEP-17). The length-mismatch diagnostic is still reported; the individual parameter checks are skipped when the count is wrong, so a malformed event yields a clear error instead of a crash.

Valid manifests are unaffected — events with the correct parameter count are validated exactly as before.

## Tests

Added regression tests to `UnitTest_ManifestStandards` covering a malformed `Transfer` event for both standards:

- `Nep11_MalformedTransferEvent_DoesNotCrashAndReportsLength`
- `Nep17_MalformedTransferEvent_DoesNotCrashAndReportsLength`

Both previously threw `IndexOutOfRangeException`; they now pass. Full `Neo.Compiler.CSharp.UnitTests` suite passes locally (1285 tests).


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1750, opened 2026-06-21
Fix: NEP-26 onNEP11Payment tokenId type should be ByteArray

Jim8y:
## Problem

The NEP-26 compliance check required the `onNEP11Payment` `tokenId` parameter (the third parameter) to be a `String`:

```csharp
onNEP11PaymentMethod.Parameters[2].Type == ContractParameterType.String
```

A NEP-11 token id is a `ByteString`, which maps to the `ByteArray` contract parameter type. This is the type used by the canonical `INEP26.OnNEP11Payment` interface:

```csharp
public void OnNEP11Payment(UInt160 from, BigInteger amount, ByteString tokenId, object? data = null);
```

and asserted by `Nep11TokenTest.Test_OnNEP11Payment_Allows_ByteString_TokenId`.

As a result, any contract that correctly implements the NEP-26 receiver produced a false-positive `Incomplete NEP standard NEP-26 implementation: onNEP11Payment` diagnostic. The in-repo `Contract_SupportedStandard26` test contract triggers exactly this.

## Fix

Expect `ByteArray` for the `tokenId` parameter, matching the interface and the rest of the framework.

## Tests

Added regression tests to `UnitTest_ManifestStandards`:

- `Nep26_OnNep11Payment_AcceptsByteArrayTokenId` — a NEP-26 receiver with a `ByteArray` tokenId produces no diagnostic.
- `Nep26_OnNep11Payment_RejectsNonByteArrayTokenId` — a wrong tokenId type is still reported.

Verified locally: `Neo.Compiler.CSharp.UnitTests` (1285) and `Neo.SmartContract.Framework.UnitTests` (253) all pass, and no testing artifacts change.


src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.cs
```diff
@@ -53,11 +53,13 @@ private static System.Collections.Generic.List<CompilationException>
             CheckNep11PayableCompliant(this ContractManifest manifest)
         {
             var onNEP11PaymentMethod = manifest.Abi.GetMethod("onNEP11Payment", 4);
+            // The tokenId is a NEP-11 token id (ByteString), which maps to the ByteArray
```
Wi1l-B0t, line 56: Why `ByteString` maps to `ByteArray`?

ajara87: ContractParameterType (ABI) and StackItemType (VM) use different naming conventions that predate each other. The canonical OnNEP11Payment interface uses ByteString at the C# level which the compiler serializes as ByteArray in the manifest

ajara87 (dismissed):

src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.cs
shargon: It should be ByteString, agree with @Wi1l-B0t

ajara87: but ContractParameterType does not have a ByteString member the enum only has ByteArray (0x12). The compiler's own type mapping in Helper.cs makes this explicit:

`case "Neo.SmartContract.Framework.ByteString": return ContractParameterType.ByteArray;`

shargon: But the abi has string, in Abi level can be

ajara87: yes, but `ByteArray` is the ABI name and ByteString is the VM for same type, different naming from different eras of the codebase. The old checker used String which is a different type (UTF-8 text) and that's exactly what was producing the false positive on every correctly implemented NEP-26 contract

ajara87 (approved):

src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.cs
Jim8y: ContractParameterType does not have ByteString. Framework ByteString is represented as ABI ByteArray, and I added a code comment here to make it clear.

Wi1l-B0t (approved):

Jim8y:
This is now covered by #1751, so closing this duplicate.

PR #1751, opened 2026-06-21
Enforce declared NEP standard compliance at build time

Jim8y:
## Problem

A contract that advertises a NEP standard in its manifest (`supportedstandards`) but does not implement it correctly compiled **successfully**. `CheckStandards` only wrote the violations to stdout and returned the manifest, so the build still exited 0:

```csharp
foreach (CompilationException ex in errors)
    Console.WriteLine(ex.Diagnostic);   // printed, never fails the build
...
return manifest;
```

Wallets, explorers and exchanges trust the `supportedstandards` field, so a contract could ship advertising a standard (e.g. NEP-17/NEP-11) it does not satisfy.

## Fix

Surface the declared-standard violations as compilation errors:

- Replace `CheckStandards` with `GetStandardsViolations`, which returns the collected violations and no longer writes to the console.
- In `CompilationContext.Compile`, when the contract declares at least one standard, add each violation to the diagnostics so `Success` becomes `false` and the compiler exits non-zero. The check is skipped for contracts that declare no standards, so existing behavior is preserved for everything else (including method-less snippets that cannot form a full manifest).

## Tests

- `UnitTest_StandardsEnforcement` (new): a non-compliant declared-standard contract now fails compilation with an `NC3006` error; a compliant one compiles.
- `UnitTest_ManifestStandards`: the helper now asserts on the returned violations instead of captured stdout.
- `Nep11MintUniquenessTest`: its inline contract marks the `symbol` getter `[Safe]` so it satisfies the NEP-11 standard it declares (the only in-repo contract that relied on the old silent behavior).

## Verification

Full local run, all green: `Neo.Compiler.CSharp.UnitTests` (1287), `Neo.SmartContract.Framework.UnitTests` (253), `Template` (41), `Testing` (81), `Analyzer` (162), `Disassembler` (12), and all example projects. No testing artifacts change (codegen is unaffected).

## Dependency

Depends on #1750 (the NEP-26 `onNEP11Payment` tokenId false-positive fix). Without it, the `Contract_SupportedStandard26` test contract would fail the new enforcement. This branch includes that commit for review; it will be rebased onto `master-n3` once #1750 merges.


src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.cs
```diff
@@ -53,11 +53,13 @@ private static System.Collections.Generic.List<CompilationException>
...
         {
             var onNEP11PaymentMethod = manifest.Abi.GetMethod("onNEP11Payment", 4);
+            // The tokenId is a NEP-11 token id (ByteString), which maps to the ByteArray
+            // contract parameter type. See INEP26.OnNEP11Payment.
             var onNEP11PaymentValid = onNEP11PaymentMethod is { ReturnType: ContractParameterType.Void } &&
                                         onNEP11PaymentMethod.Parameters.Length == 4 &&
                                         onNEP11PaymentMethod.Parameters[0].Type == ContractParameterType.Hash160 &&
                                         onNEP11PaymentMethod.Parameters[1].Type == ContractParameterType.Integer &&
-                                        onNEP11PaymentMethod.Parameters[2].Type == ContractParameterType.String &&
+                                        onNEP11PaymentMethod.Parameters[2].Type == ContractParameterType.ByteArray &&
```
Wi1l-B0t, line 62: Why `ContractParameterType.ByteArray`?

ajara87 (dismissed):

ajara87 (approved):

src/Neo.Compiler.CSharp/Manifest/ContractManifestExtension.cs
Jim8y: ContractParameterType does not have ByteString. Framework ByteString is represented as ABI ByteArray, and I added a code comment here to make it clear.

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1685, opened 2026-05-14
Preserve evaluation order for packed expressions

Jim8y:
## Summary
- Preserve C# left-to-right evaluation before emitting PACK/PACKSTRUCT operands in VM stack order.
- Cover array initializers, multidimensional initializers, tuples, anonymous objects, events, and collection expressions.
- Keep constant-only packed expressions on the compact direct emission path.

## Testing
- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter UnitTest_PackExpressionEvaluationOrder --logger "console;verbosity=normal"
- NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --no-restore --filter "Name=PackExpressionsEvaluateElementsLeftToRight|Name=Test_Assign|Name=TestEvent|Name=Initializer_Test|Name=AnonymousObjectCreation_LogsAnonymousMemberValues|Name=AnonymousObjectCreation_CompilesAnonymousGetterHelpersWithValidIndexes|Name=CollectionExpression_SpreadElement_FailsWithDiagnostic" --logger "console;verbosity=normal"
- NEO_SKIP_TEST_COVERAGE=1 dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --no-restore --filter "Name=Test_JaggedArray|Name=Test_JaggedByteArray|Name=Test_IntArrayInit|Name=Test_StructArrayInit|Name=Test_Collectionexpressions|Name=Test_MultiDimensionalInitializer|Name=Test_MultiDimensionalImplicitInitializer|Name=Test_MultiDimensionalByteInitializer" --logger "console;verbosity=normal"

PR #1752, opened 2026-06-21
Preserve evaluation order for object and collection initializers

Jim8y:
## Problem

Two packed-expression sites still evaluated their elements in **reverse source order** — the same defect class fixed for arrays, tuples, anonymous objects, collection expressions and event invocations in #1685, but these two were missed:

1. `TryOptimizedObjectCreation` — the trivial-constructor object-initializer fast path (`new Box { A = ..., B = ..., C = ... }`) emitted field values from the highest field index down to `0`:
   ```csharp
   for (int i = fields.Length - 1; i >= 0; --i) ...
   ```
2. The framework collection-initializer lowering (`new List<T> { a, b, c }`) emitted elements last-to-first:
   ```csharp
   for (var i = initializer.Expressions.Count - 1; i >= 0; i--) ...
   ```

When the initializer values have observable side effects, this evaluated them right-to-left, diverging from C#'s left-to-right guarantee.

## Fix

Route both sites through the existing `EmitPackedItemsLeftToRight` / `CanDeferExpressionEmission` framework, as every other packed site already does.

- The object-initializer path builds its slots in field-index order (value or `PushDefault`) with the optional virtual method table as the final slot, so the **all-constant case emits byte-identical instructions** (no gas change). Side-effect-free but non-constant values (parameter/local loads) now use the forward-emit + stack-reverse path, exactly matching #1685's tradeoff.

## Tests / artifacts

- Added `objectInitializer` and `listInitializer` cases to `UnitTest_PackExpressionEvaluationOrder` (both return `123` for source-order evaluation; the old code returned `321`).
- Refreshed the testing artifacts that legitimately changed (4 compiler test contracts + the NEP-11 template) — bytecode/offset changes only, no ABI signature changes.
- Updated the 4 exact-gas baselines that shifted by the added stack reversals.

## Verification

Full local run, all green: `Neo.Compiler.CSharp.UnitTests` (1283), `Framework` (253), `Template` (41), `Testing` (81), `Disassembler` (12), and all example projects.


Wi1l-B0t:
Conflicts.

Wi1l-B0t (dismissed):

Wi1l-B0t (dismissed):

ajara87 (approved):

shargon (approved):

PR #1755, opened 2026-06-21
Fix: NC4019 false positives on event names and unrelated Notify calls

Jim8y:
## Problem

`NotifyEventNameAnalyzer` (NC4019, **error** severity) reported a build error whenever the first argument of a `Notify` call did not match a **directly-declared** event carrying a `[DisplayName]` attribute:

```csharp
var events = containingType?.Members.OfType<EventFieldDeclarationSyntax>();
// only events WITH [DisplayName] are considered a match
```

This produced build-breaking false positives in three common cases:

1. **Events without `[DisplayName]`** — whose effective name is the field name (e.g. `public static event ... Transfer;`).
2. **Inherited events** — declared in a base class (e.g. a NEP-17 base's `Transfer`), since only the containing type's own members were inspected.
3. **Any method merely named `Notify`** on an unrelated type, since the check was purely syntactic.

## Fix

- Resolve event names the way the compiler does: an event's `[DisplayName]` when present, otherwise the **event field name**.
- Walk the **base-type chain** so inherited events count.
- Only analyze calls that semantically bind to `Neo.SmartContract.Framework.Services.Runtime.Notify`, and only when the event name is a string literal.
- Unknown event names are still reported, so genuine typos remain caught.

## Tests

Rewrote the unit tests (switching to the MSTest verifier already used by the other analyzer tests) to cover: the field-name fallback, an inherited event, a matching `[DisplayName]`, an unrelated `Notify` method, and a real unknown-name typo (still reported).

## Verification

Full local run, all green: `Neo.SmartContract.Analyzer.UnitTests` (166), `Template` (41), and all example projects build with no NC4019 errors.


Wi1l-B0t (dismissed):

src/Neo.SmartContract.Analyzer/NotifyEventNameAnalyzer.cs
```diff
@@ -48,39 +49,60 @@ public override void Initialize(AnalysisContext context)
...
+        }
+
+        private static IEnumerable<string> EnumerateEventNames(INamedTypeSymbol type)
+        {
+            for (INamedTypeSymbol? current = type; current is not null; current = current.BaseType)
+            {
+                foreach (var eventSymbol in current.GetMembers().OfType<IEventSymbol>())
+                {
+                    var displayName = eventSymbol.GetAttributes()
+                        .FirstOrDefault(attr => attr.AttributeClass?.Name == "DisplayNameAttribute")
```
ajara87: could we add namespace too?

ajara87 (dismissed):

src/Neo.SmartContract.Analyzer/NotifyEventNameAnalyzer.cs
```diff
@@ -48,39 +49,60 @@ public override void Initialize(AnalysisContext context)
...
+            if (invocationExpr.Expression is not MemberAccessExpressionSyntax memberAccessExpr)
+                return;
 
-            if (memberAccessExpr?.Name.Identifier.ValueText == "Notify")
-            {
-                if (invocationExpr.ArgumentList?.Arguments.Count > 0)
-                {
-                    var firstArgument = invocationExpr.ArgumentList.Arguments[0].Expression as LiteralExpressionSyntax;
-                    var passedName = firstArgument?.Token.ValueText;
+            if (memberAccessExpr.Name.Identifier.ValueText != "Notify")
```
shargon, line 55: Agree with @ajara87 ,full name should be used

shargon (changes requested):

shargon (approved):

ajara87 (approved):

PR #1757, opened 2026-06-21
Throw a descriptive message for a non-exhaustive switch expression

Jim8y:
## Problem

When no arm of a switch expression matched, the generated code threw the **governing value itself**:

```csharp
AccessSlot(OpCode.LDLOC, anonymousIndex);
AddInstruction(OpCode.THROW);
```

C# throws a `SwitchExpressionException` here, so a contract that caught the exception saw the raw scrutinee value rather than a meaningful message.

## Fix

Throw a constant descriptive string instead, mirroring how `throw new Exception(message)` is lowered. The throw path is only reached when no arm matches, so matched calls are unaffected — their executed opcodes (and therefore gas) are unchanged.

## Tests / artifacts

- `UnitTest_SwitchExpressionThrow`: a matched arm still returns its value; an unmatched value throws with a message containing `did not match` (previously the message was the scrutinee value).
- Refreshed the four testing artifacts whose unmatched fallback changed (bytecode/offset only; ABI signatures and all gas baselines unchanged).

## Verification

Full local run, all green: `Neo.Compiler.CSharp.UnitTests` (1284), `Framework` (253), `Template` (41), and all example projects.


src/Neo.Compiler.CSharp/MethodConvert/Expression/SwitchExpression.cs
```diff
@@ -72,7 +72,9 @@ private void ConvertSwitchExpression(SemanticModel model, SwitchExpressionSyntax
             Jump(OpCode.JMP_L, breakTarget);
             nextTarget.Instruction = AddInstruction(OpCode.NOP);
         }
-        AccessSlot(OpCode.LDLOC, anonymousIndex);
+        // No arm matched. C# throws a SwitchExpressionException here; throw a descriptive
+        // constant message instead of the governing value, so a caught exception is meaningful.
+        Push("The switch expression did not match any pattern.");
```
Wi1l-B0t: Can the exception message be shorter?

ajara87 (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/Expression/SwitchExpression.cs
shargon: What do you propose?

ajara87 (approved):

src/Neo.Compiler.CSharp/MethodConvert/Expression/SwitchExpression.cs
Jim8y: Changed it to "No switch arm matched." and updated the generated artifacts.

Wi1l-B0t (approved):

shargon (approved):

PR #1761, opened 2026-06-22
Simplify duplicated condition in FoldJump

Jim8y:
## Problem

The `FoldJump` jump-threading check tested the same condition twice:

```csharp
if (unconditionalJump.Contains(target.OpCode) || unconditionalJump.Contains(target.OpCode))
```

Both operands are identical, so the disjunction reduces to a single predicate — a copy-paste artifact that obscures intent.

## Fix

Collapse it to `if (unconditionalJump.Contains(target.OpCode))`. Folding a jump whose target is a *conditional* jump is intentionally not performed (see the `FoldJump` remarks), so this is a behavior-preserving cleanup, not a change in optimization scope.

## Verification

No functional change. The full `Neo.Compiler.CSharp.UnitTests` suite passes (1283), the optimizer-specific suites pass (BasicOptimizer, JumpEncoding, PeepholeOptimizer, OptimizerAutomation, SwitchJumpTable), and no testing artifacts change.


shargon (approved):

Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):

PR #1768, opened 2026-06-23
Eliminate dead result-DUP on simple assignment statements

Jim8y:
## Summary

A simple assignment (`a = b`) emits a `DUP` of the assigned value so the value can be left on the stack as the expression result. When such an assignment is used as a *statement*, that result is immediately discarded with a `DROP` — so the `DUP`/`DROP` pair is dead code on **every** assignment statement, which is the overwhelmingly common case.

This removes that dead pair, producing strictly smaller bytecode and lower gas for assignment statements, with no behavioral change.

## How

A `leaveResultOnStack` flag is threaded from `ConvertExpressionStatement` through `ConvertAssignmentExpression` into `ConvertSimpleAssignmentExpression`:

- In **statement context** the flag is `false`: the `DUP` (and, for ref-local rebinds, the trailing `LDLOC`) is not emitted, and `ExpressionStatement` no longer emits the paired `DROP`.
- In **expression context** the flag defaults to `true` and codegen is byte-identical to before.

Only the simple `=` form is affected. Compound (`+=`) and coalesce (`??=`) assignments are untouched and still produce their result.

Each LHS handler already consumes exactly one copy of the assigned value; the `DUP` only ever supplied the *result* copy. Removing it in statement context leaves the stack balanced by construction.

## Testing

- New `UnitTest_AssignmentStatementCodegen` exercises every simple-assignment LHS form as a statement (local, static field, instance field, indexer/element, tuple, discard, declaration) plus expression-context, chained, and compound assignments — a miscompile would surface as a VM fault or wrong value.
- Full compiler, framework, template, testing, analyzer, and disassembler suites pass. Regenerated test artifacts and updated gas baselines (all reductions) reflect the smaller bytecode.


shargon (dismissed):

Wi1l-B0t (approved):
Good.

ajara87 (approved):

shargon (approved):

PR #1769, opened 2026-06-23
Enforce the [Safe] ABI flag against contract-state mutation

Jim8y:
## Summary

The manifest `safe` flag is currently set purely from the presence of the `[Safe]` attribute, with **no verification that the method is actually read-only** (`AbiMethod.cs`: `Safe = GetSafeAttribute(symbol) != null`). Wallets, explorers and dApps treat `safe:true` as "free to call, no signature prompt", so a `[Safe]` method that writes storage advertises a **false on-chain guarantee**.

This is the direct analog of a Solidity `view` function performing an `SSTORE`, which `solc` rejects at compile time. This PR gives the Neo `[Safe]` flag the same enforcement.

## How

A post-codegen verification pass (`CompilationContext.SafeMethodAnalysis.cs`) runs after every method is converted and instructions are finalized. For each exported `[Safe]` method it walks the statically-resolvable **intra-contract call graph** and emits error **`NC3011`** when the method — or any method transitively reachable from it through direct calls — performs a `System.Storage.Put` / `System.Storage.Delete` syscall.

- **Call graph:** recorded as `MethodConvert` call edges at the single in-contract call chokepoint (`EmitCall`), so no fragile bytecode archaeology is needed.
- **Write detection:** read from the emitted `SYSCALL` operands, which also covers writes pulled in by inline expansion (inlined bodies are copied into the caller).
- **Severity:** error, consistent with the existing `[Safe]` setter rejection (`NC3010`).

## Soundness & scope

The check has **no false positives** over what it covers: storage writes are unambiguous state mutations, and every reported path follows a real static call edge. `Runtime.Notify` is intentionally **not** treated as a mutation (emitting an event is not a state change in the safe-to-call sense).

Deliberately **out of scope** (left for a follow-up, as they need interprocedural / cross-contract dataflow):
- writes reachable only through **virtual dispatch**;
- **dynamic** (`CALLA`) invocations;
- writes performed by **another contract** via a cross-contract call carrying the `WriteStates` flag.

## Testing

- New `UnitTest_SafeMethodEnforcement`: direct write, transitive write through a private helper, `Storage.Delete`, read-only Safe method (compiles), and a non-Safe writer (compiles).
- Full compiler (1319), framework (253), template (41), and testing (83) suites pass. All example contracts compile with no `NC3011`.
- The change emits **no bytecode**, so contract artifacts and gas baselines are unaffected (verified: test artifacts regenerate byte-identically; no artifact churn in this PR).

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.SafeMethodAnalysis.cs
+++ b/src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.SafeMethodAnalysis.cs
@@ -1,0 +1,110 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// CompilationContext.SafeMethodAnalysis.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.Compiler.ABI;
+using Neo.SmartContract;
+using Neo.VM;
+using System;
+using System.Collections.Generic;
+
+namespace Neo.Compiler
+{
+    public partial class CompilationContext
+    {
+        // Interop services that mutate persistent contract storage. A method that can reach any
+        // of these is not read-only and therefore cannot honestly carry the [Safe] ABI flag.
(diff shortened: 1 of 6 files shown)


Wi1l-B0t (approved):
Good.

ajara87 (approved):

shargon (commented):
Is not better to show a warning before compile, with an analyzer?

PR #1770, opened 2026-06-23
Add Ownable2Step base for safe two-step ownership transfer

Jim8y:
## Summary

The framework only shipped single-step `Ownable`: `SetOwner` assigns the new owner immediately, with no witness check on the incoming account. Passing a mistyped or otherwise unusable address loses contract control **irrecoverably** — the classic single-step ownership-transfer footgun that OpenZeppelin introduced `Ownable2Step` to fix.

This adds a standalone **`Ownable2Step`** base (the OpenZeppelin `Ownable2Step` analog, adapted to Neo's static-method model). Ownership changes hands only in two steps:

1. the owner proposes a pending owner via `TransferOwnership`, and
2. the transfer completes only when **that account itself** calls `AcceptOwnership` and proves control with `Runtime.CheckWitness`.

A mistyped address can never take ownership because it can never sign the acceptance.

## API

- **`TransferOwnership(newOwner)`** / **`AcceptOwnership()`** — the two-step flow. Re-proposing supersedes any in-flight pending and emits a cancellation event for the dropped candidate.
- **`CancelOwnershipTransfer()`** — owner clears an in-flight pending.
- **`RenounceOwnership()`** — the one deliberate, irreversible action; removes the owner and clears any pending so a previously-proposed account cannot seize an abandoned contract.
- **`GetOwner()` / `GetPendingOwner()`** — `[Safe]` read-only accessors.
- **`InitializeOwner(data, update)`** — call from `_deploy`; defaults to the tx sender, no-op on update.

## Design notes

- **Standalone, not extending `Ownable`** — so it does not re-expose the unsafe single-step `SetOwner` (Neo has no static-method override to hide it).
- **Storage prefixes `0xFD` (owner) / `0xFC` (pending)**, distinct from `Ownable` (`0xFF`) and `Pausable` (`0xFE`).
- **Absent pending = no transfer in flight** is the single source of truth, which collapses double-accept, accept-with-no-pending, and stale-accept-after-renounce to one null check. Storage is written **before** events on every transition.
- `[Safe]` is only on the two pure-read accessors; every state-mutating method is non-`[Safe]` and reaches `CheckWitness`.

## Testing

`Ownable2StepTest` (17 cases) covers the happy path plus every adversarial edge: transfer to self / to zero / by non-owner; accept with no pending / by the wrong account (incl. the current owner) / twice; renounce-then-stale-accept; cancel with nothing pending / by non-owner; post-renounce lockout; re-propose supersession; full event emission; and the manifest safe-flag attribution. Full framework suite green (270).

The design and this implementation were each put through an independent adversarial security review (ownership-theft, bricking, witness-bypass, storage/event ordering, stale-pending seizure) before submission.

### DIFF
--- a/src/Neo.SmartContract.Framework/Ownable2Step.cs
+++ b/src/Neo.SmartContract.Framework/Ownable2Step.cs
@@ -1,0 +1,196 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// Ownable2Step.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.SmartContract.Framework.Attributes;
+using Neo.SmartContract.Framework.Services;
+using System.ComponentModel;
+
+namespace Neo.SmartContract.Framework
+{
+    /// <summary>
+    /// A base contract providing two-step ownership transfer, the OpenZeppelin
+    /// <c>Ownable2Step</c> analog for Neo.
+    /// </summary>
+    /// <remarks>
+    /// Unlike <see cref="Ownable"/>, ownership is never handed over in a single call.
+    /// <see cref="TransferOwnership"/> only records a <em>pending</em> owner; the transfer
+    /// completes only when that account itself calls <see cref="AcceptOwnership"/> and proves
+    /// control via <c>Runtime.CheckWitness</c>. This prevents the classic footgun where a
+    /// mistyped or otherwise unusable address is set as owner in one step and control is lost
+    /// forever. There is intentionally no single-step setter on this class.
+    /// <para>
+    /// <see cref="RenounceOwnership"/> is the one deliberate, irreversible action: it removes the
+    /// owner entirely (and any in-flight pending transfer). After renouncing, every owner-gated
+    /// method is permanently uncallable.
+    /// </para>
+    /// <para>
+    /// Composition note: do not gate <see cref="CancelOwnershipTransfer"/> or
+    /// <see cref="RenounceOwnership"/> behind a pause guard. Both must remain callable so that an
+    /// in-flight transfer to a compromised account can always be neutralized.
+    /// </para>
+    /// </remarks>
+    public abstract class Ownable2Step : SmartContract
+    {
+        // The current confirmed owner. Absent (null) means renounced or not yet initialized.
+        private const byte Prefix_Owner = 0xFD;
+
+        // The proposed owner awaiting acceptance. Absent (null) means no transfer is in flight.
+        private const byte Prefix_PendingOwner = 0xFC;
+
+        /// <summary>
+        /// Returns the current owner, or <see langword="null"/> if ownership has been renounced
+        /// or the contract has not been initialized.
+        /// </summary>
+        [Safe]
+        public static UInt160? GetOwner()
+        {
+            return (UInt160)Storage.Get(new[] { Prefix_Owner })!;
+        }
+
+        /// <summary>
+        /// Returns the account that has been proposed as the new owner and is awaiting its own
+        /// acceptance, or <see langword="null"/> when no transfer is in flight.
+        /// </summary>
+        [Safe]
+        public static UInt160? GetPendingOwner()
+        {
+            return (UInt160)Storage.Get(new[] { Prefix_PendingOwner })!;
+        }
+
+        protected static bool IsOwner()
+        {
+            UInt160? owner = GetOwner();
+            return owner is not null && Runtime.CheckWitness(owner);
+        }
+
+        public delegate void OnOwnershipTransferStartedDelegate(UInt160? previousOwner, UInt160? newOwner);
+
+        /// <summary>
+        /// Raised when an owner proposes a new owner. <c>newOwner</c> is the pending candidate and
+        /// is not yet the owner; the change is not effective until <see cref="AcceptOwnership"/>.
+        /// </summary>
+        [DisplayName("OwnershipTransferStarted")]
+        public static event OnOwnershipTransferStartedDelegate OnOwnershipTransferStarted = null!;
+
+        public delegate void OnOwnershipTransferredDelegate(UInt160? previousOwner, UInt160? newOwner);
+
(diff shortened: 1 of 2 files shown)


src/Neo.SmartContract.Framework/Ownable2Step.cs
```diff
@@ -0,0 +1,198 @@
...
+            UInt160? previousOwner = GetOwner();
+            ExecutionEngine.Assert(previousOwner != newOwner, "owner must change");
+
+            // A new proposal supersedes any earlier one; emit a terminal event for the dropped
+            // candidate so off-chain indexers never see a pending offer left dangling.
+            UInt160? oldPending = (UInt160)Storage.Get(new[] { Prefix_PendingOwner })!;
+            if (oldPending is not null)
+                OnOwnershipTransferCanceled(previousOwner, oldPending);
+
+            Storage.Put(new[] { Prefix_PendingOwner }, newOwner);
```
shargon: I think we have storage methods with byte prefix  without the need of create an array in all calls for the key

ajara87 (dismissed):

src/Neo.SmartContract.Framework/Ownable2Step.cs
```diff
@@ -0,0 +1,201 @@
...
+                throw new System.InvalidOperationException("No Authorization!");
+
+            ExecutionEngine.Assert(newOwner.IsValidAndNotZero, "owner must be valid");
+
+            UInt160? previousOwner = GetOwner();
+            ExecutionEngine.Assert(previousOwner != newOwner, "owner must change");
+
+            // A new proposal supersedes any earlier one; emit a terminal event for the dropped
+            // candidate so off-chain indexers never see a pending offer left dangling.
+            UInt160? oldPending = (UInt160)PendingOwnerStorage.Get(ByteString.Empty)!;
```
shargon: Must be cached the storage map when is used multiple times in the same method

Jim8y: Addressed. The storage map is now cached within the methods that use it more than once.

src/Neo.SmartContract.Framework/Ownable2Step.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// <summary>
+        /// Step 1 of an ownership transfer. The current owner proposes <paramref name="newOwner"/>;
+        /// the owner does not change until that account calls <see cref="AcceptOwnership"/>.
+        /// </summary>
+        public static void TransferOwnership(UInt160 newOwner)
+        {
+            if (!IsOwner())
+                throw new System.InvalidOperationException("No Authorization!");
+
+            var pendingOwnerStorage = PendingOwnerStorage;
```
shargon: Move declaration after asserts

shargon (dismissed):

Wi1l-B0t (dismissed):

ajara87 (dismissed):

src/Neo.SmartContract.Framework/Ownable2Step.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        /// <summary>
+        /// Initializes the owner. Call this from the contract's <c>_deploy</c> method.
+        /// </summary>
+        /// <remarks>
+        /// On a contract update this is a no-op and never resets the owner. Note that an in-flight
+        /// pending transfer is intentionally left untouched across an update; operators should
+        /// cancel any in-flight transfer before upgrading if the upgrade changes trust assumptions.
+        /// </remarks>
+        protected static void InitializeOwner(object? data, bool update)
```
shargon, line 198: Check It's owner if is not defined?

Jim8y: Addressed. InitializeOwner now keeps the existing owner during update, but initializes one if no owner is stored. I also added a regression test for that case.

ajara87 (dismissed):

Jim8y:
Updated with latest master. Could you please take another look?

Wi1l-B0t (dismissed):

src/Neo.SmartContract.Framework/Ownable2Step.cs
Jim8y: Could you please confirm if this works for you?

src/Neo.SmartContract.Framework/Ownable2Step.cs
```diff
@@ -0,0 +1,206 @@
...
+        /// <remarks>
+        /// On a contract update this preserves the current owner. If no owner is stored yet, it
+        /// initializes one so contracts adopting <see cref="Ownable2Step"/> during an upgrade do not
+        /// remain ownerless. Note that an in-flight pending transfer is intentionally left untouched
+        /// across an update; operators should cancel any in-flight transfer before upgrading if the
+        /// upgrade changes trust assumptions.
+        /// </remarks>
+        protected static void InitializeOwner(object? data, bool update)
+        {
+            if (update && GetOwner() is not null)
```
shargon: Update Or has owner? If owner was renounced can be used during update

Jim8y: Fixed. Update now uses a separate initialized marker, so a renounced owner stays renounced across updates while an old contract with no owner state can still initialize on first upgrade.

src/Neo.SmartContract.Framework/Ownable2Step.cs
Jim8y: This path is still preserved. On update we keep an existing owner, and only initialize when neither an owner nor the initialized marker exists.

src/Neo.SmartContract.Framework/Ownable2Step.cs
```diff
@@ -0,0 +1,221 @@
...
+        protected static void InitializeOwner(object? data, bool update)
+        {
+            if (update)
+            {
+                if (OwnerInitializedStorage.GetBoolean(ByteString.Empty))
+                    return;
+
+                if (GetOwner() is not null)
+                {
+                    OwnerInitializedStorage.Put(ByteString.Empty, true);
```
shargon: is not better a throw?

Jim8y: Agreed, updated it to abort instead of accepting inconsistent state. Added a regression test as well.

shargon (approved):

ajara87 (approved):

PR #1771, opened 2026-06-23
Add AccessControl base for role-based authorization

Jim8y:
## Summary

The framework only shipped single-owner `Ownable`, so every contract that needs more than one privileged account (a minter, a pauser, a fee setter) hand-rolls its own unaudited authorization. This adds **`AccessControl`**, the OpenZeppelin `AccessControl` analog adapted to Neo's static-method model — the single most-used safety primitive in the OZ library.

A role is a 32-byte id; accounts are granted roles, and each role is administered by another role. `DEFAULT_ADMIN_ROLE` (32 zero bytes) is the root role and is its own admin; the account it is granted to at deployment can administer every other role.

## Security properties

- **Explicit-actor authorization.** Callers pass the acting account and the contract verifies it both holds the required role **and** witnessed the call (`Runtime.CheckWitness`), honoring the signer's witness scope. This avoids the confused-deputy hole of deriving the actor from `Transaction.Sender`, and lets a contract/multisig hold a role.
- **No key aliasing.** Roles are a fixed 32 bytes, validated on every entry point, so the membership key `tag ++ role ++ account` is structurally injective — no two distinct `(role, account)` pairs can collide on one storage cell.
- **No single-call brick.** The last `DEFAULT_ADMIN_ROLE` holder can never be removed (guarded on **both** revoke and renounce), so the contract can never be left un-administrable from one call; root control is handed over by granting to a successor first.
- **No init replay.** A one-time initialization flag stops a deploy-time re-init from adding a co-admin.
- **Idempotent + event-accurate.** Grant/revoke/admin-change are idempotent and emit events only on an actual state change; a maintained per-role member count backs the last-admin guard.

## API

`DEFAULT_ADMIN_ROLE` / `HasRole` / `GetRoleAdmin` / `GetRoleMemberCount` (`[Safe]` reads); `GrantRole(role, admin, account)` / `RevokeRole(role, admin, account)` / `RenounceRole(role, account)`; the `OnlyRole(role, account)` guard authors call at the top of their own role-gated methods; and protected `GrantRoleInternal` / `RevokeRoleInternal` / `SetRoleAdminInternal` cores that are **not** exported to the ABI. Storage uses reserved prefix `0xFB`.

No `[OnlyRole]` modifier attribute is shipped: a Neo `ModifierAttribute.Enter()` cannot read the decorated method's runtime arguments, so it could only gate a compile-time-constant role against the wrong actor model. The `OnlyRole` guard is the runtime-flexible substitute.

## Testing

`AccessControlTest` (24 cases): bootstrap and replay protection, the **confused-deputy attack** (admin without a witness for this call), grant/revoke/renounce by the wrong party, idempotency, the **last-admin brick on both revoke and renounce**, safe admin handoff, the admin-role **hierarchy** (re-pointing a role's admin), malformed-role faults, and the manifest safe-flag + ABI-exclusion surface (protected cores absent from the ABI). Full framework suite green (277).

The design and this implementation were each put through an independent multi-design + adversarial security-audit pass (privilege escalation, bricking, key aliasing, witness bypass, count desync) before submission.

### DIFF
--- a/src/Neo.SmartContract.Framework/AccessControl.cs
+++ b/src/Neo.SmartContract.Framework/AccessControl.cs
@@ -23,10 +23,10 @@ namespace Neo.SmartContract.Framework
     /// administered by another role.
     /// </summary>
     /// <remarks>
-    /// A role is identified by a 32-byte id. Derive named roles deterministically, e.g.
-    /// <c>CryptoLib.Sha256((ByteString)"MINTER_ROLE")</c>. <see cref="DEFAULT_ADMIN_ROLE"/> (32
-    /// zero bytes) is the root role and is its own admin; the account it is granted to at
-    /// deployment can administer every other role.
+    /// A role is identified by a compact 4-byte key. Define role enums in your contract and derive
+    /// keys with <see cref="RoleKey"/>, e.g. <c>RoleKey((int)Roles.Minter)</c>.
+    /// <see cref="DEFAULT_ADMIN_ROLE"/> (four zero bytes) is the root role and is its own admin;
+    /// the account it is granted to at deployment can administer every other role.
     /// <para>
     /// Authorization is by explicit actor: callers pass the acting account and the contract
     /// verifies it both holds the required role and witnessed the call
@@ -53,11 +53,11 @@ namespace Neo.SmartContract.Framework
     {
         private const byte Prefix = 0xFB;
 
-        // Sub-namespace tags. 0x00 is reserved for the one-time init flag only; the variable-length
-        // membership/admin/count keys use non-zero tags so they can never collide with it.
-        private const byte TAG_MEMBER = 0x01;  // [role:32][account:20] -> granted
-        private const byte TAG_ADMIN = 0x02;   // [role:32]             -> admin role override
-        private const byte TAG_COUNT = 0x03;   // [role:32]             -> member count
+        // Sub-namespace tags. 0x00 is reserved for the one-time init flag only; the role keys use
+        // non-zero tags so they can never collide with it.
+        private const byte TAG_MEMBER = 0x01;  // [role:4][account:20] -> granted
+        private const byte TAG_ADMIN = 0x02;   // [role:4]             -> admin role override
+        private const byte TAG_COUNT = 0x03;   // [role:4]             -> member count
 
         // The init flag lives under tag 0x00 (a fixed single-byte key) so it can never collide with
         // the non-zero tagged variable-length role keys.
@@ -74,17 +74,26 @@ namespace Neo.SmartContract.Framework
         private static byte[] CountKey(ByteString role)
             => new byte[] { TAG_COUNT }.Concat((byte[])role);
 
-        // Roles are fixed 32 bytes so that the flat (role ++ account) key is injective: with both
-        // fields fixed-width, two distinct (role, account) pairs can never alias to one key.
+        /// <summary>
+        /// Converts an enum-style integer role id to the fixed 4-byte role key used by storage.
+        /// </summary>
+        protected static ByteString RoleKey(int role)
+        {
+            ExecutionEngine.Assert(role >= 0, "AccessControl: role must be non-negative");
+            return (ByteString)new byte[] { (byte)role, (byte)(role >> 8), (byte)(role >> 16), (byte)(role >> 24) };
+        }
+
+        // Roles are fixed 4 bytes so that the flat (role ++ account) key is injective with lower
+        // storage overhead than 32-byte hashes.
         private static void ValidateRole(ByteString role)
-            => ExecutionEngine.Assert(role.Length == 32, "AccessControl: role must be 32 bytes");
+            => ExecutionEngine.Assert(role.Length == 4, "AccessControl: role must be 4 bytes");
 
         /// <summary>
-        /// The root admin role (32 zero bytes). It is its own admin, and the account granted this
+        /// The root admin role (four zero bytes). It is its own admin, and the account granted this
         /// role at deployment can administer every other role.
         /// </summary>
         [Safe]
-        public static ByteString DEFAULT_ADMIN_ROLE() => (ByteString)new byte[32];
+        public static ByteString DEFAULT_ADMIN_ROLE() => RoleKey(0);
 
         /// <summary>
         /// Returns whether <paramref name="account"/> has been granted <paramref name="role"/>.
--- a/src/Neo.SmartContract.Framework/AccessControl.cs
+++ b/src/Neo.SmartContract.Framework/AccessControl.cs
@@ -1,0 +1,282 @@
+// Copyright (C) 2015-2026 The Neo Project.
+//
+// AccessControl.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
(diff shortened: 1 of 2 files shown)


src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// <summary>
+        /// Asserts that <paramref name="account"/> holds <paramref name="role"/> and witnessed the
+        /// current invocation. This is the authorization core; it aborts otherwise.
+        /// </summary>
+        protected static void CheckRole(ByteString role, UInt160 account)
+        {
+            ValidateRole(role);
+            ExecutionEngine.Assert(account.IsValidAndNotZero, "AccessControl: invalid account");
+            ExecutionEngine.Assert(HasRole(role, account) && Runtime.CheckWitness(account),
+                "AccessControl: account is missing role");
```
ajara87:
```suggestion
            ExecutionEngine.Assert(HasRole(role, account), "AccessControl: account is missing role");
            ExecutionEngine.Assert(Runtime.CheckWitness(account), "AccessControl: missing witness for account");
```

ajara87 (dismissed):
lgtm

ajara87 (dismissed):

src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,282 @@
...
+            ValidateRole(role);
+            return Map.Get(MemberKey(role, account)) is not null;
+        }
+
+        /// <summary>
+        /// Returns the admin role that controls granting and revoking <paramref name="role"/>.
+        /// Defaults to <see cref="DEFAULT_ADMIN_ROLE"/> when no override has been set.
+        /// </summary>
+        [Safe]
+        public static ByteString GetRoleAdmin(ByteString role)
```
Wi1l-B0t: Only one Admin?

Jim8y: It is one admin role id per role, not one admin account. Multiple accounts can hold that admin role. I clarified this in the docs.

Jim8y:
Updated. CI is green and all review threads are resolved.

src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,284 @@
...
+        // The init flag lives under tag 0x00 (a fixed single-byte key) so it can never collide with
+        // the non-zero tagged variable-length role keys.
+        private const byte TAG_INIT = 0x00;
+
+        private static StorageMap Map => new(Storage.CurrentContext, Prefix);
+
+        private static byte[] MemberKey(ByteString role, UInt160 account)
+            => new byte[] { TAG_MEMBER }.Concat((byte[])role).Concat((byte[])account);
+
+        private static byte[] AdminKey(ByteString role)
```
shargon: Maybe we can use TEnum for roles, convert the enum to key is only 4 bytes

Jim8y: I kept the core role id as 32 bytes intentionally: it matches the usual named-role pattern, stays stable across contract versions, and keeps the storage keys fixed-width and collision-free. Enum-based helpers could be added later as convenience overloads, but I would not make the base storage format enum-only.

shargon: I think there is no need to store more data, 32 bytes for key cost more than 4 bytes

Wi1l-B0t (dismissed):

src/Neo.SmartContract.Framework/AccessControl.cs
Jim8y: Good point. I changed the role id to a compact 4-byte key and added RoleKey(int) so enum values can be used without storing 32-byte hashes. Added coverage for that path too.

src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,293 @@
...
+        public delegate void OnRoleAdminChangedDelegate(ByteString role, ByteString previousAdminRole, ByteString newAdminRole);
+
+        [DisplayName("RoleAdminChanged")]
+        public static event OnRoleAdminChangedDelegate OnRoleAdminChanged = null!;
+
+        /// <summary>
+        /// Asserts that <paramref name="account"/> holds <paramref name="role"/> and witnessed the
+        /// current invocation. This is the authorization core; it aborts otherwise.
+        /// </summary>
+        protected static void CheckRole(ByteString role, UInt160 account)
```
shargon: Why don't use Integers?

Jim8y: Addressed. Roles now use integer identifiers directly, with storage keys derived from the integer value.

ajara87 (dismissed):

Jim8y:
Updated with latest master-n3.

src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// Renounces <paramref name="role"/> for the caller. Only <paramref name="account"/> itself
+        /// (verified by witness) may renounce its own role. A no-op if not held.
+        /// </summary>
+        public static void RenounceRole(BigInteger role, UInt160 account)
+        {
+            ValidateRole(role);
+            ExecutionEngine.Assert(account.IsValidAndNotZero, "AccessControl: invalid account");
+            ExecutionEngine.Assert(Runtime.CheckWitness(account), "AccessControl: can only renounce roles for self");
+            if (!HasRole(role, account))
+                return;
```
shargon: exception?

Jim8y: Updated.

ajara87 (dismissed):

src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,284 @@
...
+            OnRoleAdminChanged(role, previous, adminRole);
+        }
+
+        /// <summary>
+        /// Bootstraps access control by granting <see cref="DEFAULT_ADMIN_ROLE"/> to the initial
+        /// admin. Call this from the contract's <c>_deploy</c> method. Defaults the admin to the
+        /// transaction sender when <paramref name="data"/> is null; a no-op on update. Can only run
+        /// once.
+        /// </summary>
+        protected static void InitializeAccessControl(object? data, bool update)
```
shargon: UINT instead of object?

src/Neo.SmartContract.Framework/AccessControl.cs
```diff
@@ -0,0 +1,282 @@
...
+            OnRoleGranted(role, account, sender);
+        }
+
+        /// <summary>
+        /// Unconditionally revokes <paramref name="role"/> from <paramref name="account"/> (no admin
+        /// or last-admin check). Idempotent. Not exported; the public callers apply the guards.
+        /// </summary>
+        protected static void RevokeRoleInternal(BigInteger role, UInt160 account, UInt160 sender)
+        {
+            ValidateRole(role);
```
ajara87, line 231:
```suggestion
            ValidateRole(role);
            GuardLastAdmin();
```
should we not add `GuardLastAdmin`?

shargon: It's peotected so people should decide if they don't want to preserve the owner ,but summary should be more descriptive

Jim8y: Updated the summary to make this explicit.

ajara87 (approved):

shargon (approved):

PR #1772, opened 2026-06-23
Add PausableOwnable for an owner-gated pause switch

Jim8y:
## Summary

`Pausable` ships `Pause`/`Unpause` as `protected` methods with **no access control**, and no documentation of that fact. An author who exposes them through a public method without adding an owner check creates a contract where **any caller can pause it**, denying service to every guarded method.

This adds **`PausableOwnable`**, the safe-by-default circuit breaker: it combines `Ownable` with a paused flag so `Pause` and `Unpause` are public but require the owner's witness. It is the form the audit recommends (a `PausableOwnable`, or making `Pause` take an owner check).

## Behavior

- `Pause()` / `Unpause()` — public, owner-gated (`IsOwner()`), assert the paused-state precondition, write storage **before** emitting `Paused`/`Unpaused(owner)`. Not `[Safe]`.
- `Paused` — `[Safe]` pure read.
- `WhenNotPaused()` / `WhenPaused()` — `protected` guards authors call at the top of their own methods.
- Storage `0xFE` (paused) + inherited `0xFF` (owner) do not collide.

It also documents on `Pausable` that its `Pause`/`Unpause` are unguarded and must be wrapped with an authorization check, pointing readers to `PausableOwnable` for the owner-gated default.

## Testing

`PausableOwnableTest` (9 cases): owner can pause/unpause with events, a non-owner cannot pause or unpause, pause-when-paused and unpause-when-not-paused abort, the `WhenNotPaused` and `WhenPaused` guards gate business methods, and the manifest safe-flag + ABI-exclusion surface (protected guards absent). Full framework suite green.


src/Neo.SmartContract.Framework/PausableOwnable.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            ExecutionEngine.Assert(Paused, "contract is not paused");
+            Storage.Delete(new[] { Prefix_Paused });
+            OnUnpaused(GetOwner());
+        }
+
+        /// <summary>
+        /// Guard for an author's method that must only run while the contract is not paused.
+        /// </summary>
+        protected static void WhenNotPaused() => ExecutionEngine.Assert(!Paused, "contract is paused");
```
shargon: Create Attributes are more friendly

ajara87 (approved):

Wi1l-B0t (approved):

src/Neo.SmartContract.Framework/PausableOwnable.cs
Jim8y: Addressed earlier. PausableOwnable now uses the pause modifier attributes.

shargon (approved):

PR #1773, opened 2026-06-23
Flag storage prefixes that collide with an inherited base's reserved prefix

Jim8y:
## Summary

`NC4056` (`StorageKeyCollisionAnalyzer`) only compared a contract's own `StorageMap`/`LocalStorageMap` prefixes against each other, so it missed the more common footgun: a contract that **inherits `Ownable` or `Pausable`** and then builds a `StorageMap` with that base class's reserved prefix.

`Ownable` stores its owner under `0xFF` and `Pausable` its paused flag under `0xFE`. A derived contract that reuses those prefixes silently shares the base's storage namespace — corrupting the stored owner or paused state — with no diagnostic today.

## How

Before scanning the contract's own `StorageMap` fields, seed the per-class prefix set with the reserved prefixes of any inherited framework base (`Ownable` → `0xFF`, `Pausable` → `0xFE`). The existing collision check then fires when a derived contract reuses one. This **reuses the existing `NC4056` rule** (no new diagnostic id); the message names the base class whose prefix is being reused.

A contract that does **not** inherit the base is unaffected and may use any prefix, so there are no false positives.

## Testing

Adds positive tests for both reserved prefixes and negatives: a non-inheriting contract using `0xFF` (no warning) and an inheriting contract using a different prefix (no warning). Full analyzer suite green (171).


Wi1l-B0t (dismissed):
Simple Patch.

ajara87 (dismissed):

shargon (commented):
I love it, could you add also Tokens nep17 and 11?

Jim8y:
Added the reserved prefixes for Ownable2Step, PausableOwnable, AccessControl, and RoyaltyNep11Token too.

ajara87 (approved):

shargon (approved):

src/Neo.SmartContract.Analyzer/StorageKeyCollisionAnalyzer.cs
```diff
@@ -35,7 +35,23 @@ public sealed class StorageKeyCollisionAnalyzer : DiagnosticAnalyzer
...
+        private static readonly Dictionary<string, byte[]> ReservedBasePrefixes = new(StringComparer.Ordinal)
+        {
+            ["global::Neo.SmartContract.Framework.Ownable"] = [0xFF],
+            ["global::Neo.SmartContract.Framework.Ownable2Step"] = [0xFD, 0xFC],
+            ["global::Neo.SmartContract.Framework.Pausable"] = [0xFE],
+            ["global::Neo.SmartContract.Framework.PausableOwnable"] = [0xFE],
+            ["global::Neo.SmartContract.Framework.AccessControl"] = [0xFB],
+            ["global::Neo.SmartContract.Framework.TokenContract"] = [0x00, 0x01],
+            ["global::Neo.SmartContract.Framework.Nep17Token"] = [0x00, 0x01],
+            ["global::Neo.SmartContract.Framework.Nep11Token`1"] = [0x02, 0x03, 0x04],
```
Wi1l-B0t, line 52: Why 'Nep11Token`1', not  'Nep11Token' ?

Jim8y: It uses Roslyn's MetadataName. Generic types include their arity there, so Nep11Token<T> is Nep11Token`1.

Wi1l-B0t (approved):

ajara87:
@shargon  could you merge?

PR #1774, opened 2026-06-23
Add RoyaltyNep11Token NEP-24 royalty base for NFTs

Jim8y:
## Summary

The framework already had an `INep24` interface and a NEP-24 manifest compliance check, but **no base class implementing the royalty standard** — so every NFT that wanted on-chain royalties hand-rolled `royaltyInfo` and its storage.

This adds **`RoyaltyNep11Token`**, a NEP-11 token base that also implements **NEP-24** (the Neo analog of EIP-2981).

## Behavior

- Royalties are expressed in **basis points** of the sale price (denominator `10000` = 100%).
- A contract-wide **default** royalty can be **overridden per token**; the per-token value takes precedence.
- `royaltyInfo(tokenId, royaltyToken, salePrice)` ( `[Safe]`, returns `Array` of maps) computes `royaltyAmount = salePrice * basisPoints / 10000` in the sale's settlement asset and returns `{ royaltyRecipient, royaltyAmount }`. It **reverts on a nonexistent token** and returns an **empty array** when no royalty is configured.
- The royalty setters (`SetDefaultRoyalty` / `SetTokenRoyalty` and their deleters) are `protected` with **no access control of their own** — wrap them behind your contract's authorization exactly as you already gate `Mint`.

Storage uses prefixes `0x05`/`0x06`, distinct from the NEP-11 base (`0x02`–`0x04`) and `TokenContract` (`0x00`–`0x01`). Only `royaltyInfo` is `[Safe]` and it performs no writes.

## Testing

`RoyaltyNep11TokenTest` (9 cases): default-royalty computation incl. floor division, per-token override, empty result when unset, revert on a nonexistent token, basis-points and recipient validation, the two delete paths, and NEP-24 manifest conformance (`royaltyInfo` Safe, `Array` return, `ByteArray`/`Hash160`/`Integer` parameters, `NEP-24` in `SupportedStandards`). The contract compiling with `[SupportedStandards(Nep11, Nep24)]` proves both standards' compliance checks pass. Full framework suite green (262).

Implementation reviewed via an independent adversarial pass (royalty math, NEP-24 ABI shape, storage-key consistency, `[Safe]` purity, validation) before submission.


src/Neo.SmartContract.Framework/RoyaltyNep11Token.cs
```diff
@@ -0,0 +1,128 @@
...
+        protected static void DeleteTokenRoyalty(ByteString tokenId)
+        {
+            new StorageMap(Prefix_TokenRoyalty).Delete(tokenId);
+        }
+
+        private static ByteString SerializeRoyalty(UInt160 recipient, BigInteger basisPoints)
+        {
+            ExecutionEngine.Assert(recipient.IsValidAndNotZero, "The royalty recipient is invalid.");
+            ExecutionEngine.Assert(basisPoints >= 0 && basisPoints <= RoyaltyDenominator,
+                "The royalty basis points must be between 0 and 10000.");
```
Wi1l-B0t:
```suggestion
                "The royalty basis points must be between [0, 10000].");
```
Inclusive.

Wi1l-B0t (dismissed):

src/Neo.SmartContract.Framework/RoyaltyNep11Token.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /// <summary>
+        /// Sets the contract-wide default royalty paid to <paramref name="recipient"/>, expressed
+        /// as <paramref name="basisPoints"/> of the sale price. No access control; gate this behind
+        /// your own authorization.
+        /// </summary>
+        protected static void SetDefaultRoyalty(UInt160 recipient, BigInteger basisPoints)
+        {
+            Storage.Put(new[] { Prefix_DefaultRoyalty }, SerializeRoyalty(recipient, basisPoints));
```
shargon: Use storage without key arrays, we have byte prefix

src/Neo.SmartContract.Framework/RoyaltyNep11Token.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// configured for the token.
+        /// </summary>
+        [Safe]
+        public static Map<string, object>[] RoyaltyInfo(ByteString tokenId, UInt160 royaltyToken, BigInteger salePrice)
+        {
+            // Reverts when the token does not exist.
+            OwnerOf(tokenId);
+            ExecutionEngine.Assert(salePrice >= 0, "The argument \"salePrice\" must be non-negative.");
+
+            ByteString? data = new StorageMap(Prefix_TokenRoyalty)[tokenId]
```
shargon: Do the same with this prefix? A property

Jim8y: Addressed. Token royalty storage now uses a property too.

ajara87 (dismissed):

src/Neo.SmartContract.Framework/RoyaltyNep11Token.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /// <summary>
+        /// Sets a per-token royalty for <paramref name="tokenId"/>, overriding the default. No
+        /// access control; gate this behind your own authorization.
+        /// </summary>
+        protected static void SetTokenRoyalty(ByteString tokenId, UInt160 recipient, BigInteger basisPoints)
+        {
+            // Reverts when the token does not exist.
+            OwnerOf(tokenId);
```
shargon, line 111: CheckWitness?

Jim8y: This is intentionally not hard-coded here. The method is protected; the derived contract should gate it with its own policy, such as owner, creator, minter, or role-based authorization.

Jim8y:
Updated with latest master. Could you please take another look?

src/Neo.SmartContract.Framework/RoyaltyNep11Token.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// </summary>
+        [Safe]
+        public static Map<string, object>[] RoyaltyInfo(ByteString tokenId, UInt160 royaltyToken, BigInteger salePrice)
+        {
+            // Reverts when the token does not exist.
+            OwnerOf(tokenId);
+            ExecutionEngine.Assert(salePrice >= 0, "The argument \"salePrice\" must be non-negative.");
+
+            ByteString? data = TokenRoyaltyStorage[tokenId]
+                ?? DefaultRoyaltyStorage.Get(ByteString.Empty);
```
Wi1l-B0t, line 69: Why `ByteString.Empty`?

Jim8y: It is the singleton key for the default royalty under the byte-prefix map. I added a comment to make that explicit.

src/Neo.SmartContract.Framework/RoyaltyNep11Token.cs
Jim8y: Could you please confirm if this works for you?

Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

Issue #1776, opened 2026-06-23
First-party source-level (DAP) debugger for Neo C# contracts


Jim8y:
## Summary

Neo smart-contract developers have **no first-party source-level debugger**. Today, debugging a C# contract means `Runtime.Log` tracing, writing unit tests against the testing engine, or reaching for the external NEO Blockchain Toolkit. There is no built-in way to set a breakpoint in your `.cs` file, step through it line by line, and inspect locals — the experience every other modern toolchain ships.

This issue proposes adding a **first-party Debug Adapter Protocol (DAP) debugger** to `neo-devpack-dotnet`, built on infrastructure the repo already has, and tracks the work as a series of small, self-contained PRs.

## Why now — the two hardest pieces already exist

A source-level debugger needs two things that are notoriously expensive to build. Both are already in-repo and production-grade:

1. **A source map.** The compiler emits NEP-19 debug info (`CompilationContext.CreateDebugInformation`), and `NeoDebugInfo` (in `Neo.SmartContract.Testing`) parses it and already does the address → `file:line:col` reverse lookup (`GetSourceLocation`).
2. **A per-instruction execution hook.** `TestingApplicationEngine` is already an `ApplicationEngine` subclass that overrides `PreExecuteInstruction` / `PostExecuteInstruction`. The coverage instrumentation proves these fire reliably on every opcode with full access to the execution context, evaluation stack, call stack, and local/argument/static slots.

A code-verified feasibility study (against Neo.VM 3.10.0) concluded this is **practical**.

### Key technical finding

Neo.VM's `ExecuteNext()` and `State` setter are `internal`, so you **cannot** drive the VM one instruction at a time from outside. You don't need to: because the engine already exposes a `protected` pre-instruction hook, execution can be **paused by blocking inside that hook** on a synchronization primitive and **resumed by signalling it from the debug-adapter thread**. The entire stepping model is "subclass + thread handshake" — **no fork of Neo.VM and no upstream change required.**

## Architecture

- **New library `src/Neo.SmartContract.Debugging/`** — the DAP server plus a `DebuggableApplicationEngine : TestingApplicationEngine` execution backend.
- **A thin stdio host exe** speaking DAP (JSON-RPC over stdio) via the standard `Microsoft.VisualStudio.Shared.VSCodeDebugProtocol` library.
- **A minimal VS Code extension** contributing a `neo-csharp` debug type + `launch.json` schema.

[...]

PR #1775, opened 2026-06-23
[debugger 1] Add source breakpoint resolver

Jim8y:
## Summary

The compiler already emits a NEP-19 source map (`debug.json`) and the testing engine already exposes a per-instruction execution hook, but there is no **first-party source-level debugger**. This is the first, self-contained building block: a new **`Neo.SmartContract.Debugging`** library with a breakpoint resolver.

`BreakpointResolver.ResolveBreakpoint(file, line)` is the **inverse of `NeoDebugInfo.GetSourceLocation`** — it maps a source location to the instruction address a debugger should stop at:

- A requested line binds to the **first instruction of the nearest executable line at or after it**, so a breakpoint on a blank/comment line snaps to the next statement (matching how editors bind breakpoints), and the result reports the line/column it actually bound to.
- Files are matched against the debug-info documents by **full path first, then file name**, so a breakpoint still binds when the debugger and the debug info disagree on the absolute path (relocated sources, different roots).
- Returns `null` when the file is unknown or no executable code exists at or after the requested line.

This is the breakpoint-binding primitive a Debug Adapter (DAP) will use; it has **no execution or protocol dependencies yet** — a deliberately small first slice.

## Context

This is **phase 1, step 1** of a planned first-party DAP debugger (a feasibility study mapped the existing in-repo assets — source map, per-instruction hook on `TestingApplicationEngine` — against DAP needs). Subsequent steps (pausable debug engine, DAP server, stepping, variables) build on this primitive and will land as their own PRs. There are open product questions for maintainers (test-mode scope, alignment with the existing Blockchain Toolkit debugger, debug-build requirement) — this PR commits to none of them; it only adds a pure, well-tested mapping utility.

## Testing

`BreakpointResolverTests` (9 cases): exact-line resolution, multiple sequence points per line (lowest address wins), binding to the next executable line, before-first-line and beyond-all-code, file-name fallback matching, unknown file, multi-method/multi-document resolution, and argument validation. Full solution builds with the new library and test project wired into the `.sln`.

PR #1777, opened 2026-06-23
[debugger 2] Add a per-instruction execution hook to the testing engine

Jim8y:
## Summary

A source-level debugger needs to **observe execution one instruction at a time and pause it**. Neo.VM's single-step `ExecuteNext()` and `State` setter are `internal`, so the VM cannot be driven step-by-step from outside — but `TestingApplicationEngine` already overrides `PreExecuteInstruction`, which is a reliable per-instruction interception point (the same hook the coverage instrumentation uses).

This exposes that point as an opt-in **`TestEngine.OnPreExecuteInstruction`** event, raised before each instruction executes (after the base gas/limit checks), with the engine and the instruction.

- **Opt-in / zero-overhead when unused:** the hook only fires when something is subscribed, so normal test runs are unaffected.
- **Pause by blocking:** because the handler runs on the execution thread, a debugger can pause the virtual machine simply by blocking inside the handler and resume it by returning. This is the foundation the debug adapter builds on, with **no fork of Neo.VM required**.

## Context

This is **phase 1, step 2** of the first-party DAP debugger tracked in #1776 (step 1 — the breakpoint resolver — is #1775). The next step wires this hook together with the resolver into a `DebugSession` (pause/resume on a worker thread, breakpoints) in the `Neo.SmartContract.Debugging` project.

## Testing

`DebugHookTests` (3 cases): the hook fires for every executed instruction in address order; a **blocking handler pauses execution on a worker thread and returning resumes it to completion** (the pause/resume mechanism the debugger relies on); and execution is unchanged when nothing is subscribed. Full `Neo.SmartContract.Testing` suite green (86).

### DIFF
--- a/src/Neo.SmartContract.Testing/TestEngine.cs
+++ b/src/Neo.SmartContract.Testing/TestEngine.cs
@@ -51,6 +51,20 @@ namespace Neo.SmartContract.Testing
         public delegate void OnRuntimeNotifyDelegate(UInt160 sender, string eventName, Neo.VM.Types.Array state);
         public event OnRuntimeNotifyDelegate? OnRuntimeNotify;
 
+        public delegate void OnPreExecuteInstructionDelegate(ApplicationEngine engine, Instruction instruction);
+
+        /// <summary>
+        /// Raised before each instruction is executed during <see cref="Execute"/>. A debugger can
+        /// observe execution through this hook; because the handler runs on the execution thread,
+        /// blocking inside it pauses the virtual machine until the handler returns.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


Jim8y:
Closing — relocating the debugger to neo-project/neo-express (see #1776). The neo-express debugger will hook its own engine / replay neo-trace files, so this in-process TestEngine hook isn't the path forward. The "Avoid partial coverage on debug hook raise" commit is preserved on the branch; if a per-instruction observation hook on the testing engine is wanted as a standalone feature later, it can be re-proposed on its own.

PR #1781, opened 2026-06-24
Optimize: use NOT instead of PUSH0;NUMEQUAL

Wi1l-B0t:
Use NOT instead of PUSH0;NUMEQUAL to check if int is zero or not.


Jim8y (approved):
Looks good to me.

shargon (approved):
Can be done also in optimizer?

Wi1l-B0t:
> Can be done also in optimizer?

I will try it.

PR #1784, opened 2026-06-26
Fix Nep11Token tokenId length limit to account for the account-token key

Jim8y:
## Problem

`Nep11Token` rejects tokenIds longer than **64** bytes:

```csharp
if (tokenId.Length > 64) throw new Exception("The argument \"tokenId\" should be 64 or less bytes long.");
```

But the per-account balance key is `Prefix(1) + owner(20) + tokenId` (see `UpdateBalance`: `ByteString key = owner + tokenId`). With `MaxStorageKeySize = 64`, a tokenId of **44–64 bytes** passes validation yet builds a storage key over the limit — so `Mint`/`Transfer` **FAULT at the storage write** instead of failing with the clear length message. `Mint` also had no length guard at all.

## Fix

- Tighten the guard to **43 bytes** (since 1 + 20 + 43 = 64) in `OwnerOf`, `Properties`, and `Transfer`, and correct the message.
- Add the missing guard to `Mint`.

The default 32-byte `Sha256` `NewTokenId` is unaffected. TokenIds in the 44–64 range never actually worked (they faulted), so this only converts a runtime FAULT into a clear, catchable exception — not a behavioral regression for any previously-working token.

## Tests

Updated the length-boundary assertions in the framework `Nep11TokenTest` and the compiler `UnitTest_NEP11` (valid 43, too-long 44, new message). Regenerated the `Contract_SupportedStandard11Enum` and `Contract_NEP11` artifacts. Both NEP-11 suites green.


shargon (approved):

Wi1l-B0t (commented):
Minor Fix.

Wi1l-B0t (approved):

ajara87 (approved):

PR #1785, opened 2026-06-26
Guard Update and Destroy in the default contract template

Jim8y:
## Problem

The **Basic** template — the default for `nccs new` — exposes `Update()` and `Destroy()` with **no authorization**:

```csharp
public static void Update(ByteString nefFile, string manifest, object? data = null)
{
    ContractManagement.Update(nefFile, manifest, data);   // anyone can call this
}

public static void Destroy()
{
    ContractManagement.Destroy();                          // anyone can call this
}
```

A beginner who deploys the scaffold verbatim ships a contract that **any account can update or destroy**. The other templates (NEP-17, Ownable, …) already gate these behind an owner check; the default one does not, which is the most dangerous place for the gap.

## Fix

Adopt the owner pattern the NEP-17 template already uses:
- store the deployer as the owner in `_deploy` (`data` or the tx sender);
- add `GetOwner()` / `IsOwner()`;
- guard `Update` and `Destroy` with `if (!IsOwner()) throw new InvalidOperationException("No authorization.");`.

The generated contract still compiles to a NEF — verified by generating the template and running `nccs` on it.

## Tests

`UnitTest_TemplateManager.TestGenerateBasicContract` now asserts the generated Basic contract contains `IsOwner` and the authorization guard.


src/Neo.Compiler.CSharp/TemplateManager.cs
```diff
@@ -256,19 +265,27 @@ public static string GetStoredMessage(string key)
...
+            UInt160 initialOwner = (UInt160)data;
+            ExecutionEngine.Assert(initialOwner.IsValid && !initialOwner.IsZero, ""owner must be valid"");
+            Storage.Put(new[] { Prefix_Owner }, initialOwner);
+
+            Storage.Put(""Deployed"", Runtime.Time);
         }
 
         public static void Update(ByteString nefFile, string manifest, object? data = null)
         {
+            if (!IsOwner())
```
ajara87: Add this check in a helper to be reusable?

src/Neo.Compiler.CSharp/TemplateManager.cs
```diff
@@ -236,6 +236,15 @@ namespace {{Namespace}}
...
         private const string HelloPrefix = ""Hello, "";
+        private const byte Prefix_Owner = 0xff;
+
+        [Safe]
+        public static UInt160 GetOwner()
+        {
+            return (UInt160)Storage.Get(new[] { Prefix_Owner });
+        }
+
+        private static bool IsOwner() => Runtime.CheckWitness(GetOwner());
```
Wi1l-B0t, line 249: Is it inline-able?

ajara87 (approved):

src/Neo.Compiler.CSharp/TemplateManager.cs
Jim8y: Addressed. The owner guard helper is now reusable, and both IsOwner and EnsureOwner are marked with AggressiveInlining.

Wi1l-B0t (approved):

shargon (commented):
Lgtm

PR #1786, opened 2026-06-26
Fix default(struct) to default-initialize its fields instead of an empty struct

Jim8y:
## Problem

`default(T)` for a custom struct emits `NEWSTRUCT0` — a struct with **zero** items:

```csharp
else if (type.IsValueType)
{
    AddInstruction(OpCode.NEWSTRUCT0);   // zero-field struct
}
```

But a struct's default should be a struct whose fields are themselves default-initialized (the same shape as `new T()`). So:

```csharp
struct S { public int Value; }
S s = default;
return s.Value;   // PICKITEM index 0 on a 0-length struct -> VM FAULT
```

faults at runtime. The same fallback also (incorrectly) emitted `NEWSTRUCT0` for an **enum**, whose default is its underlying zero.

## Fix

- Route a struct's `default` through the same `CreateObject` path `new T()` uses (emit each field's default, then `PACKSTRUCT`) — no constructor is run, so it stays `default` semantics.
- Emit `PUSH0` for an enum's `default`.

Primitives, strings, classes and `BigInteger` are unchanged.

## Tests

- `default(struct).Value` now returns `0` instead of faulting (new `TestStructDefaultFieldAccess`).
- The struct default is a one-field struct, not empty (`TestStructDefault` updated from `Count == 0` to `Count == 1`).
- `default(enum)` is `0` (new `TestEnumDefault`).

Only `Contract_Default` uses `default(struct)`/`default(enum)`, so it is the only regenerated artifact; the full `UnitTest_Default` suite (27) and the struct/enum/initializer suites (61) are green.

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Expression/LiteralExpression.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Expression/LiteralExpression.cs
@@ -110,10 +110,17 @@ internal partial class MethodConvert
                 {
                     AddInstruction(OpCode.PUSHNULL);
                 }
+                else if (type.TypeKind == TypeKind.Enum)
+                {
+                    // An enum's default value is its underlying zero, not a struct.
+                    AddInstruction(OpCode.PUSH0);
+                }
                 else if (type.IsValueType)
                 {
-                    // For structs and other value types, we need to create a default instance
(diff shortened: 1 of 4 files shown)


ajara87 (approved):
lgtm but would be good add more test

shargon (approved):

PR #1788, opened 2026-06-27
Fix NEP11 and Ownable nccs templates that did not compile

Jim8y:
## Problem

The contracts produced by `nccs new --template NEP11` and `nccs new --template Ownable` did not compile, so the very first build of a freshly generated project failed.

**NEP11** (`GetNep11ContractTemplate`):
- overrode a `GetKey(string)` method that `Nep11Token<T>` does not declare → **CS0115**;
- redeclared the base's `Prefix_TokenId` with a conflicting value (`0xfe` vs the base `0x02`) → **CS0108**, wiring a private storage scheme the base never reads;
- the `GetKey`/`ConcatKey`/`Prefix_TokenId` machinery was therefore both broken and dead — the base class manages token storage itself.

**Ownable** (`GetOwnableContractTemplate`):
- `SetData(string key, object value)` called `Storage.Put(key, value)`, but no `Storage.Put` overload accepts `object` → **CS1503**.

## Fix

- NEP11: remove the broken `GetKey` override, the `ConcatKey` helper that only it used, the shadowing `Prefix_TokenId` constant, and its unused `_deploy` write. Tokens continue to be stored by the `Nep11Token<T>` base.
- Ownable: type `SetData`'s value as `ByteString` to match `Storage.Put(ByteString, ByteString)`.

These are the hand-written templates emitted by `nccs new`; the maintained `dotnet new` templates under `src/Neo.SmartContract.Template` were already correct.

## Tests

Adds `AllTemplatesGenerateCompilableContracts`, which generates **every** template and compiles the generated source against the framework, asserting success. This reproduced both failures before the fix and now passes for all five templates — a non-compiling template can no longer ship undetected. All 27 template/new-command tests pass.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1789, opened 2026-06-27
Stop NC4007 from flagging supported string.PadLeft/PadRight

Jim8y:
## Problem

`StringMethodUsageAnalyzer` (NC4007) listed `PadLeft` and `PadRight` in its unsupported-method array and reported them as a **build-blocking error**. But the compiler fully supports all four overloads — `SystemCall.Register` maps `PadLeft(int)`, `PadLeft(int, char)`, `PadRight(int)`, `PadRight(int, char)` to `EmitStringPad`, and `UnitTest_StringPad` already verifies them against .NET semantics.

So the analyzer fails the build on legal code that the compiler compiles and runs correctly — a fail-closed false positive.

## Fix

Remove `PadLeft`/`PadRight` from `_unsupportedStringMethods`. `ToLowerInvariant`/`ToUpperInvariant` remain listed because the compiler supports only their `char` overloads, not the string-returning ones.

## Tests

- `PadLeftAndPadRight_ShouldNotReportDiagnostic` — all four overloads produce no diagnostic.
- `UnsupportedStringMethod_ShouldReportDiagnostic` — a still-unsupported method (`Normalize`) continues to report NC4007.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1790, opened 2026-06-27
Correct Helper.Reverse documentation: it reverses in place

Jim8y:
## Problem

`Helper.Reverse(this Array source)` is documented as *"Returns a reversed copy of parameter 'source'."* — but it emits `DUP` + `REVERSEITEMS`, and `REVERSEITEMS` reverses the backing buffer/array **in place**. `DUP` only duplicates the stack reference, so the returned value and the argument alias the same, now-reversed storage.

A developer who reuses `source` after calling `Reverse` (expecting the original order) gets silently mutated data. The in-repo callers (`UInt160.Parse`, `UInt256.Parse`) discard `source` immediately, so they're unaffected — which is exactly why the misleading contract has gone unnoticed.

## Fix

Correct the XML summary to state the in-place mutation/aliasing contract. The opcodes are unchanged — this is a documentation-only fix (changing it to a true copy would alter behavior current callers may rely on and cost an extra allocation, so that would belong in a separate, explicitly-named copying overload).


Wi1l-B0t (approved):

ajara87 (approved):

src/Neo.SmartContract.Framework/Helper.cs
shargon, line 230: Receive an array and return a ByteArray? It must be an array also, isn't it?

shargon: Ping @Jim8y

PR #1791, opened 2026-06-27
Optimize: less OpCodes for JMP with numeric compare

Wi1l-B0t:
Less OpCodes for JMP with numeric compare.


ajara87 (approved):

Jim8y (approved):
LGTM.

shargon (commented):
Lgtm but can be moved to optimized also

PR #1793, opened 2026-06-28
[merge 1/4] fix(testing): HTML-encode untrusted text in coverage report

Jim8y:
## Summary

The instruction coverage HTML report (`IntructionHtmlFormat`) interpolated three values straight into the HTML body without encoding:

- the contract name (`contract.Name`, derived from the manifest name),
- ABI method signatures (`method.Method`),
- instruction descriptions (`hit.Description`, which can include CALLT token method names).

All three originate from contract metadata that can be attacker-controlled. A crafted contract could therefore inject active markup into a coverage report that a developer or CI later opens or hosts (stored XSS). This is a single root cause — the formatter never encodes interpolated untrusted text — so all three are fixed together.

## Fix

Encode the three untrusted values with `System.Net.WebUtility.HtmlEncode` before writing them into the HTML body. Percentages (`:P2`), hit counts, and the internally generated markup (`kind`, `icon`, `branch`) are unchanged.

## Tests

- New `TestHtmlReport_EncodesUntrustedText`: injects a `<script>` payload into the rendered contract name and asserts the raw payload is absent and the encoded form (`&lt;script&gt;...`) is present, while method signatures still render (exercising the method/description paths).
- Full `Neo.SmartContract.Testing.UnitTests` suite passes (84).


Wi1l-B0t (approved):
Minor Fix.

ajara87 (approved):

shargon (commented):
Great

PR #1794, opened 2026-06-28
[merge 3/4] refactor(compiler): move CFG utilities to Neo.Compiler.ControlFlow

Jim8y:
## Summary

The basic-block / control-flow utilities under `Optimizer/Analysers/` — `BasicBlock`, `ContractInBasicBlocks`, `InstructionCoverage`, `JumpTarget`, `OpCodeTypes`, `TryCatchFinallyCoverage` — are general **control-flow-graph** infrastructure. They are consumed not only by the optimizer but also by the security analysers (`ReEntrancyAnalyzer`, `WriteInTryAnalyzer`, `MissingCheckWitnessAnalyzer`, etc.), so living under `Optimizer/` in namespace `Neo.Optimizer` mislocates them.

This moves them to a dedicated `ControlFlow/` folder under namespace **`Neo.Compiler.ControlFlow`** and updates all references (optimizer strategies, security analysers, and tests).

## Notes

- **Pure relocation, no behavioral change.** Full `Neo.Compiler.CSharp.UnitTests` suite passes (1323); full solution builds clean.
- Inside the moved files, `Instruction` is aliased to `Neo.VM.Instruction` *within the namespace body*. This is required because the new namespace nests under `Neo.Compiler`, whose internal IR `Instruction` would otherwise shadow the public VM type at a nearer scope.
- **Breaking change for external consumers**: the moved public types change namespace from `Neo.Optimizer` to `Neo.Compiler.ControlFlow`. Downstream users of these types (as a library) must update their `using` directives.

## Follow-up

A subsequent security PR (control-flow-sensitive missing-witness analysis) will branch from this one and build on the relocated CFG module.

PR #1795, opened 2026-06-28
[merge 4/4] feat(security): control-flow-sensitive missing-witness analysis

Jim8y:
> Based on #1794 (CFG relocation). Retarget to `master-n3` after that merges.

## Problem

`MissingCheckWitnessAnalyzer` suppressed a method as soon as **any** `Runtime.CheckWitness` appeared anywhere in it. A method could pass review while a storage write sat on a path the witness never guarded:

```csharp
public static void Exploit(UInt160 owner, byte[] k, byte[] v)
{
    if (Runtime.CheckWitness(owner)) { Runtime.Log("authorized"); } // witness gates only the if-body
    Storage.Put(Storage.CurrentContext, k, v);                       // reachable regardless -> unguarded
}
```

## Fix

Add `WitnessFlowAnalysis`, a control-flow-sensitive refinement over the relocated `Neo.Compiler.ControlFlow` CFG. For a method that contains a witness, it computes — as a forward **must-dataflow** — whether every path reaching each storage write passes a witness guard (a `CheckWitness` result consumed by `ASSERT` or a conditional jump). A write reachable while unguarded is reported.

### Soundness (no false positives)

A missing-witness warning is a trust signal, so the refinement only reports when control flow is fully modelled, and **bails out** (leaving the method to the existing presence-based result) whenever it meets something it cannot model precisely:

- inter-procedural (`CALL`) or dynamic (`CALLA`) calls,
- exception handlers (`TRY`/`ENDTRY`/`FINALLY`),
- a witness result it cannot trace to an assert / conditional branch.

Because uncertain methods fall back to the prior behaviour, **existing results cannot regress** — confirmed by the full suite.

## Tests

New cases in `MissingCheckWitnessTests`:
- `exploit` (branch-consumed witness, unguarded write after rejoin) → **flagged**
- `writeThenCheck` (write before the witness assertion) → **flagged**
- `guarded` (write inside the witness-guarded branch) → **not flagged**

Full `Neo.Compiler.CSharp.UnitTests` suite passes (1324), including the existing `SafeUpdate` / `SafeUpdateViaHelper` / `unsafeUpdate` cases.

## Scope

This is stage 1 (intra-procedural). Cross-procedural witness guards (a helper that asserts a witness) currently bail to the baseline; a follow-up can add helper guard summaries.


src/Neo.Compiler.CSharp/SecurityAnalyzer/MissingCheckWitnessAnalyzer.cs
```diff
@@ -93,13 +98,41 @@ public static MissingCheckWitnessVulnerability AnalyzeMissingCheckWitness(
...
+                    vulnerableMethods.Add(method.Name);
+                    continue;
+                }
+
+                // A witness exists somewhere. The baseline conservatively treats the method as safe.
+                // Refine with a control-flow-sensitive pass: only report when a storage write is
+                // provably reachable on a path that no witness check dominates. The refinement is
+                // sound (no false positives) and bails out (leaving the method unreported) whenever
+                // the control flow cannot be modelled precisely.
+                if (cfg is not null && WitnessFlowAnalysis.HasUnguardedWrite(cfg, method.Offset) == true)
```
Wi1l-B0t:
```suggestion
                if (cfg is not null && WitnessFlowAnalysis.HasUnguardedWrite(cfg, method.Offset))
```

src/Neo.Compiler.CSharp/SecurityAnalyzer/WitnessFlowAnalysis.cs
```diff
@@ -0,0 +1,251 @@
...
+                return bi;
+            }
+
+            // Otherwise the consumer must be the block's terminating conditional jump.
+            if (j != ins.Count - 1 || !IsConditionalJump(consumer))
+                return null;
+            if (b.jumpTargetBlocks.Count != 1)
+                return null;
+
+            bool jumpOnWitnessTrue = consumer switch
```
ajara87: This should bail out for conditional jumps other than JMPIF/JMPIFNOT. Currently JMPEQ/JMPNE/etc. pass IsConditionalJump but fall through to `_ => false` which can mark the wrong successor as the witnesstrue path and produce incorrect analysis results

src/Neo.Compiler.CSharp/SecurityAnalyzer/MissingCheckWitnessAnalyzer.cs
Jim8y: Updated with `is true` for the nullable result, and kept the lifecycle reporting path from the previous PR.

src/Neo.Compiler.CSharp/SecurityAnalyzer/WitnessFlowAnalysis.cs
Jim8y: Fixed. Unsupported conditional jumps now bail out instead of choosing a witness-true successor, and I added a JMPEQ regression test.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (approved):

PR #1813, opened 2026-07-10
Reject contracts exceeding the static slot limit

Jim8y:
## Summary

- Reject static slot allocation after the VM's 255-slot limit.
- Apply the same boundary to declared, captured, anonymous, and vtable slots.
- Cover both the valid 255-slot boundary and the 256-slot failure.

## Root cause

Static slot indexes and the `INITSSLOT` count are byte-sized, but new slots were allocated by casting the total count directly to `byte`. At 256 slots, the count wrapped and slot indexes could alias.

## Validation

- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter FullyQualifiedName~UnitTest_SlotLimits --no-restore`
- `dotnet build src/Neo.Compiler.CSharp/Neo.Compiler.CSharp.csproj -c Release --no-restore`

### DIFF
--- a/src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
+++ b/src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
@@ -742,7 +742,7 @@ namespace Neo.Compiler
         {
             if (!_staticFields.TryGetValue(symbol, out byte index))
             {
-                index = (byte)StaticFieldCount;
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1814, opened 2026-07-10
Enforce read-only effects for Safe methods

Jim8y:
## Summary

- Validate `CALLT` tokens reached from methods marked `[Safe]`.
- Reject dynamic or computed `CallFlags` unless they are provably read-only.
- Emit read-only tokens for safe extern methods, safe native methods, and compiler-generated read-only StdLib calls.

## Root cause

Safe-method enforcement inspected storage syscalls and directly analyzable `System.Contract.Call` arguments, but it did not inspect method tokens and accepted unresolved call flags. A method advertised as read-only could therefore retain write-capable external calls.

## Validation

- Safe-method and adjacent compiler tests: 37 passed
- `SampleFaunFeatures` contract compilation


ajara87:
UT Fails

Jim8y:
Fixed. Generated artifacts are updated and the full compiler test suite passes.

Wi1l-B0t (dismissed):

ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1815, opened 2026-07-10
Distinguish contracts with identical simple names

Jim8y:
## Summary

- Use Roslyn symbol identity when selecting and filtering contracts.
- Accept fully qualified target names and reject ambiguous simple names.
- Reject duplicate output base names before writing any artifacts.

## Root cause

Contract selection compared only the simple class name. Contracts in different namespaces could be compiled into the same context, and their default output names could overwrite each other in nondeterministic order.

## Validation

- Contract identity and output tests: 5 passed
- Existing prepared-contract selection tests


shargon (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1816, opened 2026-07-10
Honor evaluated MSBuild compile items

Jim8y:
## Summary

- Read evaluated MSBuild `Compile` items instead of recursively enumerating source files.
- Honor evaluated `DefineConstants` while preserving compiler-option symbols.
- Keep source-only temporary projects controlled by `CompilationOptions`.

## Root cause

The compiler partially interpreted project XML itself. Conditional item groups, imported properties, disabled default compile items, and include globs could therefore produce different source inputs from a normal MSBuild evaluation.

## Validation

- Project-evaluation and reference tests: 10 passed
- Debug compatibility test: 1 passed


shargon (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1819, opened 2026-07-10
Initialize late-discovered static slots

Jim8y:
## Summary

- Process static-field classes and vtables to a deterministic fixed point.
- Initialize each discovered class and vtable once, in slot order.
- Cover nested vtable methods that discover additional vtables and static fields.

## Root cause

Static fields and vtables were each enumerated from a single snapshot. Converting a vtable method could allocate later slots after their initialization pass had already finished, leaving those slots included in `INITSSLOT` but uninitialized.

## Validation

- `UnitTest_InitSlot`: 4 passed
- Polymorphism and static-variable tests: 9 passed


shargon (dismissed):

Wi1l-B0t (dismissed):

ajara87:
@Jim8y conflicts

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1820, opened 2026-07-10
Analyze witness checks on every write path

Jim8y:
## Summary

- Track storage and lifecycle operations through the existing basic-block graph.
- Report operations reachable on a path without a preceding witness check.
- Propagate helper summaries only after every normal return is positively proven to pass a witness check.

## Root cause

The analyzer aggregated witness and write occurrences across a method and its helpers. A witness in one branch or helper could suppress a warning for a write on another path.

## Validation

- Missing-witness tests: 22 passed
- Full security-analyzer slice: 79 passed


ajara87 (approved):

ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1821, opened 2026-07-10
Reject floating-point default expressions

Jim8y:
## Summary

- Detect floating-point default expressions before emitting folded constants.
- Report the existing floating-point diagnostic for `float`, `double`, and `decimal`.
- Preserve valid non-floating default expressions.

## Root cause

Casts such as `(int)default(float)` were constant-folded to integer zero before the unsupported floating-point type was validated, producing `PUSH0` instead of a diagnostic.

## Validation

- Floating-point default tests: 7 passed
- Existing default-expression tests: 17 passed


ajara87 (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
```diff
@@ -82,6 +85,27 @@ private bool TryConvertConstant(SemanticModel model, ExpressionSyntax syntax, Sy
...
+            .OfType<ExpressionSyntax>()
+            .FirstOrDefault(expression =>
+            {
+                if (expression is not DefaultExpressionSyntax &&
+                    !expression.IsKind(SyntaxKind.DefaultLiteralExpression))
+                    return false;
+
+                var typeInfo = model.GetTypeInfo(expression);
+                var type = typeInfo.Type ?? typeInfo.ConvertedType;
+                return type?.SpecialType is SpecialType.System_Single
```
Wi1l-B0t: `Half` will be rejected?

Jim8y: Added. `Half` is now rejected with the same floating-point diagnostic.

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1822, opened 2026-07-10
Initialize vtables before static fields

Jim8y:
## Summary

- Emit discovered vtable initialization before static-field initialization.
- Preserve the declared order and side effects of static-field initializers.
- Cover objects created both directly and through a helper method.

## Root cause

Static-field initializers ran before vtable slots were populated. An object created during initialization could capture a null vtable permanently and later fail during virtual dispatch.

## Validation

- Vtable initialization, slot, polymorphism, and static-variable tests: 13 passed
- Direct and helper-based object creation executed through the test VM


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1823, opened 2026-07-10
Preserve escaped function pointer targets during optimization

Jim8y:
## Summary

- Treat PUSHA targets as independent optimizer entry points.
- Preserve returned function pointers when the script has no CALLA instruction.
- Regenerate the two test artifacts affected by the retained targets.

## Why

The reachability pass only discovered PUSHA targets when CALLA was present. A function pointer returned from a public method could therefore point to private code removed by optimization and execute at an unrelated offset.

## Tests

- Added a VM regression test that invokes the returned pointer.
- Full Neo.Compiler.CSharp.UnitTests Release suite (1360 passed)


src/Neo.Compiler.CSharp/ControlFlow/EntryPoint.cs
```diff
@@ -62,18 +62,32 @@ public static Dictionary<int, EntryType> EntryPointsByMethod(ContractManifest ma
         /// <returns>A dictionary containing entry points.</returns>
         public static Dictionary<int, EntryType> EntryPointsByCallA(NefFile nef)
```
ajara87:
```suggestion
        [Obsolete]
        public static Dictionary<int, EntryType> EntryPointsByCallA(NefFile nef)
```
add obsolete?

Jim8y: Good point. I marked EntryPointsByCallA obsolete and kept the compatibility test without warning noise.

src/Neo.Compiler.CSharp/ControlFlow/EntryPoint.cs
```diff
@@ -60,20 +61,35 @@ public static Dictionary<int, EntryType> EntryPointsByMethod(ContractManifest ma
         /// </summary>
         /// <param name="nef">The NEF file.</param>
         /// <returns>A dictionary containing entry points.</returns>
+        [Obsolete("Use AllEntryPoints to include method and PUSHA entry points.")]
```
Wi1l-B0t: Just remove it, `nccs` is not a library.

ajara87: agree but then the method should be removed to no maintain the old logic

Jim8y: Removed the unused wrapper and its compatibility test.

Wi1l-B0t (dismissed):

ajara87 (dismissed):

Wi1l-B0t:
Conflicts.

shargon (approved):

Wi1l-B0t (approved):

PR #1824, opened 2026-07-10
Refresh the debug hash after optimization

Jim8y:
## Summary

- Refresh the debug information hash after rebuilding an optimized script.
- Add a regression test that verifies the fixture is rewritten and the final hashes match.

## Why

The optimizer updated the NEF script and checksum but left the top-level debug hash unchanged. Tools validating debug information against the final NEF could therefore reject a valid optimized build.

## Tests

- Debug info, optimizer automation, and DumpNef tests (30 passed)


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1825, opened 2026-07-10
Preserve debug method records after optimization

Jim8y:
## Summary

- Resolve deleted method entry instructions to the first surviving instruction within the original method range.
- Apply the same resolution to debug ranges and ABI offsets.
- Add a regression test covering Experimental optimization.

## Why

When a method entry instruction was removed without an explicit sequence-point mapping, the optimizer removed the entire debug method record. Contract_Types retained 34 manifest methods but only 12 ABI debug records with Experimental optimization.

## Tests

- Verified Contract_Types retains 34 manifest methods and 34 ABI debug records through the CLI.
- Full Neo.Compiler.CSharp.UnitTests Release suite (1360 passed)


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1826, opened 2026-07-10
Generate assembly from the final optimized script

Jim8y:
## Summary

- Generate .asm from the final optimized NEF instead of the pre-optimization compiler instructions.
- Reuse the rendered output for .nef.txt.
- Add a CLI regression test that compares an ADD-to-INC rewrite with the assembly output.

## Why

With Experimental optimization enabled, the NEF contained the rewritten instructions while .asm still reported the original offsets and opcodes. This made the assembly output unreliable when reviewing the deployed script.

## Tests

- Optimized assembly, CLI parameter, and DumpNef tests (13 passed)

### DIFF
--- a/src/Neo.Compiler.CSharp/Program.cs
+++ b/src/Neo.Compiler.CSharp/Program.cs
@@ -701,14 +701,18 @@ namespace Neo.Compiler
                 if (options.Assembly)
                 {
                     var asmPath = Path.Combine(outputFolder, $"{baseName}.asm");
-                    if (!TryFileOperation("write", asmPath, () => File.WriteAllText(asmPath, context.CreateAssembly())))
+                    var dumpNefContents = string.Empty;
+                    if (!TryFileOperation("write", asmPath, () =>
+                    {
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):

ajara87 (approved):

src/Neo.Compiler.CSharp/Program.cs
```diff
@@ -723,14 +723,18 @@ private static int ProcessOutput(Options options, string folder, CompilationCont
                 if (options.Assembly)
                 {
                     var asmPath = Path.Combine(outputFolder, $"{baseName}.asm");
-                    if (!TryFileOperation("write", asmPath, () => File.WriteAllText(asmPath, context.CreateAssembly())))
+                    var dumpNefContents = string.Empty;
+                    if (!TryFileOperation("write", asmPath, () =>
+                    {
+                        dumpNefContents = DumpNef.GenerateDumpNef(nef, debugInfo, manifest);
```
shargon, line 729: But it's writing a different thing, before was the assembly, now the nef, isn't it?

PR #1827, opened 2026-07-10
Make optimizer strategy discovery deterministic

Jim8y:
## Summary

- Discover optimizer strategies in a single batch.
- Apply deterministic ordering for strategies with equal priority.
- Preserve dynamic strategy registration and document the discovery rules.

## Why

The optimizer scanned the assembly twice during initialization. Duplicate registration was suppressed, but the second scan did no useful work, and equal-priority execution depended on reflection order.

## Tests

- Added a regression test for equal-priority ordering.
- Full `Neo.Compiler.CSharp.UnitTests` Release suite (1360 passed).

### DIFF
--- a/src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
+++ b/src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
@@ -16,7 +16,6 @@ using Neo.SmartContract.Manifest;
 using Neo.VM;
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
 
 namespace Neo.Optimizer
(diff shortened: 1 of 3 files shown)


src/Neo.Compiler.CSharp/Optimizer/Strategies/Optimizer.cs
```diff
@@ -45,61 +43,42 @@ static Optimizer()
...
-
-                // Validate method signature
-                if (method.ReturnType != typeof((NefFile, ContractManifest, JObject?)) ||
-                    method.GetParameters().Length != 3 ||
-                    method.GetParameters()[0].ParameterType != typeof(NefFile) ||
-                    method.GetParameters()[1].ParameterType != typeof(ContractManifest) ||
-                    method.GetParameters()[2].ParameterType != typeof(JObject))
+                foreach (MethodInfo method in type.GetMethods(BindingFlags.Public | BindingFlags.Static)
+                    .OrderBy(method => method.Name, StringComparer.Ordinal)
+                    .ThenBy(method => method.MetadataToken))
```
ajara87: Is sorting necessary? It adds CPU overhead and does not seem to be needed.

Jim8y: Good point. The final `orderedStrategies.Sort(CompareStrategies)` already provides deterministic ordering, so the discovery-time sorts are redundant. I removed them.

ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
LGTM

PR #1829, opened 2026-07-11
Fix: check method token count overflow

Wi1l-B0t:
Fix: check method token count(for CALLT) overflow.


src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -726,8 +726,11 @@ internal MethodConvert ConvertMethod(SemanticModel model, IMethodSymbol symbol)
 
         internal ushort AddMethodToken(UInt160 hash, string method, ushort parametersCount, bool hasReturnValue, CallFlags callFlags)
         {
-            int index = _methodTokens.FindIndex(p => p.Hash == hash && p.Method == method && p.ParametersCount == parametersCount && p.HasReturnValue == hasReturnValue && p.CallFlags == callFlags);
+            int index = _methodTokens.FindIndex(p => p.Hash == hash && p.Method == method &&
+                p.ParametersCount == parametersCount && p.HasReturnValue == hasReturnValue && p.CallFlags == callFlags);
             if (index >= 0) return (ushort)index;
+            if (_methodTokens.Count >= ushort.MaxValue)
```
ajara87:
```suggestion
            if (_methodTokens.Count > ushort.MaxValue)
```

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -726,8 +726,11 @@ internal MethodConvert ConvertMethod(SemanticModel model, IMethodSymbol symbol)
...
+                throw new CompilationException(DiagnosticId.SyntaxNotSupported, "Too many contract method calls.");
```
ajara87:
```suggestion
                throw new CompilationException(DiagnosticId.SyntaxNotSupported, $"Contracts with more than {ushort.MaxValue} method tokens (CALLT) are not supported.");
```

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
Wi1l-B0t: It should be `>=`.

ajara87: why?

Wi1l-B0t: Check before `_methodTokens.Add`.

Jim8y: NEF limits method tokens to 128, so checking `ushort.MaxValue` still allows the compiler to produce invalid NEF files. Please enforce the 128-token protocol limit and add boundary tests for 128 and 129 tokens.

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
```diff
@@ -726,8 +726,13 @@ internal MethodConvert ConvertMethod(SemanticModel model, IMethodSymbol symbol)
 
         internal ushort AddMethodToken(UInt160 hash, string method, ushort parametersCount, bool hasReturnValue, CallFlags callFlags)
         {
-            int index = _methodTokens.FindIndex(p => p.Hash == hash && p.Method == method && p.ParametersCount == parametersCount && p.HasReturnValue == hasReturnValue && p.CallFlags == callFlags);
+            int index = _methodTokens.FindIndex(p => p.Hash == hash && p.Method == method &&
+                p.ParametersCount == parametersCount && p.HasReturnValue == hasReturnValue && p.CallFlags == callFlags);
             if (index >= 0) return (ushort)index;
+
+            const int MaxMethodTokens = 128;
+            if (_methodTokens.Count > MaxMethodTokens)
```
Jim8y: The check is still off by one because it runs before `Add`. When `Count` is 128, this allows a 129th token. This should be `>= MaxMethodTokens`; please add the 128/129 boundary test.

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
Wi1l-B0t: Done.

src/Neo.Compiler.CSharp/CompilationEngine/CompilationContext.cs
Wi1l-B0t: Done.

Jim8y (changes requested):
The guard is correct now, but the requested boundary coverage is still missing. Please add tests that accept exactly 128 unique method tokens and reject the 129th; patch coverage is currently failing.

Wi1l-B0t:
> The guard is correct now, but the requested boundary coverage is still missing. Please add tests that accept exactly 128 unique method tokens and reject the 129th; patch coverage is currently failing.

Only one line. It just a guard.  
Hard to cover this line and doesn't matter.

Jim8y:
The off-by-one changed twice during this PR, so the 128/129 boundary is exactly what needs coverage. A one-line guard can still have protocol-level consequences. Please add the test.

Jim8y (approved):
LGTM.

ajara87 (approved):

shargon (commented):
Lgtm

PR #1830, opened 2026-07-11
Optimize: omit div overflow checking if divisor is not -1

Wi1l-B0t:
Omit div overflow checking if divisor is constant and not -1.

Dividing a value by a constant is a common expression.


ajara87 (dismissed):

Jim8y (commented):
The code checks rightExpr, which is the divisor, not the dividend. Please rename the PR title and description to use divisor.

Wi1l-B0t:
> The code checks rightExpr, which is the divisor, not the dividend. Please rename the PR title and description to use divisor.

Done.

Jim8y (approved):
LGTM.

shargon (approved):

PR #1832, opened 2026-07-11
[N3] - Omit check if operand is constant

ajara87:
Omit check if operand is constant. When the operand is a compile-time constant (e.g. a const field) the outcome is fully known at compile time and the guard is dead code.


Wi1l-B0t:
Expression `- constant`(like `-1`)  is literal integer value.
So this PR won't have any optimization effect, because optimization won't happen in this case.

Jim8y (changes requested):
ConvertExpression folds the full PrefixUnaryExpression through TryConvertConstant before ConvertPrefixUnaryExpression, so this branch is not reached for compile-time constants. The 8.33% patch coverage is consistent with that. Please add a test that demonstrates an emitted-code change; otherwise this optimization should be closed.

ajara87:
__

PR #1835, opened 2026-07-12
Fix: arguments not consumed if `Math.Clamp` throws exception

Wi1l-B0t:
Arguments not consumed if throw in `Math.Clamp`.
These values should be popped before `THROW`.


ajara87 (approved):

Jim8y (approved):
LGTM.

shargon (commented):
Lgtm, but how it was tested before?

PR #1836, opened 2026-07-12
Optimize: less opcodes for `Enum.Parse(EnumType, value)`

Wi1l-B0t:
Less opcodes for `Enum.Parse(EnumType, value)`.


ajara87 (approved):

Jim8y (changes requested):
Before merge, please fix the inline control flow and add a non-tail-call regression test. The handler emits a real `RET`, so `var value = Enum.Parse<E>("A"); return (int)value + 10;` returns `1` instead of `11`.

shargon:
> Before merge, please fix the inline control flow and add a non-tail-call regression test. The handler emits a real `RET`, so `var value = Enum.Parse<E>("A"); return (int)value + 10;` returns `1` instead of `11`.

Could you add this tests?

Wi1l-B0t:
> Before merge, please fix the inline control flow and add a non-tail-call regression test. The handler emits a real `RET`, so `var value = Enum.Parse<E>("A"); return (int)value + 10;` returns `1` instead of `11`.

What?

Wi1l-B0t:
> Before merge, please fix the inline control flow and add a non-tail-call regression test. The handler emits a real `RET`, so `var value = Enum.Parse<E>("A"); return (int)value + 10;` returns `1` instead of `11`.

This is another issue, and will be fixed later.

Jim8y:
Confirmed: the non-tail-call RET behavior already exists on master-n3 and is unchanged by this optimization. The branch is updated to the latest master-n3; all 14 Enum tests and the full CI pass. @shargon, please confirm this can be merged independently.

Jim8y (approved):
LGTM.

shargon (approved):

Issue #1841, opened 2026-07-13
Roadmap: complete Neo C# contract compatibility diagnostics


Jim8y:
## Summary

Neo smart contracts are written in C#, but they intentionally support only a subset of the C# language and .NET runtime surface. Some supported constructs also have NeoVM-specific behavior that differs from standard .NET semantics.

The long-term goal is to make these boundaries explicit, complete, and consistent so contract developers receive an actionable diagnostic before deployment whenever they:

- use unsupported syntax, types, APIs, members, or overloads;
- rely on behavior that cannot preserve standard C# semantics;
- violate ABI, manifest, framework, or Neo contract constraints;
- apply an unsafe automatic fix;
- write a construct that the compiler cannot lower safely.

This issue tracks the work required to turn the current analyzer collection into a complete Neo C# contract compatibility system.

## Scope

This roadmap covers:

- Roslyn analyzers and code fixes;
- compiler preflight validation;
- syntax, type, API, and overload support;
- documented semantic differences;
- template and package integration;
- IDE, `dotnet build`, and `nccs` diagnostic consistency;
- differential and property-based testing;
- high-confidence security and resource diagnostics;
- diagnostic documentation and release gates.

This roadmap does not claim that static analysis can prove arbitrary business logic correct. The intended guarantee is narrower and measurable: every known Neo C# language restriction or semantic difference must be classified and enforced consistently.

## Original audit baseline

At `master-n3` commit `7e6c38d8191157524050db61a6be967902a2a416`:

- The C# syntax matrix contains 137 probes: 96 supported and 41 unsupported.
- The analyzer release catalog contains 52 diagnostic IDs.
- The syntax probe harness runs only `UnsupportedSyntaxAnalyzer`.
- An unsupported probe currently passes when either the analyzer reports an error or the compiler rejects it later.
- `nccs` reads Roslyn compiler diagnostics but does not execute the complete Neo analyzer suite.
- Only the solution contract template references `Neo.SmartContract.Analyzer`; the NEP-11, NEP-17, Oracle, and Ownable templates do not.
- Diagnostic descriptors do not provide per-rule help links.
- Compiler support registrations, analyzer rules, syntax documentation, and tests are maintained independently.
- Some automatic fixes can change program meaning or leave code that still cannot be compiled.
- Known semantic differences are not all represented by compiler diagnostics.

Relevant locations:

[...]

PR #1844, opened 2026-07-13
Fix: use a valid lambda attribute syntax probe

Jim8y:
## Summary

- replace the invalid `[DoesNotReturn]` and `System.Action` combination in the supported lambda-attribute probe;
- use a lambda attribute that is valid under standard C# delegate rules;
- clarify that lambda attributes are metadata and do not change invocation through the underlying delegate.

## Root cause

The original example is rejected by the C# compiler with `CS8770` because `Action.Invoke()` does not carry the matching `DoesNotReturn` annotation. The Neo syntax matrix should demonstrate a valid supported construct rather than rely on compiler behavior that ordinary contract projects cannot build.

The replacement uses `[System.Obsolete]` on a `Func<int>` lambda, invokes the delegate, and remains valid in both standard C# and Neo contract compilation.

## Validation

- the original example was rejected by the standard C# compiler with `CS8770` even with analyzers disabled;
- the corrected `net10.0` contract built with warnings treated as errors;
- the corrected contract compiled with `nccs` and produced a NEF and manifest;
- all 139 syntax probe tests passed;
- `git diff --check` passed.

Part of #1841.


Wi1l-B0t (approved):
Doc update.

shargon (approved):

PR #1849, opened 2026-07-13
Feature: add the Neo C# contract profile schema

Jim8y:
## Summary

- add a Draft 2020-12 schema for the Neo C# contract profile
- require state-specific diagnostics, implementation references, and evidence
- enforce SemVer 2.0 syntax for profile versions
- add a complete fixture and valid and invalid schema test matrices

## Motivation

Compiler support, analyzer enforcement, tests, and documentation currently have no shared compatibility contract. This schema defines stable capability IDs, support states, matching surfaces, diagnostics, semantic differences, and the evidence required before a capability enters the profile.

This pull request defines and validates the format only. It does not claim that the current C# capability inventory is complete.

## Validation

- 30 focused schema tests
- valid and invalid SemVer matrices, including prerelease and build metadata edge cases
- schema and fixture JSON parsing
- 230 analyzer tests with the seven audited fixes combined
- combined Release solution build
- combined full solution test run: 2,117 passed
- focused format and diff checks

Part of #1841.

### DIFF
--- a/profiles/neo-csharp-profile.schema.json
+++ b/profiles/neo-csharp-profile.schema.json
@@ -1,0 +1,511 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/neo-project/neo-devpack-dotnet/master-n3/profiles/neo-csharp-profile.schema.json",
+  "title": "Neo C# Contract Compatibility Profile",
+  "description": "Machine-readable compatibility contract shared by the Neo C# compiler, analyzers, documentation, and tests.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "$schema",
+    "schemaVersion",
+    "profileVersion",
+    "devpackVersion",
+    "capabilities"
+  ],
+  "properties": {
+    "$schema": {
+      "type": "string",
+      "minLength": 1
+    },
+    "schemaVersion": {
+      "const": 1
+    },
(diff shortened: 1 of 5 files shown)


profiles/neo-csharp-profile.schema.json
```diff
@@ -0,0 +1,511 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
shargon, line 2: How it's generated?

Jim8y: It is not generated. This PR defines the versioned schema as the source of truth, and the tests validate the schema and its fixture. The complete capability inventory will be added separately and validated against it.

shargon: And how it will be maintained, manually?

Jim8y: The schema itself is a versioned format, so changes are reviewed manually. The capability inventory will not rely on manual synchronization alone: follow-up work will validate compiler handlers, analyzer diagnostics, tests, and generated documentation against it in CI.

shargon (dismissed):

profiles/neo-csharp-profile.schema.json
shargon: It should be good have a readme about how modify it.

shargon (approved):

ajara87 (approved):

PR #1861, opened 2026-07-13
Fix: distinguish dynamic types from named identifiers

Jim8y:
## Summary

- use Roslyn type information before reporting `NC4037`
- keep rejecting values whose actual type is `dynamic`
- allow ordinary locals and escaped members that are only named `dynamic`

## Why

The analyzer compared only identifier text. A statically typed local named `dynamic` and a member named `@dynamic` therefore received the unsupported dynamic-binding error even though neither construct used dynamic dispatch.

Part of #1841.

## Validation

- controlled regression test reproduced two false `NC4037` diagnostics before the fix
- dynamic behavior tests: 3/3 passed
- analyzer tests: 180/180 passed
- real identifier-only contract compiled to NEF and manifest with `nccs`
- Release solution build passed
- focused `dotnet format --verify-no-changes`: passed


Wi1l-B0t (approved):
Minor Fix.

shargon (approved):

PR #1864, opened 2026-07-13
Fix: resolve NC4023 field types semantically

Jim8y:
## Summary

- resolve static field types through the semantic model before applying `NC4023`
- recognize fully qualified and aliased Framework `UInt160`, `UInt256`, and `ECPoint` types
- compare exact referenced Framework symbols instead of source text or short names
- avoid rejecting nested, source-defined, and unrelated types with matching names

## Rationale

The analyzer previously compared field type source text with three short names. Qualified and aliased Framework types bypassed the rule, while unrelated same-named types could be rejected incorrectly.

## Validation

- 12 focused static-field tests
- qualified and aliased Framework positive cases
- nested and source-defined lookalike negative cases
- 230 analyzer tests with the seven audited fixes combined
- combined Release solution build
- combined full solution test run: 2,117 passed
- focused format and diff checks

Part of #1841.

### DIFF
--- a/src/Neo.SmartContract.Analyzer/StaticFieldInitializationAnalyzer.cs
+++ b/src/Neo.SmartContract.Analyzer/StaticFieldInitializationAnalyzer.cs
@@ -24,6 +24,10 @@ namespace Neo.SmartContract.Analyzer
     {
         public const string DiagnosticId = "NC4023";
         private const string Category = "Usage";
+        private const string FrameworkAssemblyName = "Neo.SmartContract.Framework";
+        private const string UInt160MetadataName = FrameworkAssemblyName + ".UInt160";
+        private const string UInt256MetadataName = FrameworkAssemblyName + ".UInt256";
+        private const string ECPointMetadataName = FrameworkAssemblyName + ".ECPoint";
 
         private static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
             DiagnosticId,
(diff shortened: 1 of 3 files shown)


ajara87 (dismissed):

src/Neo.SmartContract.Analyzer/StaticFieldInitializationAnalyzer.cs
```diff
@@ -39,19 +43,50 @@ public override void Initialize(AnalysisContext context)
...
+                compilationContext.RegisterSyntaxNodeAction(
+                    nodeContext => AnalyzeNode(nodeContext, uint160Type, uint256Type, ecPointType),
+                    SyntaxKind.FieldDeclaration);
+            });
         }
 
-        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        private void AnalyzeNode(
+            SyntaxNodeAnalysisContext context,
+            INamedTypeSymbol? uint160Type,
```
shargon: It's better to have a class of know types, isn't it? This can be reused

Jim8y: Good point. I extracted the Framework symbols into a reusable `KnownFrameworkTypes` class and updated this analyzer to use it.

shargon (approved):

ajara87 (approved):

PR #1866, opened 2026-07-13
Fix: validate supported Math overloads

Jim8y:
## Summary

- validate `System.Math` calls against the exact integer overloads registered by the compiler
- reject unsupported overloads even when another overload with the same method name is supported
- include the resolved signature in `NC4005` diagnostics

## Reason

The previous method-name blacklist allowed unregistered overloads such as `Math.DivRem(int, int, out int)` to pass analysis and fail later in compiler lowering.

## Validation

- confirmed the new overload regression test expected one diagnostic and received none before the change
- confirmed `nccs` independently rejected the same overload with `NC1002`
- passed 9 focused Math analyzer tests
- passed all 180 analyzer tests
- passed the Release solution build with no errors
- passed focused formatting verification
- compiled a `net10.0` consumer covering all 41 registered Math overloads with no warnings or errors
- compiled the same 41-overload contract with `nccs --optimize All` and generated its NEF and manifest

Related to #1841.

### DIFF
--- a/src/Neo.SmartContract.Analyzer/SystemMathUsageAnalyzer.cs
+++ b/src/Neo.SmartContract.Analyzer/SystemMathUsageAnalyzer.cs
@@ -22,21 +22,35 @@ namespace Neo.SmartContract.Analyzer
     {
         public const string DiagnosticId = "NC4005";
 
-        private static readonly ImmutableHashSet<string> UnsupportedMathMethods = ImmutableHashSet.Create(
-            "Acos", "Asin", "Atan", "Atan2", "Ceiling", "Cos", "Cosh",
-            "Exp", "Floor", "IEEERemainder", "Log", "Log10", "Pow", "Round",
-            "Sin", "Sinh", "Sqrt", "Tan", "Tanh",
-            "Truncate");
+        private static readonly ImmutableHashSet<SpecialType> SignedIntegralTypes = ImmutableHashSet.Create(
+            SpecialType.System_SByte,
+            SpecialType.System_Int16,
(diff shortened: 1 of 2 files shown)


Wi1l-B0t (approved):

src/Neo.SmartContract.Analyzer/SystemMathUsageAnalyzer.cs
```diff
@@ -57,14 +71,48 @@ private void AnalyzeSyntax(SyntaxNodeAnalysisContext context)
...
+                methodSymbol.ToDisplayString(DiagnosticDisplayFormat));
 
             context.ReportDiagnostic(diagnostic);
         }
+
+        private static bool IsSupportedMathMethod(IMethodSymbol method)
+        {
+            return method.Name switch
+            {
+                "Abs" or "Sign" => HasUniformParameterType(method, 1, SignedIntegralTypes),
```
Wi1l-B0t, line 88: `CopySign` is not supported now?

shargon (approved):

src/Neo.SmartContract.Analyzer/SystemMathUsageAnalyzer.cs
Jim8y: `Math.CopySign(double, double)` is not supported because it operates on floating-point values. Integral methods such as `int.CopySign` and `BigInteger.CopySign` remain supported through their separate compiler handlers.

ajara87 (approved):

PR #1871, opened 2026-07-13
Fix: validate supported BigInteger overloads

Jim8y:
## Summary

- replace the single-method BigInteger blacklist with exact supported signature validation
- report unsupported overloads and formatting, conversion, and comparison APIs with NC4006
- avoid duplicate diagnostics for ordinary calls while retaining method-group validation

## Validation

- controlled duplicate and missing-diagnostic cases reproduced before the fix
- 2 focused BigInteger analyzer tests passed
- 180 analyzer tests passed
- 2,058 solution tests passed
- a real net10.0 analyzer consumer reported four expected NC4006 diagnostics
- an independent nccs compilation rejected an unsupported call with NC1002
- a contract covering the registered BigInteger method surface compiled to NEF and manifest artifacts
- Release solution build completed with 0 errors


src/Neo.SmartContract.Analyzer/BigIntegerUsageAnalyzer.cs
```diff
@@ -61,24 +56,153 @@ private void AnalyzeSyntax(SyntaxNodeAnalysisContext context)
...
+                "ModPow" or "Clamp" => HasParameters(method, bigIntegerType, bigIntegerType, bigIntegerType),
+                "Add" or "Subtract" or "Multiply" or "Divide" or "Remainder" or "Compare" or
+                    "GreatestCommonDivisor" or "DivRem" or "CopySign" or "Max" or "Min" =>
+                    HasParameters(method, bigIntegerType, bigIntegerType),
+                "Negate" or "IsEvenInteger" or "IsOddInteger" or "IsNegative" or "IsPositive" or
+                    "IsPow2" or "LeadingZeroCount" or "Log2" or "PopCount" or "Abs" =>
+                    HasParameters(method, bigIntegerType),
+                "Parse" => HasParameters(method, SpecialType.System_String),
+                "TryParse" => method.Parameters.Length == 2 &&
+                              IsSpecialParameter(method, 0, SpecialType.System_String) &&
```
shargon, line 127: I prefer use nameof(BigInteger.XXXX)

shargon (changes requested):

ajara87 (approved):

src/Neo.SmartContract.Analyzer/BigIntegerUsageAnalyzer.cs
Jim8y: Updated the available BigInteger members to use `nameof(...)`. Generic math members that are unavailable in netstandard2.0 remain string literals. @shargon could you please recheck?

shargon (approved):

PR #1873, opened 2026-07-14
Fix: resolve Notify event names by parameter

Jim8y:
## Problem

`NotifyEventNameAnalyzer` reads the first syntactic argument as the event name. A valid call can reorder named arguments, causing the analyzer to inspect `state` and miss an invalid `eventName`.

## Fix

- Resolve the invocation through `IInvocationOperation`.
- Select the event-name argument by parameter ordinal instead of source position.
- Keep the existing exact `Runtime.Notify` target check and literal-only analysis.

## Validation

- Focused `NotifyEventNameAnalyzer` tests: 10 passed.
- Complete analyzer tests: 181 passed.
- Complete solution tests: 2,059 passed.
- A standalone consumer build reports NC4019 for a misspelled reordered named argument.
- The same consumer build succeeds for a matching reordered named argument.
- Focused formatting verification passed.


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
Lgtm

PR #1879, opened 2026-07-14
Fix: report volatile declarations once

Jim8y:
## Problem

A field declaration with multiple variables produces one NC4014 warning per variable even though it has only one `volatile` modifier. The diagnostics are also attached to the variable names instead of the unsupported keyword.

## Fix

- Report one diagnostic for the field declaration.
- Attach the diagnostic to the `volatile` keyword.
- Preserve the existing code fix, including multi-variable declarations.

## Validation

- A controlled test reproduced two NC4014 diagnostics for one modifier.
- A standalone consumer reproduced two warnings before the change and one warning on `volatile` afterward.
- Focused Volatile analyzer tests: 6 passed.
- Complete analyzer tests: 181 passed.
- Complete solution run: 2,058 of 2,059 passed; the sole live RPC test received an empty public-node response and passed on an isolated rerun.
- Focused formatting verification passed.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1883, opened 2026-07-14
Fix: use valid compound assignment probe

Jim8y:
## Summary

- replace the invalid static compound-assignment declaration with the C# 14 instance `void` form;
- exercise the operator from contract code so the probe reaches the intended language boundary;
- describe the current rejection at the embedded Roslyn front end.

## Root cause

The probe used `static Counter operator +=(Counter, int)`. C# 14 compound-assignment operators are instance operators that return `void`, so the old sample was rejected as malformed C# and did not test the feature it claimed to cover.

The corrected sample builds under the .NET 10 SDK. The Roslyn version used internally by `nccs` does not yet recognize the declaration and rejects it before Neo lowering, so the feature remains classified as unsupported.

## Validation

- standard `net10.0` consumer build with warnings treated as errors: passed;
- direct `nccs` compilation: rejected the valid declaration at the compiler front end;
- complete syntax-probe suite: 139 passed;
- Release solution build: passed with only two pre-existing nullable warnings;
- Markdown lint baseline comparison: 41 existing findings before and after, no new findings;
- `git diff --check`: passed.

Part of #1841.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Looks good

PR #1884, opened 2026-07-14
Fix: encode string constants as UTF-8

Jim8y:
## Summary

- encode ordinary C# string constants with strict UTF-8
- preserve raw one-byte values for constants converted to `ByteString`
- recognize `ByteString` conversions semantically across return, local, argument, and explicit-cast contexts

## Why

The shared string push path treated every character up to `0xff` as a raw byte. A normal string literal such as `"\u00e9"` therefore emitted invalid UTF-8 and could fail decoding or produce incorrect string search and equality results in NeoVM. Raw `ByteString` constants intentionally rely on the existing byte-preserving behavior, so that path remains compatible.

## Validation

- Unicode string constant tests: 3 passed against a deployed NeoVM contract
- related string and byte-string tests: 76 passed
- full compiler test suite: 1416 passed
- Release solution build: 0 errors
- focused `dotnet format --verify-no-changes`: passed

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
@@ -61,6 +61,8 @@ internal partial class MethodConvert
             if (value == null)
                 return false;
 
+            bool isByteStringConstant = value is string && IsByteStringConstant(model, syntax);
+
             typeSymbol = GetTypeSymbol(syntaxNode, model);
 
             if (typeSymbol != null)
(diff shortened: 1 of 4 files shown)


ajara87 (approved):

Wi1l-B0t (approved):

src/Neo.Compiler.CSharp/MethodConvert/Expression/Expression.cs
```diff
@@ -85,6 +90,19 @@ private bool TryConvertConstant(SemanticModel model, ExpressionSyntax syntax, Sy
         }
     }
 
+    private static bool IsByteStringConstant(SemanticModel model, ExpressionSyntax syntax)
+    {
+        var byteStringType = model.Compilation.GetTypeByMetadataName("Neo.SmartContract.Framework.ByteString");
+        if (byteStringType is null)
+            return false;
```
shargon, line 97: And buffer?

PR #1885, opened 2026-07-14
Fix: preserve NUL in string char searches

Jim8y:
## Summary

- preserve `\0` as a one-byte search value for `string.Contains(char)` and `string.IndexOf(char)`
- avoid treating the integer representation of the NUL character as an empty byte string
- cover empty, absent, leading, and embedded NUL searches in deployed contract tests

## Validation

- 1,413 compiler tests
- 52 string tests, including focused NUL search execution
- Release solution build
- focused format check

Progresses #1841


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
Looks good

PR #1886, opened 2026-07-14
Fix: narrow the multiple-catch diagnostic location

Jim8y:
## Summary

- report NC4024 at the second `catch` keyword instead of highlighting the complete `try` statement
- keep the existing single diagnostic and total catch-count message
- cover one, two, and three catch blocks plus an incomplete second catch during editing

## Validation

- 4 focused analyzer tests
- 182 analyzer tests
- Release solution build
- focused format check

Progresses #1841


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1887, opened 2026-07-14
Fix: preserve range endpoint evaluation order

Jim8y:
## Summary

- evaluate range receivers, start endpoints, and end endpoints exactly once in C# order
- defer direct and nested from-end offsets until both endpoint expressions have run, including on null receivers
- preserve conditional, switch, checked, cast, null-forgiving, mixed, negative, and out-of-range endpoint semantics
- report unsupported System.Index endpoint forms before they can produce invalid VM code

## Validation

- controlled endpoint-order and unsupported Index regressions
- 11 focused deployed and compile-time index/range tests
- compile-and-deploy checks under None, Basic, and All optimization levels
- 1,423 compiler tests
- Release solution build
- focused format check

Progresses #1841


shargon (commented):
Looks well tested

shargon (approved):

ajara87 (approved):

Wi1l-B0t (approved):

PR #1891, opened 2026-07-14
Fix: reject unsupported System.Index construction

Jim8y:
## Summary

- reject Framework `System.Index` construction before object creation can silently emit no stack value
- cover explicit, target-typed, parameterless, named, dynamic, negative, and inline construction forms
- report `NC2001` over the complete object-creation expression with actionable index guidance
- preserve user-defined `Index` types, source-defined `System.Index` shadows, ordinary objects, and inline caret/range access

## Validation

- controlled `nccs` compile-and-run reproduction of empty-stack and invalid `PICKITEM` failures
- 22 focused constructor diagnostic and deployed runtime tests
- existing `IndexOrRange` regression test
- 1,435 compiler tests
- Release solution build
- focused format check

Progresses #1841


Wi1l-B0t (approved):
Minor Fix.

shargon (approved):

PR #1901, opened 2026-07-18
Fix: validate supported BitOperations overloads

Jim8y:
## Summary
- add NC4060 for unsupported `BitOperations` methods and overloads
- allow only the exact `uint` and `ulong` signatures registered by the compiler
- cover aliases, static imports, and user-defined same-named types
- document the supported and unsupported surface

## Root cause
The compiler supports a limited `BitOperations` surface, but the analyzer did not distinguish registered overloads from methods that have no lowering.

## Validation
- 4 focused analyzer tests
- 282 analyzer tests
- 139 syntax checklist tests
- Release solution build with 0 errors
- focused formatting verification

Progresses #1841.


src/Neo.SmartContract.Analyzer/BitOperationsUsageAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class BitOperationsUsageAnalyzer : DiagnosticAnalyzer
+    {
+        public const string DiagnosticId = "NC4060";
+
+        private const string BitOperationsTypeName = "global::System.Numerics.BitOperations";
+
+        private static readonly SymbolDisplayFormat FullyQualifiedFormat =
+            SymbolDisplayFormat.FullyQualifiedFormat;
```
ajara87, line 28: FullyQualifiedFormat is assigned SymbolDisplayFormat.FullyQualifiedFormat directly this field just aliases a static property. You can use the original directly or remove the field.

ajara87 (approved):

shargon (approved):

PR #1902, opened 2026-07-18
Fix: preserve fixed-width shift semantics

Jim8y:
## Summary
- mask fixed-width shift counts according to C# rules
- preserve fixed-width truncation in checked and unchecked shifts
- keep checked conversion behavior for small compound-assignment targets
- refresh affected compiler artifacts and gas baselines

## Root cause
NeoVM shifts consume the full integer count and produce arbitrary-precision integer results. The compiler passed fixed-width C# operands through without applying the five- or six-bit count mask, and the checked path treated shift overflow as an exception even though C# shifts truncate their fixed-width result.

## Validation
- 26 focused shift and compound-assignment tests
- 1,479 compiler tests excluding the process-global temp-directory suite
- 4 serialized temp-directory security tests
- 321 framework tests
- Release solution build with 0 errors
- focused formatting verification

Progresses #1841.


src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -113,8 +119,67 @@ private void ConvertBinaryExpression(SemanticModel model, BinaryExpressionSyntax
...
+    /// </summary>
+    /// <remarks>
+    /// C# masks the shift count to five bits for operands up to 32 bits and to
+    /// six bits for 64-bit operands. BigInteger shift counts are not masked.
+    /// </remarks>
+    private bool MaskFixedWidthShiftCount(ITypeSymbol? leftType)
+    {
+        if (leftType is null) return false;
+
+        int? mask = leftType.SpecialType switch
```
ajara87, line 150:
```suggestion
        if (leftType is null) return false;
        
        // Unwrap Nullable<T> so that int?, long?, etc. are treated identically
        // to their non-nullable counterparts, consistent with EnsureIntegerInRange
        // and CheckLeftShiftOverflow.
        if (leftType is INamedTypeSymbol
            {
                OriginalDefinition.SpecialType: SpecialType.System_Nullable_T
            } nullableType)
        {
            leftType = nullableType.TypeArguments[0];
        }
        
        int? mask = leftType.SpecialType switch
```
`MaskFixedWidthShiftCount` checks `leftType.SpecialType` directly but nullable fixed width operands have `Nullable<T>` as their outer symbol. In that case this returns false and `CheckLeftShiftOverflow` unwraps the type and retains the old range checking behavior. Could we unwrap `Nullable<T>` before selecting the mask

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -113,8 +119,67 @@ private void ConvertBinaryExpression(SemanticModel model, BinaryExpressionSyntax
...
+                EnsureIntegerInRange(type);
+        }
+    }
+
+    /// <summary>
+    /// Applies the C# shift-count mask for fixed-width integral operands.
+    /// </summary>
+    /// <remarks>
+    /// C# masks the shift count to five bits for operands up to 32 bits and to
+    /// six bits for 64-bit operands. BigInteger shift counts are not masked.
```
ajara87:
```suggestion
    /// six bits for 64-bit operands. BigInteger shift counts are not masked.
    /// Nullable wrappers (e.g. int?) are unwrapped before the mask is selected.
```

shargon (approved):

ajara87 (approved):

Wi1l-B0t (approved):

PR #1904, opened 2026-07-18
Test: run the complete analyzer suite in syntax probes

Jim8y:
## Summary

- run all 32 built-in contract analyzers for every syntax probe
- keep analyzer errors visible alongside compiler diagnostics
- cover analyzer-only restrictions in the syntax compatibility matrix

## Why

The probe harness only ran `UnsupportedSyntaxAnalyzer`. Unsupported APIs and type restrictions handled by other analyzers could therefore be missed by the compatibility suite.

## Validation

- `dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter FullyQualifiedName~SyntaxTests --no-restore` (139 passed)
- targeted `dotnet format --verify-no-changes`

Part of #1841.


ajara87 (dismissed):

tests/Neo.Compiler.CSharp.UnitTests/SyntaxProbes/Helper.cs
```diff
@@ -49,7 +49,40 @@ internal static class Helper
     }.GetParseOptions());
     private static readonly Lazy<ImmutableArray<MetadataReference>> AnalyzerReferences = new(CreateAnalyzerReferences);
     private static readonly Lazy<ImmutableArray<DiagnosticAnalyzer>> SyntaxAnalyzers = new(() =>
-        ImmutableArray.Create<DiagnosticAnalyzer>(new UnsupportedSyntaxAnalyzer()));
+    [
+        new BanCastMethodAnalyzer(),
+        new BigIntegerCreationAnalyzer(),
```
shargon: Is not better to get them by reflection?

tests/Neo.Compiler.CSharp.UnitTests/SyntaxProbes/Helper.cs
```diff
@@ -48,15 +49,33 @@ internal static class Helper
...
     {
         var source = BuildMethodBodySource(codeBlock);
         AssertCompilationResult(source, expectSuccess: true, "Expected snippet to compile successfully.");
     }
 
+    private static ImmutableArray<DiagnosticAnalyzer> CreateSyntaxAnalyzers() =>
+        GetLoadableAnalyzerTypes()
+            .Where(type => !type.IsAbstract && typeof(DiagnosticAnalyzer).IsAssignableFrom(type))
+            .OrderBy(type => type.FullName, StringComparer.Ordinal)
+            .Select(type => (DiagnosticAnalyzer)Activator.CreateInstance(type)!)
```
ajara87:
```suggestion
            .Select(type =>
                        {
                            var instance = Activator.CreateInstance(type)
                                ?? throw new InvalidOperationException($"Could not instantiate analyzer {type.FullName}. Ensure it has a public parameterless constructor.");
                            return (DiagnosticAnalyzer)instance;
                        })
```
`Activator.CreateInstance(type)!` suppresses nullability warnings but it can still return null at runtime (e.g., no accessible parameterless constructor). Add an explicit null check and throw a clear error message

shargon (dismissed):

Wi1l-B0t (dismissed):

ajara87 (dismissed):

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Looks good now

PR #1917, opened 2026-07-20
Preserve lifted nullable shift semantics

Jim8y:
## Summary

- return null when either operand of a lifted shift is null
- preserve left-to-right, exactly-once operand evaluation
- retain fixed-width shift masking and normalization for non-null values

## Validation

- 1,497 compiler tests passed
- focused tests passed under None, Basic, and All optimization modes
- covered left and right shifts, nullable counts, null propagation, and side effects
- format verification passed

Related to #1841.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #1927, opened 2026-07-23
fix Analyzer settings for v3.10.1 release

superboyiii:
The Analyzer nupkg uploaded successfully. The job failed when NuGet auto-pushed the sibling snupkg: Analyzer sets IncludeBuildOutput=false (DLL under analyzers/) but still generated an empty symbol package with no .pdb files, so nuget.org returned 400. That aborted the script before the remaining packages were published. Details in https://github.com/neo-project/neo-devpack-dotnet/actions/runs/29987845379/job/89144878777


shargon (dismissed):

scripts/release-ordered.sh
```diff
@@ -118,9 +118,15 @@ pack_and_push() {
         --no-restore
     
     if [[ "$DRY_RUN" == "false" ]]; then
-        # Push the package
+        # Push the main nupkg; nuget also uploads a sibling .snupkg when present.
         echo "Publishing $project_name to $NUGET_SOURCE..."
-        dotnet nuget push "$OUTPUT_DIR/$project_name.*.nupkg" \
+        local package_file
+        package_file=$(ls -1 "$OUTPUT_DIR/$project_name."*.nupkg 2>/dev/null | head -1)
```
ajara87: Shouldn’t we clean the output directory first? If there are packages from older versions head -1 could pick a version different from $VERSION

superboyiii: > Shouldn’t we clean the output directory first? If there are packages from older versions head -1 could pick a version different from $VERSION

Good catch! You're right.

Jim8y (approved):
LGTM.

superboyiii:
@ajara87 Done

ajara87 (approved):
lgtm

PR #1929, opened 2026-07-23
Fix: the result consumed by JMPIF

Wi1l-B0t:
This PR is a fix for `char.IsLetterOrDigit`:
* Drop the input char in the end;
* Dup the result, because JMPIF will pop the top of the stack;


Jim8y (changes requested):
The fix is correct, but the current test also passes with the old implementation because the generated bool proxy converts the Integer result to `true`. Please add a VM-level regression test that invokes `testCharIsLetterOrDigit` and asserts that the returned `StackItem.Type` is `Boolean` for the digit and letter branches.

Wi1l-B0t:
> The fix is correct, but the current test also passes with the old implementation because the generated bool proxy converts the Integer result to `true`. Please add a VM-level regression test that invokes `testCharIsLetterOrDigit` and asserts that the returned `StackItem.Type` is `Boolean` for the digit and letter branches.

Done.

shargon (approved):

ajara87 (approved):

Jim8y (approved):
LGTM.

PR #1934, opened 2026-07-26
Fix: encode runtime char values as UTF-8 strings

Jim8y:
## Summary

- encode valid runtime `char` values in Neo's UTF-8 string representation
- use the conversion for `new string(char, count)`, `char.ToString()`, and `StringBuilder.Append(char)`
- make constant and runtime surrogate handling consistent by replacing isolated UTF-16 surrogate code units with U+FFFD
- validate `string(char, count)` limits using the encoded UTF-8 byte length

## Scope

A C# `char` is one UTF-16 code unit. Valid non-surrogate values encode to one, two, or three UTF-8 bytes. Four-byte Unicode scalar values require a surrogate pair in a `string`; regular string values already preserve them. Neo's strict UTF-8 string model cannot retain an isolated surrogate, so converting surrogate `char` values separately produces replacement characters.

## Validation

- boundary, surrogate, size-limit, and multilingual execution tests passed under None, Basic, and All optimization modes
- 1,523 compiler tests passed
- focused format verification passed

Part of #1841.


Wi1l-B0t:
It's necessary to support UTF-8 ?

ajara87 (dismissed):

shargon:
conflicts

ajara87 (dismissed):

Wi1l-B0t:
> ## Summary
> * encode runtime `char` values as valid one-, two-, or three-byte UTF-8
> * use the encoding for `new string(char, count)`, `char.ToString()`, and `StringBuilder.Append(char)`
> * preserve constant-character folding and replace isolated surrogates consistently
> 
> ## Validation
> * boundary and multilingual execution tests passed under None, Basic, and All optimization modes
> * 1,517 compiler tests passed
> * focused format verification passed
> 
> Part of #1841.

Only one-, two-, or three-byte UTF-8 ?
An UTF-8 rune may be 4-byte.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
```diff
@@ -784,9 +784,92 @@ private static void HandleBoolToString(MethodConvert methodConvert, SemanticMode
...
+        Within(0xd800, 0xdfff);
+        JumpIfTrue(replacementTarget);
+
+        NewBuffer(3);
+        SetUtf8BufferByte(charSlot, 0, shift: 12, mask: 0x0f, prefix: 0xe0);
+        SetUtf8BufferByte(charSlot, 1, shift: 6, mask: 0x3f, prefix: 0x80);
+        SetUtf8BufferByte(charSlot, 2, shift: 0, mask: 0x3f, prefix: 0x80);
+        JumpAlways(endTarget);
+
+        replacementTarget.Instruction = Push("\ufffd");
```
Wi1l-B0t, line 843: What's the purpose of this `ConvertCharToUtf8`?

Partial support for UTF8 makes this framework more confusing,
Full support for UTF-8 is too complicated, so it's better not to support it at all.

shargon (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
Jim8y: The scope is conversion of one `char` into the Neo strict UTF-8 string representation, not general UTF-8 support. Isolated UTF-16 surrogates cannot be represented as valid UTF-8, so they are replaced with U+FFFD. Constant and runtime paths now follow the same rule, and the limitation is covered by tests. Converting surrogate chars separately therefore produces replacement characters; supplementary characters must be passed as a string.

Jim8y:
Yes. The previous runtime conversion treated a `char`'s numeric value as raw bytes, so values such as `é`, `Ω`, and `中` produced incorrect string data.

A C# `char` is one UTF-16 code unit, so valid non-surrogate values need at most three UTF-8 bytes. Four-byte Unicode scalar values require a surrogate pair in a `string`; regular string values already preserve them.

Wi1l-B0t (approved):

shargon (approved):

PR #1944, opened 2026-08-05
Fix: diagnose floating-point casts and literals

Jim8y:
## Summary

- diagnose float, double, and decimal casts and literals outside typed declarations
- avoid duplicate diagnostics when declaration analysis already covers the root cause
- preserve the existing diagnostic IDs and guidance

## Validation

- 312 analyzer tests passed
- focused format verification passed
- `git diff --check` passed

Part of #1841.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #1946, opened 2026-08-05
Optimize: less opcodes and GAS for converting `BigInteger` to small signed integer

Wi1l-B0t:
Less opcodes and GAS implementation for converting `BigInteger` to small signed integer


src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
```diff
@@ -464,7 +465,7 @@ private static void HandleBigIntegerRangeCheckedConversion(
     /// </remarks>
     private static void HandleBigIntegerToSByte(MethodConvert methodConvert, SemanticModel model, IMethodSymbol symbol, ExpressionSyntax? instanceExpression, IReadOnlyList<SyntaxNode>? arguments)
     {
-        HandleBigIntegerRangeCheckedConversion(methodConvert, model, symbol, instanceExpression, arguments, sbyte.MinValue, sbyte.MaxValue);
+        HandleBigIntegerSizeCheckedConversion(methodConvert, model, symbol, instanceExpression, arguments, sizeof(sbyte));
```
shargon, line 468: sizeof sbyte it's the same as byte, but the range it's different

shargon (changes requested):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
Wi1l-B0t: So ?

It's OK if the `sbyte` is signed type .

Wi1l-B0t: > sizeof sbyte it's the same as byte, but the range it's different

The UT has already covered the range check.

shargon: if you check by sizeof, sbyte has the same size of byte, but the range it's different

Jim8y: This optimization is only used by signed conversion handlers (`sbyte`, `short`, `int`, and `long`); unsigned conversions still use explicit min/max checks. `Neo.VM.Types.Integer.Size` uses the signed `BigInteger.GetByteCount()`, so `Size <= sizeof(T)` matches the signed range, including `T.MinValue`, and rejects both adjacent out-of-range values. The boundary tests cover these cases.

Wi1l-B0t: > if you check by sizeof, sbyte has the same size of byte, but the range it's different

This has nothing to do with the `byte` type.

shargon: unsigned can be optimized, signed won't

Wi1l-B0t: Only if signed.

Because the `int` type in neo-vm is signed integer type.

You can ask the AI to give a more detailed explanation.

Wi1l-B0t: The unit tests cover edge cases.

Jim8y: Neo.VM.Types.Integer.Size uses signed BigInteger.GetByteCount(), so Size <= sizeof(T) exactly matches the signed range. For example, 127 and -128 use one byte, while 128 and -129 use two; the adjacent boundaries are covered for all four signed types.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
```diff
@@ -451,6 +435,23 @@ private static void HandleBigIntegerRangeCheckedConversion(
...
+            methodConvert.ConvertExpression(model, instanceExpression);
+        if (arguments is not null)
+            methodConvert.PrepareArgumentsForMethod(model, symbol, arguments);
+
+        JumpTarget endTarget = new();
+        methodConvert.Dup();                              // Duplicate value for range check
+        methodConvert.Size();                             // Get the size of the value
+        methodConvert.Push(sizeofSigned);                 // Push the size of the type
+        methodConvert.JumpIfLessOrEqual(endTarget);       // Compare the size of the value to the size of the type
+        methodConvert.Throw();                            // Throw the value if it is out of range
```
Jim8y, line 451: Please preserve the existing overflow exception payload. The previous path dropped the value and threw "Overflow", while this path throws the out-of-range integer itself. The size check is valid, but the optimization should not change observable fault behavior.

Wi1l-B0t: The `HandleBigIntegerRangeCheckedConversion` doesn't throw "Overflow".

Jim8y: You're right. I rechecked the current base, and `HandleBigIntegerRangeCheckedConversion` also throws the original out-of-range value. This optimization preserves that behavior, so this comment is not blocking.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.BigInteger.cs
shargon: But BigInteger use one byte for sign (sometimes)

shargon: image

Wi1l-B0t: So the signed int types can be optimized, unsigned won't.

It takes **5 bytes** to represent `uint.MaxValue` for BigInteger.

But the `sizeof(uint)` is **4**.

Wi1l-B0t: > But BigInteger use one byte for sign (sometimes)

For the int type, the length of `ToByteArray` is less than 4.

shargon (approved):

Jim8y (approved):
Validated against the latest master-n3: the synthetic merge is clean, signed boundary tests pass (37/37), full CI and Codecov are green, and all review threads are resolved.

PR #1947, opened 2026-08-06
Optimize: less opcodes and GAS for `Math.BigMul`

Wi1l-B0t:
Less opcodes and GAS implementation for `Math.BigMul`


ajara87 (approved):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Math.cs
```diff
@@ -294,8 +294,9 @@ private static void HandleMathBigMul(MethodConvert methodConvert, SemanticModel
         JumpTarget endTarget = new();
         methodConvert.Mul();                                       // Multiply the values
         methodConvert.Dup();                                       // Duplicate result for range check
-        methodConvert.Within(long.MinValue, long.MaxValue);        // Check if within long range
-        methodConvert.JumpIfTrue(endTarget);                       // Jump if within range
+        methodConvert.Size();                                      // Get the size of the result
+        methodConvert.Push(sizeof(long));                          // Check if within long range
```
shargon, line 298: What's happend with an ulong value, it's valid

shargon (changes requested):

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Math.cs
Jim8y: This handler is only registered for `Math.BigMul(int, int)`, which returns `long`. A result above `long.MaxValue` is therefore not valid here; the unsigned overload is not registered by this handler.

Wi1l-B0t: Not for ulong.

Only `Math.BigMul(int, int)` is supported now.

Wi1l-B0t: > What's happend with an ulong value, it's valid

Please check again.

shargon, line 298: Again, range can't be checked with size, only if it's unsigned

Wi1l-B0t: Only if it's signed.

You can ask the AI to give a more detailed explanation.

Jim8y: Neo.VM.Types.Integer.Size is based on signed BigInteger.GetByteCount(). Therefore, Size <= 8 accepts exactly long.MinValue..long.MaxValue; both adjacent out-of-range values require nine bytes. This handler is registered only for Math.BigMul(int, int).

shargon (approved):

Jim8y (approved):
Validated against current master-n3: focused Math tests pass (39/39), full CI and Codecov are green, and all review threads are resolved.

PR #1952, opened 2026-08-10
Optimize: less opcodes and GAS for integer popcount

Wi1l-B0t:
Less opcodes and GAS implementation for integer popcount.


ajara87 (approved):
lgtm

shargon (approved):

PR #1954, opened 2026-08-12
Optimize: replace EQUAL+JMPIF with JMPEQ in EmitSwitch case dispatch

ajara87:
`HandleEnumGetName` in `SystemCall.Enum` previously implemented the dispatch loop manually (DUP + Push + JMPNE -> nextCheck + NOP). It has been refactored to call EmitSwitch which emits DUP + Push + JMPEQ per case (removing the intermediate NOP label and replacing the inverted conditional with a direct equality jump).

Tests added

- TestEnumGetName_EmitSwitch_EmitsJmpeqPerCase
- TestEnumGetName_EmitSwitch_GasDeltaIsConstantPerSkippedCase

Removing NUMEQUAL saves 8 opcode price units -> 240 datoshi per evaluated case (with the default execution fee factor of 30).


Wi1l-B0t (dismissed):
`EmitSwitch` is unused at present.

ajara87:
> `EmitSwitch` is unused at present.

You're right that EmitSwitch has no callers currently but since it's a helper in ControlFlowDsl intended to be used by system method handlers applying the correct pattern now avoids introducing a less efficient dispatch if it gets used in the future

Jim8y (changes requested):
`EmitSwitch` has no callers, and the removed tests only compared hand-written scripts without exercising this helper. Please either add a real caller with a focused test that reaches `EmitSwitch`, or defer this optimization until the helper is used. Also update the description: removing `NUMEQUAL` saves 8 opcode price units, or 240 datoshi with the default execution fee factor, per evaluated case.

ajara87:
> `EmitSwitch` has no callers, and the removed tests only compared hand-written scripts without exercising this helper. Please either add a real caller with a focused test that reaches `EmitSwitch`, or defer this optimization until the helper is used. Also update the description: removing `NUMEQUAL` saves 8 opcode price units, or 240 datoshi with the default execution fee factor, per evaluated case.

Done. The removed test was intended to verify the reduction in gas consumption with the new algorithm not to validate the method itself since there was no caller

Wi1l-B0t (dismissed):

Jim8y (changes requested):
HandleEnumGetName now converts every enum constant with Convert.ToInt64(t.ConstantValue). This throws while compiling a valid ulong-backed enum member above long.MaxValue, such as ulong.MaxValue. Please preserve the full integral constant domain when building the BigInteger cases and add a focused regression for a ulong-backed enum.

ajara87:
> HandleEnumGetName now converts every enum constant with Convert.ToInt64(t.ConstantValue). This throws while compiling a valid ulong-backed enum member above long.MaxValue, such as ulong.MaxValue. Please preserve the full integral constant domain when building the BigInteger cases and add a focused regression for a ulong-backed enum.

done

Jim8y (changes requested):
The functional fixes now pass all 21 focused enum tests, including `ulong.MaxValue`, but this head adds a duplicate `using System.Linq;` in `UnitTest_Enum.cs` and the build emits CS0105. Please remove the duplicate import.

shargon (approved):

Jim8y (approved):
Validated against the current master-n3: the focused enum suite passes 21/21 and the full format check is clean. Integral backing coverage is split across the focused tests, and TestEngine resets FeeConsumed before each execution, so the two remaining bot comments are non-blocking.

PR #1955, opened 2026-08-13
Fix: add exception message for `Char.Parse`

Wi1l-B0t:
Add exception message for `Char.Parse`.

`Char.Parse` should throw exception message rather than `null` value or raw value.


ajara87 (dismissed):

Jim8y (changes requested):
Please add focused assertions for the new Char.Parse fault payloads. The regression should verify the Null message for null input and the NotOneChar message for empty and multi-character inputs; the current tests only assert that these calls fault.

Wi1l-B0t:
> Please add focused assertions for the new Char.Parse fault payloads. The regression should verify the Null message for null input and the NotOneChar message for empty and multi-character inputs; the current tests only assert that these calls fault.

Done.

ajara87 (approved):

Jim8y (approved):

shargon (commented):
Lgtm

PR #1961, opened 2026-08-18
Optimize: less opcodes for `EmitSwitch`

Wi1l-B0t:
Less opcodes implementation for `EmitSwitch`.

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/Helpers/ControlFlowDsl.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/Helpers/ControlFlowDsl.cs
@@ -245,10 +245,12 @@ internal partial class MethodConvert
     /// Emits a switch statement. The <paramref name="valueEmitter"/> must push the discriminant value.
     /// Each case body is executed after the switch value has been dropped.
     /// </summary>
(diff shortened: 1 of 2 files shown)


Jim8y (dismissed):
The generated script is smaller, but execution GAS is unchanged: UnitTest_Enum still passes all 21 tests with the existing exact AssertGasConsumed values. The focused skipped-case test also still documents DUP although this implementation now emits LDLOC, and there is no assertion demonstrating a GAS improvement. Please either make and test a real GAS reduction, or narrow this PR to bytecode size and update the focused test/comment.

Wi1l-B0t:
> The generated script is smaller, but execution GAS is unchanged: UnitTest_Enum still passes all 21 tests with the existing exact AssertGasConsumed values. The focused skipped-case test also still documents DUP although this implementation now emits LDLOC, and there is no assertion demonstrating a GAS improvement. Please either make and test a real GAS reduction, or narrow this PR to bytecode size and update the focused test/comment.

It has less opcodes.

The contract will be smaller, so it takes less GAS.

shargon (dismissed):

ajara87 (dismissed):

shargon:
Conflicts

shargon (approved):

ajara87 (approved):

PR #1962, opened 2026-08-20
Optimize: less opcodes for `Enum.Parse` & `Enum.IsDefined`

Wi1l-B0t:
Less opcodes implementation for `Enum.Parse` & `Enum.IsDefined`.

### DIFF
--- a/src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Enum.cs
+++ b/src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Enum.cs
@@ -104,7 +104,6 @@ internal partial class MethodConvert
                 },
                 thenEmitter: () =>
                 {
-                    methodConvert.Drop(2);                       // Stack: [type, inputString]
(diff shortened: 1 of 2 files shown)


ajara87 (approved):

shargon (approved):

Jim8y (changes requested):
Bytecode is smaller, but successful Enum.IsDefined matches now cost 2 additional datoshi: the optimized early RET becomes a JMP plus two NIP operations. The current Enum tests do not assert this size/GAS trade-off. Please add focused byte-size and GAS assertions and document the intended optimization metric.

Wi1l-B0t:
> Bytecode is smaller, but successful Enum.IsDefined matches now cost 2 additional datoshi: the optimized early RET becomes a JMP plus two NIP operations. The current Enum tests do not assert this size/GAS trade-off. Please add focused byte-size and GAS assertions and document the intended optimization metric.


Because the test cases are too simple.

PR #1964, opened 2026-08-20
Preserve expanded params evaluation order

Jim8y:
Expanded params elements were emitted in NeoVM stack order before their side effects were evaluated, reversing C# source order. This change stages argument values left-to-right before constructing the params array and reordering call operands.

Named arguments that occupy their positional slot are now advanced exactly once, so following positional values bind before the expanded params array.

Validation:
- `UnitTest_ArgumentEvaluationOrder`: 4/4 passed
- `Neo.Compiler.CSharp.UnitTests`: 1533/1533 passed
- Release solution build passed
- `dotnet format` verification passed


ajara87:
conflicts @Jim8y

ajara87:
When calling a method with a named argument that occupies its correct positional slot followed by positional arguments and a params parameter (e.g. Combine(a: 1, 2, 3) with Combine(int a, int b, params int[] c)) the result is incorrect.

Fix added on #1968 . @Jim8y, please, could you take a look?

ajara87 (dismissed):

shargon:
conflicts @Jim8y

Jim8y:
Updated the branch to current master-n3, kept the named/positional params regression coverage, and refreshed the generated reentrancy artifact. Please recheck the conflict and params fix.

shargon (approved):

ajara87 (approved):

PR #1968, opened 2026-08-21
Fix positional index

ajara87:
When calling a method with a named argument that occupies its correct positional slot followed by positional arguments and a params parameter (e.g. Combine(a: 1, 2, 3) with Combine(int a, int b, params int[] c)) the result is incorrect.

Expected: a=1, b=2, c=[3]. Actual: a=1, b=3, c=[3] (the value 2 is lost, and b incorrectly takes the last positional argument instead).

Related to #1964


shargon (commented):
Lgtm

PR #1970, opened 2026-08-22
Less opcodes and GAS for `string.EndsWith`

Wi1l-B0t:
Less opcodes and GAS implementation for `string.EndsWith`.


Jim8y (approved):
Reviewed the stack transformation, boundary coverage, and generated artifact. The targeted EndsWith regression test passes locally and CI is green.

shargon (approved):

PR #1972, opened 2026-08-22
Optimize char.IsLetter/char.IsAsciiLetter using ASCII case folding

ajara87:
Replaces the dual range check ('A'-'Z' OR 'a'-'z') in `EmitCharIsAsciiLetter` with a single ASCII case folding chec: (value | 0x20) is within 'a'..'z'. This reduces the emitted opcodes from 9 to 5 per call site saving 390 datoshi of GAS per invocation.


Jim8y (approved):
Reviewed the ASCII folding equivalence, non-ASCII and out-of-range coverage, and generated artifact. The targeted char property tests pass locally and CI is green.

shargon (approved):

Wi1l-B0t:
Good.

PR #1973, opened 2026-08-23
Fix expected gas values

ajara87:
Some expected GAS values are outdated


Wi1l-B0t (dismissed):
Just testing.

Jim8y (changes requested):
Please enable GAS assertions in UnitTest_Char and GuardHelpersTest before updating their expected values. Both classes still set TestGasConsume to false, so most of this PR is not validated.

ajara87:
> TestGasConsume

All changes have been validated by setting all TestGasConsume values to true locally. Afterwards they were restored to their original values to avoid altering the existing behavior. Perhaps they were set to false for a specific reason.

ajara87:
@Jim8y all TestGasConsume has been set to true

shargon:
UT fault @ajara87

ajara87:
> UT fault @ajara87

yes, I'm checking it. It seems is not because of this changes it's because CoverInstruction (I’ll open an issue as soon as I’ve investigated it)

ajara87:
Temporarily MaxControlFlowAnalysisDepth is reduced to 256 due to this issue https://github.com/neo-project/neo-devpack-dotnet/issues/1980

ajara87:
> Temporarily MaxControlFlowAnalysisDepth is reduced to 256 due to this issue #1980

set up to 1024 again.

shargon (approved):

Jim8y (approved):
Verified the current head with GAS assertions enabled: UnitTest_Char, UnitTest_Enum, SyntaxTests, and GuardHelpersTest pass locally, format is clean, and required CI is green.

PR #1974, opened 2026-08-23
Less opcodes and GAS for `string.Insert`

Wi1l-B0t:
less opcodes and GAS implementation for `string.Insert`.


ajara87 (approved):

Jim8y (approved):
Reviewed the stack transformation, null and boundary behavior, generated artifact, and GAS regression coverage. The targeted string tests pass locally and CI is green.

shargon (commented):
Huge optimization

PR #1977, opened 2026-08-24
Less opcodes for `char.IsLetterOrDigit`

Wi1l-B0t:
Less opcodes implementation for `char.IsLetterOrDigit`.

From #1972.


Jim8y (approved):
Reviewed the ASCII folding equivalence, generated artifact, and existing non-ASCII and out-of-range coverage. The targeted char tests pass locally and CI is green.

shargon (approved):

Issue #1980, opened 2026-08-24
Stack overflow risk in recursive InstructionCoverage.CoverInstruction control-flow analysis


ajara87:
### Summary
`InstructionCoverage.CoverInstruction` in `src/Neo.Compiler.CSharp/ControlFlow/InstructionCoverage.cs` performs recursive control-flow traversal with a logical depth guard:

    internal const int MaxControlFlowAnalysisDepth = 1024;
    if (analysisDepth > MaxControlFlowAnalysisDepth)
        throw new BadScriptException($"Control flow analysis depth exceeds {MaxControlFlowAnalysisDepth}");

This guard assumes the CLR call stack can accommodate 1024 recursive frames of `CoverInstruction` before the check fires. In practice this depends on the OS default thread stack size and build configuration (Debug frames are larger/less optimized than Release):

- Windows, Debug config: default thread stack size ~1MB -> StackOverflowException is thrown by the CLR *before* reaching depth 1024, crashing the test process (StackOverflowException cannot be caught).
- Windows, Release config: smaller/optimized frames -> reaches depth 1024, BadScriptException thrown as expected.
- Linux (GitHub Actions ubuntu-latest, Debug config, used by CI): larger default thread stack size -> reaches depth 1024, BadScriptException thrown as expected.

### Repro
1. On Windows, run:
   dotnet test tests/Neo.Compiler.CSharp.UnitTests -c Debug \
     --filter "FullyQualifiedName~InstructionCoverage_ThrowsBadScriptForTooDeepCallChain"
2. Observe StackOverflowException instead of the expected BadScriptException which aborts the whole test host process.

### Test affected
tests/Neo.Compiler.CSharp.UnitTests/Optimizer/ControlFlowValidationTests.cs
  -> InstructionCoverage_ThrowsBadScriptForTooDeepCallChain

### Root cause
Recursive implementation of CoverInstruction ties the logical analysis-depth guard to the physical CLR stack capacity which varies by OS/build config/ thread stack size. StackOverflowException is uncatchable and terminates the process so the guard is not a reliable safety mechanism on all platforms.

### Suggested fix
Refactor CoverInstruction to an iterative algorithm using an explicit Stack<T>/work-list instead of CLR recursion so `MaxControlFlowAnalysisDepth` is enforced purely as a logical counter independent of the thread's physical stack size. Alternatively, run the analysis on a dedicated thread with an explicitly large stack size as a stopgap.

### Environment where reproduced
[...]

PR #1981, opened 2026-08-24
Fix: convert recursive InstructionCoverage.CoverInstruction to an iterative algorithm to prevent StackOverflowException

ajara87:
Related to https://github.com/neo-project/neo-devpack-dotnet/issues/1980

`InstructionCoverage.CoverInstruction` relied on tail self-recursion for most control flow transitions (splitting basic blocks at jump targets TRY/ENDTRY/ENDFINALLY handling unconditional jumps and continuing after a CALL/CALLA returns OK). Because C# does not guarantee tail call optimization contracts with long chains of basic blocks could exhaust the CLR call stack and crash the compiler with a `StackOverflowException` instead of failing gracefully.


Jim8y (changes requested):
CoverInstruction still recurses for CALL/CALL_L targets, so the Windows Debug call-chain repro from #1980 can still overflow before the logical depth guard. Please make call-target traversal iterative and add a regression for the failing chain.

src/Neo.Compiler.CSharp/ControlFlow/InstructionCoverage.cs
```diff
@@ -197,189 +197,389 @@ public BranchType HandleAbort(int entranceAddr, int abortFromAddr, Stack<TryStat
...
-                {
-                    BranchType noJump = CoverInstruction(addr + instruction.Size, tryStack, continueFromBasicBlockEntranceAddr: entranceAddr, analysisDepth: analysisDepth + 1);
-                    BranchType jump = CoverInstruction(ComputeJumpTarget(addr, instruction), tryStack, jumpFromBasicBlockEntranceAddr: entranceAddr, analysisDepth: analysisDepth + 1);
-                    if (noJump == BranchType.OK || jump == BranchType.OK)
+                    if (conditionalJump.Contains(instruction.OpCode) || conditionalJump_L.Contains(instruction.OpCode))
                     {
-                        // See if we are in a try. There may still be runtime exceptions
-                        HandleThrow(entranceAddr, addr, tryStack, analysisDepth);
-                        return BranchType.OK;  // No need to set coveredMap[entranceAddr] because it's OK when covered
+                        BranchType noJump = CoverInstruction(addr + instruction.Size, tryStack, continueFromBasicBlockEntranceAddr: entranceAddr, analysisDepth: frame.AnalysisDepth + 1);
```
cschuchardt88: **[suggestion]** CALL/CALL_L/CALLA and tail block transitions are iterative, but `CoverInstruction` still recurses for both sides of every conditional jump, and `HandleThrow`/`HandleAbort` still recurse into catch/finally. Each of those calls starts a brand-new `Stack<Frame>` and `RunFrame`, so a long `JMPIF` chain (or deeply nested TRY) still consumes CLR stack. #1980 is that Windows Debug (~1MB stack) overflows *before* the logical 1024-depth guard. The new 50k CALL test does not exercise this.

**Suggestion:** Drive conditional branches and `HandleThrow`/`HandleAbort` through the same `Frame`/`PendingKind` machine so `analysisDepth` is the only stack that matters. Add a JMPIF-chain analogue of `InstructionCoverage_ThrowsBadScriptInsteadOfOverflowingForVeryDeepCallChain`.

src/Neo.Compiler.CSharp/ControlFlow/InstructionCoverage.cs
```diff
@@ -197,189 +197,389 @@ public BranchType HandleAbort(int entranceAddr, int abortFromAddr, Stack<TryStat
         /// <returns>Whether it is possible to return without exception</returns>
         /// <exception cref="BadScriptException"></exception>
         /// <exception cref="NotImplementedException"></exception>
+        /// <summary>
```
cschuchardt88: **[nit]** Several new comments narrate the rewrite rather than a non-obvious invariant: Frame as “virtual invocation of the former recursive algorithm”, `TailChainEntrances` restating the old assignment pattern, and the empty-CALLA comment about the initial `ABORT` default. The CALLA `TODO` is also sitting after `return null`, so it is unreachable.

**Suggestion:** Drop the history-of-the-diff comments. Keep only constraints that are not obvious from the code. Put the CALLA TODO back on the live target-iteration path.

src/Neo.Compiler.CSharp/ControlFlow/InstructionCoverage.cs
Jim8y: Addressed in 36efadae. Conditional branches and exception handlers now use the shared frame stack, with deep JMPIF and TRY regressions.

src/Neo.Compiler.CSharp/ControlFlow/InstructionCoverage.cs
Jim8y: Addressed in 36efadae. The rewrite-history comments are removed and the CALLA TODO is on the live iteration path.

Jim8y (approved):
The remaining recursive paths now use the shared frame stack, and the depth regressions cover conditional and TRY traversal. The required check passes.

Wi1l-B0t (approved):

shargon (commented):
Lgtm

PR #1983, opened 2026-08-25
Fix signed remainder overflow handling

Jim8y:
## Summary
- handle `int.MinValue % -1` and `long.MinValue % -1` consistently with C#
- apply the same overflow handling to compound remainder assignments
- preserve caller-owned stack values across ordinary and aggressive-inline catch paths
- keep stack-depth snapshots isolated from reusable temporary slots

## Tests
- Neo.Compiler.CSharp.UnitTests: 1541 passed
- remainder overflow tests with None and All optimization: 6 passed
- Release solution build
- dotnet format


src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -342,7 +351,7 @@ private void CheckDivideOverflow(SemanticModel model, ITypeSymbol? type, Express
         Push(minValue.Value);
         JumpIfNotEqual(endTarget);
 
-        Drop(2);
+        Clear();
```
shargon: Drop 2 was wrong?

cschuchardt88: **[bug]** `Drop(2)` was insufficient for compound assignment and nested expressions. Replacing it with `Clear()` fixes same-method catch tests, but NeoVM `ExecutionContext.Clone()` shares `EvaluationStack` across `CALL`. `Clear()` therefore deletes the caller’s pending operands.

Example the new tests do not cover: `100 + SafeMod(int.MinValue, -1)` where `SafeMod` catches overflow and returns 0. After `Compute` pushes `100` and `CALL`s `SafeMod`, `CLEAR` wipes `[100]`. Result FAULTs or is `0` instead of `100`. `CheckLeftShiftOverflow` still uses `Drop(2)`, now inconsistent with division/remainder.

**Suggestion:** Do not `CLEAR` the shared evaluation stack. Restore to the depth at the matching `TRY` (snapshot `DEPTH` at `TRY` / drop `DEPTH - tryDepth` before `THROW`), or pass an explicit extra-drop count from compound assignment. Add a test where a helper catches `MinValue % -1` (and `MinValue / -1`) and is used as an operand.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_RemainderOverflow.cs
```diff
@@ -0,0 +1,317 @@
...
+                     "Contract.UncheckedLong(",
+                     "Contract.CompoundInt(",
+                     "Contract.CompoundLong("
+                 })
+        {
+            CollectionAssert.Contains(GetMethodOpCodes(context, method), OpCode.THROW);
+        }
+
+        var contract = Deploy(context);
+        Assert.ThrowsExactly<TestException>(() => contract.CheckedInt(int.MinValue, -1));
```
cschuchardt88: **[nit]** `SignedRemainderOverflowThrowsInAllContexts` only asserts `ThrowsExactly<TestException>`. `UnitTest_DivisionOverflow` also checks `Assert.Contains("Overflow", ex.Message)`. A wrong throw (divide-by-zero, `RET` mismatch after a polluted stack) would still satisfy the remainder tests.

**Suggestion:** Assert the exception message contains `"Overflow"` for the uncaught `MinValue % -1` cases.

cschuchardt88 (changes requested):

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
Wi1l-B0t: Why `Clear`?

`CLEAR` clears all items from the evaluation stack, and is more expensive than `DROP`.

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs
```diff
@@ -204,6 +204,40 @@ private void InsertLocalSlotInitialization()
...
+            insertionIndex = _instructions.Count;
+
+        byte index = _methodStackDepthIndex.Value;
+        var store = index >= 7
+            ? new Instruction { OpCode = OpCode.STLOC, Operand = [index] }
+            : new Instruction { OpCode = OpCode.STLOC0 + index };
+        _instructions.InsertRange(insertionIndex, [new Instruction { OpCode = OpCode.DEPTH }, store]);
+    }
+
+    private void RestoreMethodStackDepth()
```
cschuchardt88, line 236: **[bug]** `RestoreMethodStackDepth` restores to the snapshot taken at the enclosing `MethodConvert` entry (`InsertMethodStackDepthInitialization` after `INITSLOT`). That is correct for `CALL` because NeoVM `ExecutionContext.Clone()` shares `EvaluationStack`, so callee-entry `DEPTH` already includes the caller's pending operands. It is not correct for `TryProcessInlineMethods`: AggressiveInlining converts the callee body with the caller's `MethodConvert`, so the snapshot is the outer method's entry depth. Overflow inside an inlined `try/catch` therefore `DROP`s the caller's pending operands before `THROW`. `100 + InlinedSafeMod(int.MinValue, -1)` then faults or returns `0` instead of `100`. `CallerOwnedOperand` only covers the non-inlined `CALL` path.

**Suggestion:** Snapshot depth at the inlined fragment (and/or matching `TRY`) rather than only at method entry, and restore to the innermost snapshot. Add a test with `[MethodImpl(MethodImplOptions.AggressiveInlining)]` on a helper that catches `MinValue % -1` / `MinValue / -1` and is used as an operand (`100 + SafeMod(...)`) under both None and All optimization.

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs
```diff
@@ -204,6 +204,40 @@ private void InsertLocalSlotInitialization()
         }
     }
 
+    private void EnsureMethodStackDepth()
```
cschuchardt88, line 217: **[suggestion]** This PR and sibling neo-devpack-dotnet#1984 independently add the same `_methodStackDepthIndex` / `EnsureMethodStackDepth` / `InsertMethodStackDepthInitialization` / `RestoreMethodStackDepth` helpers (plus the `Convert()` insertion). The two copies are currently identical, so git may auto-merge, but any later tweak in one PR will conflict or leave two slightly different restore implementations.

**Suggestion:** Land the snapshot/restore helper in one PR first (or extract it to a shared commit) and have the other reuse it. Keep a single restore policy for `/`, `%`, and DivRem.

cschuchardt88 (changes requested):

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -342,7 +351,7 @@ private void CheckDivideOverflow(SemanticModel model, ITypeSymbol? type, Express
         Push(minValue.Value);
         JumpIfNotEqual(endTarget);
 
-        Drop(2);
+        RestoreMethodStackDepth();
```
shargon, line 498: why 2 is not fixed?

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_DivisionOverflow.cs
```diff
@@ -28,18 +28,23 @@ public abstract class Contract_DivisionOverflow(Neo.SmartContract.Testing.SmartC
...
     /// DUP [2 datoshi]
     /// PUSHM1 [1 datoshi]
-    /// JMPNE 17 [2 datoshi]
+    /// JMPNE 1C [2 datoshi]
     /// OVER [2 datoshi]
     /// PUSHINT32 00000080 [1 datoshi]
-    /// JMPNE 0F [2 datoshi]
-    /// DROP [2 datoshi]
+    /// JMPNE 14 [2 datoshi]
+    /// DEPTH [2 datoshi]
```
Wi1l-B0t, line 43: Why add DEPTH?

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs
```diff
@@ -204,6 +215,58 @@ private void InsertLocalSlotInitialization()
...
+        if (insertionIndex < 0)
+            insertionIndex = _instructions.Count;
+
+        byte index = _methodStackDepthIndex.Value;
+        var store = StoreLocalInstruction(index);
+        _instructions.InsertRange(insertionIndex, [new Instruction { OpCode = OpCode.DEPTH }, store]);
+    }
+
+    private void RestoreMethodStackDepth()
+    {
```
Wi1l-B0t, line 237: Why add `RestoreMethodStackDepth`?

ajara87:
@Jim8y conflicts

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs
Jim8y: Added an inline-boundary depth snapshot and an `AggressiveInlining` regression under both `None` and `All`.

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
Jim8y: `DROP 2` only removes the operands. Nested and compound expressions can leave additional temporaries, so the overflow path restores the nearest inline or method depth before throwing.

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs
Jim8y: NeoVM `THROW` does not unwind the evaluation stack. Restoring the saved depth removes operands and any nested temporaries without dropping caller-owned values.

Wi1l-B0t (approved):

shargon (approved):

ajara87 (approved):

PR #1984, opened 2026-08-25
Fix signed DivRem overflow handling

Jim8y:
## Summary
- add signed overflow checks to `int.DivRem`, `long.DivRem`, and the corresponding `Math.DivRem` overloads
- omit the guard when constant operands prove the operation safe
- preserve caller-owned stack values across ordinary and aggressive-inline catch paths
- keep stack-depth snapshots isolated from reusable temporary slots

## Tests
- Neo.Compiler.CSharp.UnitTests: 1541 passed
- DivRem overflow tests with None and All optimization: 6 passed
- Release solution build
- dotnet format


shargon (dismissed):

ajara87 (dismissed):

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Integer.cs
```diff
@@ -54,12 +54,10 @@ public void divRemInt_test()
...
                     Assert.AreEqual(expected.Remainder, (int)(BigInteger)result[0]);
                     Assert.AreEqual(expected.Quotient, (int)(BigInteger)result[1]);
-                    AssertGasConsumed(1109280);
+                    AssertGasConsumed(1109430);
                 }
 
-            var overflowResult = Contract.DivRemByte(int.MinValue, -1)!;
-            Assert.AreEqual(0, (int)(BigInteger)overflowResult[0]);
-            Assert.AreEqual(-(BigInteger)int.MinValue, (BigInteger)overflowResult[1]);  // no exception
+            Assert.ThrowsExactly<TestException>(() => Contract.DivRemInt(int.MinValue, -1));
```
cschuchardt88: **[suggestion]** Overflow tests only cover `MinValue / -1` (must throw) and `±10 / ±3` (must succeed). They never run `int.MinValue` or `long.MinValue` with a divisor other than `-1`, and `CaughtInt` / `CaughtMathLong` never run the non-overflow path (`7 + quotient`). A swapped second `JMPNE` (divisor vs `-1`) would still pass the current suite.

**Suggestion:** Add `DivRemInt(int.MinValue, 1)`, `DivRemInt(int.MinValue, -2)` (quotient `1073741824`, remainder `0`), and the long equivalents; also `CaughtInt(10, 3) == 10` and `CaughtMathLong(10, 3) == 10`. Optionally assert the fault message is `"Overflow"`.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Math.cs
```diff
@@ -242,6 +243,60 @@ private static void HandleMathULongDivRem(MethodConvert methodConvert, SemanticM
...
+        var endTarget = new JumpTarget();
+        methodConvert.Dup();
+        methodConvert.Push(minValue);
+        methodConvert.JumpIfNotEqual(endTarget);
+
+        methodConvert.Over();
+        methodConvert.Push(-1);
+        methodConvert.JumpIfNotEqual(endTarget);
+
+        methodConvert.Clear();
```
Wi1l-B0t: Why `Clear()`?

`CLEAR` clears all items from the evaluation stack.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_Integer.cs
```diff
@@ -71,7 +69,8 @@ public void divRemLong_test()
             var expected = long.DivRem(10, 3);
             Assert.AreEqual(expected.Remainder, (long)(BigInteger)result[0]);
             Assert.AreEqual(expected.Quotient, (long)(BigInteger)result[1]);
-            AssertGasConsumed(1109280);
+            AssertGasConsumed(1109430);
+            Assert.ThrowsExactly<TestException>(() => Contract.DivRemLong(long.MinValue, -1));
```
Wi1l-B0t: Add exception message assert.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.Math.cs
Wi1l-B0t: `CLEAR` is more expensive than `DROP`.

src/Neo.Compiler.CSharp/MethodConvert/Helpers/ConvertHelpers.cs
```diff
@@ -204,6 +204,40 @@ private void InsertLocalSlotInitialization()
...
+        _instructions.InsertRange(insertionIndex, [new Instruction { OpCode = OpCode.DEPTH }, store]);
+    }
+
+    private void RestoreMethodStackDepth()
+    {
+        EnsureMethodStackDepth();
+        var checkTarget = new JumpTarget();
+        var endTarget = new JumpTarget();
+        checkTarget.Instruction = Depth();
+        LdLoc(_methodStackDepthIndex!.Value);
```
cschuchardt88: **[suggestion]** `RestoreMethodStackDepth` drops back to the depth saved at method entry (`DEPTH`/`STLOC` inserted after `INITSLOT`). That is correct when the overflowing `DivRem` is in the same `MethodConvert` as the `try` (e.g. `CaughtInt`: `7 + int.DivRem(...)`) or in a separately CALLed helper (the helper has its own isolated evaluation stack). It is wrong for source-level inlining: `TryProcessInlineMethods` converts an `[MethodImpl(MethodImplOptions.AggressiveInlining)]` helper into the caller, so `EnsureMethodStackDepth` allocates on the caller and the snapshot is the caller's entry depth (0). Values pushed before the inlined `try` — the `100` in `100 + Safe(...)` — are then dropped, and the catch path returns `0` or FAULTs on `ADD` instead of `100`. Bytecode inlining that copies the helper's own `DEPTH` at the call site would still work; source inlining would not. The new `CallerInt` / `CallerMathLong` tests do not use `AggressiveInlining`, so this gap is untested.

**Suggestion:** Snapshot the depth at `TRY` (or at the inlined-fragment start) rather than at method entry, and restore to that. A regression test matching `CallerInt` plus `[MethodImpl(MethodImplOptions.AggressiveInlining)]` on `SafeIntHelper` should expect `100` for `(int.MinValue, -1)`.

ajara87:
Since ExecuteThrow in NeoVM (JumpTable.Control.cs) discards the entire invocation frame including its evaluation stack when no enclosing try is found in the current context, the DEPTH/STLOC prologue and the RestoreMethodStackDepth cleanup before THROW are only actually required when the guarded DivRem call is lexically inside a try/catch in the same method. When there's no enclosing try the frame is dropped on overflow anyway so nothing needs restoring.

Since _tryStack already tracks this nesting at compile time, could EmitSignedDivRemOverflowGuard skip EnsureMethodStackDepth() / RestoreMethodStackDepth() (and thus the per-method DEPTH/STLOC prologue) when _tryStack.Count == 0 at the call site? That would remove the gas overhead entirely for methods without a wrapping try/catch which is likely the common case.

tests/Neo.Compiler.CSharp.UnitTests/TestingArtifacts/Contract_Math.cs
```diff
@@ -346,11 +346,26 @@ public abstract class Contract_Math(Neo.SmartContract.Testing.SmartContractIniti
...
+    /// STLOC0 [2 datoshi]
     /// LDARG1 [2 datoshi]
     /// LDARG0 [2 datoshi]
     /// DUP [2 datoshi]
+    /// PUSHINT32 00000080 [1 datoshi]
+    /// JMPNE 18 [2 datoshi]
+    /// OVER [2 datoshi]
+    /// PUSHM1 [1 datoshi]
+    /// JMPNE 14 [2 datoshi]
+    /// DEPTH [2 datoshi]
```
Wi1l-B0t, line 361: Why `DEPTH`?

cschuchardt88: NeoVM `THROW` does not unwind leftover evaluation-stack items. On `int.MinValue / -1` the overflow path still has the DivRem operands (and any caller-owned values, because `CALL` shares `EvaluationStack`) sitting on the stack. A later `catch` would then see the wrong stack.

The compiler snapshots `DEPTH` into a local at method entry (`DEPTH` / `STLOC0` after `INITSLOT`). This `DEPTH` is the restore loop:

```
DEPTH
LDLOC0      // saved method-entry depth
JMPLE end
DROP
JMP back
```

until the stack is back to that depth, then `PUSH "Overflow"` / `THROW`. A fixed `DROP`/`DROP` is not enough when there are extra items above the two operands.

cschuchardt88:
@ajara87 `ExecuteThrow` does walk the whole invocation stack and FAULTs if no try exists anywhere, so leftover eval-stack items do not matter for an uncaught overflow. Skipping the `DEPTH`/`STLOC` prologue in that case is a real gas win.

`_tryStack.Count == 0` at the DivRem site is the wrong predicate for that. NeoVM `CALL` is `CurrentContext.Clone()`, and clones share `EvaluationStack`. `ContextUnloaded` does not drop eval-stack items — it only clears locals/args. So:

```csharp
try {
    return 100 + Helper(int.MinValue, -1); // Helper has no try of its own
} catch {
    return 0;
}
```

THROW in `Helper` finds the caller's catch, pops the helper frame, then `Push`es the exception onto the **shared** stack. Without the restore loop, the DivRem operands are still under the exception and the catch sees the wrong stack.

The compiler `_tryStack` is per `MethodConvert`. It is empty in `Helper` even though a caller in the same script can catch. Same hole if `TryProcessInlineMethods` converts a no-try helper into a method that already has a `try`.

A skip is safe only when no enclosing try can catch on this invocation stack — which you cannot see from `_tryStack` of the current method alone. Whole-contract call-graph analysis could do it; `_tryStack.Count == 0` cannot.

shargon (approved):

Wi1l-B0t (approved):

ajara87 (approved):

PR #1986, opened 2026-08-25
Fix lifted nullable binary operator semantics

Jim8y:
## Summary
- preserve null propagation for lifted arithmetic and bitwise operators
- implement C# three-valued semantics for nullable Boolean `&` and `|`
- apply the same behavior to compound assignments without changing evaluation order
- handle lifted `BigInteger` operations correctly

## Tests
- added coverage for integral types, nullable enums, `BigInteger`, Boolean operators, compound assignments, evaluation order, overflow checks, and both optimization modes
- passed all 1,542 compiler unit tests


shargon (dismissed):

ajara87 (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -150,6 +165,120 @@ type is INamedTypeSymbol
...
+        LdLoc(leftSlot);
+        IsNull();
+        JumpIfTrue(nullTarget);
+        LdLoc(rightSlot);
+        IsNull();
+        JumpIfTrue(nullTarget);
+
+        LdLoc(leftSlot);
+        LdLoc(rightSlot);
+        EmitBinaryOperator(model, expression, model.GetTypeInfo(expression).Type!);
```
cschuchardt88, line 229: **[bug]** Lifted `bool? ^ bool?` goes through `ConvertLiftedBinaryExpression` (operator `^` is in the lifted set; the three-valued path is only `&`/`|`). The emit type is `bool?`. `GetStackItemType` does not unwrap `Nullable<T>`, so `bool?` becomes `StackItemType.Any` and `EmitBinaryOperator` skips the XOR→Boolean `Nz()` fix. NeoVM `XOR` yields Integer 0/1. Direct `BoolXor` tests still pass via engine coercion; `(true ^ true) & true` becomes `true` because Integer 0 is not `EQUAL` to `PUSHF`. Same hole for `^=`.

**Suggestion:** Unwrap `Nullable<T>` before `GetStackItemType` / `isBoolean`. Pass underlying `bool` into `EmitBinaryOperator` so `Nz()` runs. Add tests for `(true ^ true) & true`, `(false ^ false) | false`, and `bool? ^= bool?` with both non-null.

src/Neo.Compiler.CSharp/MethodConvert/Expression/BinaryExpression.cs
```diff
@@ -150,6 +165,120 @@ type is INamedTypeSymbol
...
+        var decisiveTarget = new JumpTarget();
+        var nullTarget = new JumpTarget();
+        var endTarget = new JumpTarget();
+
+        LdLoc(leftSlot);
+        if (isAnd)
+            PushF();
+        else
+            PushT();
+        AddInstruction(OpCode.EQUAL);
```
cschuchardt88: **[suggestion]** Three-valued `&`/`|` detect a decisive operand with `EQUAL` against `PUSHF`/`PUSHT`. That is correct for ABI/literal `bool?` values (Boolean stack items). It is not correct for Integer 0/1, which `JMPIF`/`JMPIFNOT` would treat as false/true. Combined with the XOR issue, any Integer-typed `bool?` is neither decisive nor null and takes the fallthrough. Tests only cover null mixed with a Boolean, plus `BoolXor` returning to the engine.

**Suggestion:** After `ISNULL`, use `JMPIFNOT` (AND) / `JMPIF` (OR). Extend tests with the full 3×3 tables for `&`/`|`/`^` (including both-non-null) and mixed `bool & bool?`.

cschuchardt88 (changes requested):

Wi1l-B0t (dismissed):

src/Neo.Compiler.CSharp/MethodConvert/Expression/AssignmentExpression.ComplexAssignment.cs
```diff
@@ -274,8 +274,56 @@ private void EmitComplexAssignmentOperator(SemanticModel model, ITypeSymbol type
...
+        LdLoc(leftSlot);
+        LdLoc(rightSlot);
+        EmitComplexAssignmentOperatorCore(model, type, operatorToken, right, type.GetStackItemType());
+        Jump(OpCode.JMP_L, endTarget);
+
+        nullTarget.Instruction = PushNull();
+        endTarget.Instruction = Nop();
+    }
+
+    private void EmitComplexAssignmentOperatorCore(SemanticModel model, ITypeSymbol type, SyntaxToken operatorToken, ExpressionSyntax right, VM.Types.StackItemType itemType)
```
cschuchardt88: **[nit]** `EmitComplexAssignmentOperatorCore` takes `itemType` and immediately overwrites it with `GetStackItemType(GetNonNullableValueType(type))`. Callers therefore pass a meaningless value (`type.GetStackItemType()` on `int?`/`bool?` is `Any`, not the underlying integer/boolean). Harmless, but the extra parameter is dead after the unwrap that this PR added.

**Suggestion:** Drop the `itemType` parameter and compute it once after `GetNonNullableValueType`.

cschuchardt88 (approved):

ajara87 (approved):

Wi1l-B0t (approved):

PR #1989, opened 2026-08-28
Move `HandleLength` to `System.Array.cs`.

Wi1l-B0t:
`HandleLength` is a handle for `Array` rather than `char`.


Jim8y (approved):
Verified the pure handler relocation, array-length coverage, exact-head CI, and Codecov.

shargon (approved):

ajara87 (approved):

PR #1990, opened 2026-08-30
Don't convert bool contant value to integer

Wi1l-B0t:
Don't convert bool contant value to integer.
It's unnecessary and may cause unexpected behavior.


ajara87 (dismissed):

Jim8y (dismissed):
Please keep the helper-level regression and change it to assert that both bool and string constants are rejected. Deleting the test removes coverage for the intended bool behavior and the existing string guard, so this PR does not prove the change.

Wi1l-B0t:
> Please keep the helper-level regression and change it to assert that both bool and string constants are rejected. Deleting the test removes coverage for the intended bool behavior and the existing string guard, so this PR does not prove the change.

Done.

ajara87 (approved):

shargon (approved):

PR #1996, opened 2026-09-03
Fix:  `BigInteger.ToByteArray` should return [0x00] if zero

Wi1l-B0t:
`BigInteger.ToByteArray` should return [0x00] if zero, but it returns empty byte array now.


shargon (commented):
Empty is cheaper, and is also 0 and false

Wi1l-B0t:
> Empty is cheaper, and is also 0 and false

> If the value is zero, returns an array of one byte whose element is 0x00.
https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.tobytearray?view=net-10.0

Jim8y (approved):

ajara87 (approved):

PR #1999, opened 2026-09-05
Fix class override virtual table construction and lookup

Jim8y:
Base-typed calls on derived class instances could fault because overrides were omitted from virtual tables. Instances with additional derived fields also read the wrong table slot. Include class overrides in the table and locate it after the runtime instance's fields, preserving existing record and value-type layouts and direct override calls.

Add deployed None/All regressions for inheritance, sealed overrides, fields, initializers, base calls, properties and record/struct compatibility. Regenerate affected artifacts and update 12 exact GAS expectations from measured execution. Validation on master-n3 4330bf7d: all 28 new cases, 128 Integer tests and 12 existing Polymorphism/Reentrancy tests pass.


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
Lgtm

PR #2000, opened 2026-09-05
Preserve static initializers when constant folding is invalidated

Jim8y:
The optimizer could delete a static field's initial value even after a later initializer write invalidated constant folding. Build the deletion set from the final valid candidates so both constant and computed rewrites retain the required initialization.

Add 12 deployed regressions across None, Experimental and All. Preserve the regenerated StaticConstruct artifact and its measured 90-datoshi GAS increase. Validation on master-n3 4330bf7d: all 12 regressions and 128 Integer tests pass; four regressions fail on the original source.


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
Lgtm

PR #2002, opened 2026-09-05
Preserve instance receiver evaluation before call arguments

Jim8y:
Evaluate instance receivers before arguments, including string handlers, delegate calls, property getters, and target-typed construction. Preserve calling-convention stack layout and release receiver temporaries after the call.

Validation: 33 focused regressions and all 1,879 compiler tests pass locally. The target-typed construction regression fails before the fix under both None and All optimization. No CI configuration changes.


ajara87:
The receiver-order fix is bypassed for special methods.

`TryProcessSpecialMethods` runs before `preserveInstanceEvaluationOrder` so system/inline handlers still control their own evaluation order. Several supported string handlers currently evaluate the arguments before the receiver. For example, `HandleStringStartsWith` calls `PrepareArgumentsForMethod` before `ConvertExpression(instanceExpression)`.

This still violates C# receiver-before-arguments semantics. A deterministic repro is:

private static string marker;

private static string GetReceiver()
{
    marker = "a";
    return "abc";
}

public static bool Test()
{
    marker = "b";
    return GetReceiver().StartsWith(marker);
}

C# evaluates `GetReceiver()` first so marker becomes "a" before the argument is read and the result must be true. The current special handler reads "b" first and returns false.

The delegate Invoke special path has the same issue: it currently prepares the arguments before evaluating instanceExpression.

Added fix to this on https://github.com/Jim8y/neo-devpack-dotnet/pull/1

ajara87:
Conflicts @Jim8y

Wi1l-B0t:
Conflicts.

Jim8y:
@ajara87 Fixed target-typed `new()` receiver ordering and added delegate regressions. All 1,879 compiler tests pass locally. Please recheck the latest commit.

ajara87:
@Jim8y conflicts

Jim8y:
Rebased onto current master-n3 (including #2003) and resolved the merge conflict. The receiver-order and ref/out regression suites pass locally (84 tests). I also removed the nullable namespace warning in the touched helper. Please re-review commit b542447.

shargon (approved):

ajara87 (approved):

PR #2004, opened 2026-09-05
Keep unsigned bitwise complements within their declared width

Jim8y:
Unsigned bitwise complement returned the VM's signed INVERT result, so complementing uint or ulong zero produced -1. Mask the result to its unsigned width, including checked expressions and non-null nullable operands, while preserving signed and promoted small-integer behavior.

Validation on master-n3 4330bf7d: 18 deployed None/All regression cases and 128 Integer tests pass. Coverage includes boundary values, BigInteger controls and single operand evaluation. Nullable null propagation remains a separate fix.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

PR #2014, opened 2026-09-05
Fix: execution fault if value is zero for `Helper.ToByteArray`

Wi1l-B0t:
Execution fault if value is zero for `Helper.ToByteArray(this byte source)`, 
and unexpected result if value is zero for  `Helper.ToByteArray(this sbyte source)`.


ajara87:
UT Fails

ajara87 (approved):

Jim8y (approved):

shargon (commented):
Lgtm

PR #2015, opened 2026-09-05
Less GAS for pushing constant Buffer

Wi1l-B0t:
Less GAS implementation for pushing constant Buffer.


shargon (dismissed):
if left is cheaper than change type, we should check the type of change type

Wi1l-B0t:
> if left is cheaper than change type, we should check the type of change type

 Check the **price** of change type ?

Jim8y (commented):
Please add a focused runtime test for the `new byte[] { ... }` path and an empty constant array, and verify that the resulting value remains a mutable Buffer. Also document the current `CONVERT Buffer` versus `LEFT` price assumption, since the earlier pricing question is still relevant.

Wi1l-B0t:
> Please add a focused runtime test for the `new byte[] { ... }` path and an empty constant array, and verify that the resulting value remains a mutable Buffer. Also document the current `CONVERT Buffer` versus `LEFT` price assumption, since the earlier pricing question is still relevant.

Added.

ajara87 (approved):

shargon:
> > if left is cheaper than change type, we should check the type of change type
> 
> Check the **price** of change type ?

This optimization should be unnecessary when dynamic pricing by trade code comes into effect this month. Convert should be cheaper than this push an cat

Jim8y (changes requested):
Please add focused runtime coverage for both non-empty and empty constant byte-array paths, asserting the returned value is a mutable Buffer. Also document the current CONVERT Buffer versus LEFT price assumption, since dynamic opcode pricing may change this trade-off.

PR #2017, opened 2026-09-06
Less GAS for converting `string`/`ByteString` to `BigInteger`

Wi1l-B0t:
Less GAS implementation for converting `string`/`ByteString` to `BigInteger`.


shargon (commented):
Probably this and https://github.com/neo-project/neo-devpack-dotnet/pull/2015 will be optimized with dynamic price, @roman-khimov do you know that this will be cheaper?

ajara87 (approved):

Jim8y (commented):
The conversion sequence is a substantial GAS optimization. Before approval, please confirm the `PUSH0 + ADD` path matches the previous `CONVERT Integer` behavior for null, empty, zero, negative, and multi-byte values, including a missing `GetIntegerOrZero` key. Shargon's dynamic-price question should also be answered in the PR.

Wi1l-B0t:
> The conversion sequence is a substantial GAS optimization. Before approval, please confirm the `PUSH0 + ADD` path matches the previous `CONVERT Integer` behavior for null, empty, zero, negative, and multi-byte values, including a missing `GetIntegerOrZero` key. Shargon's dynamic-price question should also be answered in the PR.

Covered by previous tests.

Jim8y (changes requested):
Please add focused runtime coverage proving the PUSH0 + ADD path matches CONVERT Integer for null, empty, zero, negative, and multi-byte values, including a missing GetIntegerOrZero key. The PR should also answer Shargon’s dynamic-price question with measured opcode pricing or an explicit current-network assumption.

PR #2019, opened 2026-09-07
Less GAS for `AccessControl.RoleBytes`

Wi1l-B0t:
Less GAS implementation for `AccessControl.RoleBytes`.

### DIFF
--- a/src/Neo.SmartContract.Framework/AccessControl.cs
+++ b/src/Neo.SmartContract.Framework/AccessControl.cs
@@ -74,8 +74,9 @@ namespace Neo.SmartContract.Framework
         private static byte[] CountKey(BigInteger role)
             => new byte[] { TAG_COUNT }.Concat(RoleBytes(role));
 
-        private static byte[] RoleBytes(BigInteger role)
-            => (byte[])(ByteString)role;
+        [OpCode(OpCode.PUSHDATA1, "00")] // Empty string
+        [OpCode(OpCode.CAT)]
+        private static extern byte[] RoleBytes(BigInteger role);
 
         private static void ValidateRole(BigInteger role)
             => ExecutionEngine.Assert(role >= 0, "AccessControl: role must be non-negative");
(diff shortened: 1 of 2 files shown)


ajara87:
`RoleBytes(0)` may no longer match the previous encoding: `HandleBigIntegerToByteArray` special cases zero to return `[0x00]` (mirroring .NET's `BigInteger.ToByteArray()`), while the old `(byte[])(ByteString)role` cast likely produced an empty array for zero (NeoVM's minimal encoding). Since `role == 0` is `DEFAULT_ADMIN_ROLE` this could shift the storage key for the root role and break compatibility with existing deployments. Could you add a test comparing both encodings for `role = 0` (and a couple of small non-zero values) to confirm this is a no-op change?

Wi1l-B0t:
> `RoleBytes(0)` may no longer match the previous encoding: `HandleBigIntegerToByteArray` special cases zero to return `[0x00]` (mirroring .NET's `BigInteger.ToByteArray()`), while the old `(byte[])(ByteString)role` cast likely produced an empty array for zero (NeoVM's minimal encoding). Since `role == 0` is `DEFAULT_ADMIN_ROLE` this could shift the storage key for the root role and break compatibility with existing deployments. Could you add a test comparing both encodings for `role = 0` (and a couple of small non-zero values) to confirm this is a no-op change?

Fixed.

ajara87 (dismissed):

Jim8y (commented):
The zero-role compatibility concern was identified, but the latest update does not add an explicit regression test for the legacy `RoleBytes(0)` storage key. Please add coverage for role 0 and a few small non-zero roles and verify the resulting `AdminKey`/`CountKey` bytes before approval.

Wi1l-B0t:
> The zero-role compatibility concern was identified, but the latest update does not add an explicit regression test for the legacy `RoleBytes(0)` storage key. Please add coverage for role 0 and a few small non-zero roles and verify the resulting `AdminKey`/`CountKey` bytes before approval.

Covered by previous tests.

`DEFAULT_ADMIN_ROLE` is zero.

shargon (commented):
We should check convert, convert should be cheaper than this with dynamic opcode price

Wi1l-B0t:
> We should check convert, convert should be cheaper than this with dynamic opcode price

The previous implementation has **two** CONVERTs. 
So it's expensive.

Jim8y (changes requested):
Please add an explicit AccessControl regression that compares the legacy and optimized storage keys for role 0 plus a few small non-zero roles, asserting the resulting AdminKey/CountKey bytes. Existing generic tests do not prove DEFAULT_ADMIN_ROLE compatibility, so this remains a blocker.

PR #2026, opened 2026-09-11
Fix: don't load trim-char by LdArg1

Wi1l-B0t:
Don't load trim-char by `LdArg1`.
Cannot load argument by LdArg for system string methods.


Wi1l-B0t:
PR **2026**  🎉.

ajara87 (approved):

shargon (approved):

shargon:
> PR **2026** 🎉.

No more until next year 😛

PR #2027, opened 2026-09-13
Less opcodes and GAS for string to lower

Wi1l-B0t:
Less opcodes and GAS for string to lower.
the `SetItem` is not much expensive than `Append` after dynamic pricing.


ajara87 (approved):
lgtm

Jim8y (approved):
Reviewed the buffer-copy and byte-level casing loop, including empty, NUL, and UTF-8 preservation coverage. The implementation and CI results look correct.

shargon (commented):
Lgtm

PR #2033, opened 2026-09-13
Fix: wrong result if pad with `'\0'` for `string.PadLeft/PadRight`

Wi1l-B0t:
Wrong result if pad with `'\0'` for `string.PadLeft/PadRight`.


ajara87 (approved):

Jim8y (approved):
The NUL-padding fix is scoped, the focused regression coverage is present, and the source build plus required CI checks are green. Approved.

shargon (commented):
Lgtm

PR #2034, opened 2026-09-13
Add `string.LastIndexOf(char)`

Wi1l-B0t:
Add `string.LastIndexOf(char)` support.


src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
```diff
@@ -1970,4 +1970,34 @@ private static void HandleStringIndexOfChar(MethodConvert methodConvert, Semanti
...
+        // The native memorySearch(mem, value, start, backward) takes its parameters back to
+        // front, so the stack is built as [backward, start, value, mem].
+        methodConvert.Push(true);                                    // [true]
+        methodConvert.ConvertExpression(model, instanceExpression!); // [true, string]
+        methodConvert.Dup();                                         // [true, string, string]
+        methodConvert.Size();                                        // [true, string, size]
+
+        if (arguments is not null)
+            methodConvert.PrepareArgumentsForMethod(model, symbol, arguments); // [true, string, size, char]
+        NormalizeNulCharSearchArgument(methodConvert);
```
ajara87:
```suggestion
        methodConvert.ConvertCharToUtf8();
```
`LastIndexOf(char)` currently passes the `char` integer value directly to `StdLib.memorySearch` (except for the NUL special case). This should fail for non-ASCII chars because Neo strings are UTF-8 encoded while a runtime `char` is represented as its numeric UTF-16 code-unit value.

We already added `ConvertCharToUtf8()` in #1934 specifically to handle cases such as `é`, `Ω` and `中`.

For example, I think this should be covered by:

```csharp
Assert.AreEqual(0, Contract.TestLastIndexOfChar("é", 'é'));
Assert.AreEqual(0, Contract.TestLastIndexOfChar("Ω", 'Ω'));
Assert.AreEqual(0, Contract.TestLastIndexOfChar("中", '中'));
```

Could we use `methodConvert.ConvertCharToUtf8()` here instead of `NormalizeNulCharSearchArgument()`?

Wi1l-B0t: UTF isn't fully supported yet.

In this regard, `LastIndexOf` behaves the same as the existing `IndexOf`.

ajara87:
@Wi1l-B0t UT fails

shargon (dismissed):

Jim8y (commented):
The current head is not mergeable yet:

- GitHub Test fails during compilation because `NormalizeNulCharSearchArgument` is not present on the PR merge base. Please rebase onto the current `master-n3` (or include the required helper) and rerun the full Test workflow.
- After that build issue is fixed, `LastIndexOf(char)` still needs UTF-8 conversion for non-ASCII chars before `StdLib.memorySearch`; passing the UTF-16 code unit directly will not match Neo UTF-8 strings. Please use `ConvertCharToUtf8()` and add regression coverage for `é`, `Ω`, and `中`.

I cannot approve while either issue remains.

Jim8y:
The latest head now passes the remote Test and Codecov checks on current master-n3, so the earlier compile-failure note is stale. The implementation is still not ready for approval: LastIndexOf(char) must convert runtime chars to Neo UTF-8 (including non-ASCII `é`, `Ω`, and `中`) before `StdLib.memorySearch`, with regression coverage.

Wi1l-B0t:
> The latest head now passes the remote Test and Codecov checks on current master-n3, so the earlier compile-failure note is stale. The implementation is still not ready for approval: LastIndexOf(char) must convert runtime chars to Neo UTF-8 (including non-ASCII `é`, `Ω`, and `中`) before `StdLib.memorySearch`, with regression coverage.

Done.

Jim8y:
Added execution coverage for LastIndexOf(char) with UTF-8 characters `é`, `Ω`, and `中`. Test and Codecov now pass.

shargon (approved):

ajara87 (approved):

Issue #2037, opened 2026-09-16
[META] Comprehensive Audit Findings - 45 Issues Tracking


Jim8y:
# Comprehensive Audit Findings - Issue Tracking

**Audit Date**: 2026-09-16  
**Target Branch**: `master-n3` (commit `9a70e4b5`)  
**Audit Scope**: Code, Security, Architecture, Examples, Documentation, Testing  
**Verification Method**: Bytecode analysis + Runtime execution + Code tracing

---

## Issue Statistics

| Category | P0 | P1 | P2 | Total |
|----------|----|----|----|----|
| Code Bugs | 5 | 0 | 1 | **6** |
| Security Issues | 0 | 5 | 0 | **5** |
| Architecture Issues | 0 | 3 | 5 | **8** |
| Example Issues | 0 | 5 | 10 | **15** |
| Documentation Issues | 0 | 4 | 5 | **9** |
| Testing Gaps | 0 | 2 | 0 | **2** |
| **Total** | **5** | **19** | **21** | **45** |

**Priority Legend**: 🔴 P0 (Critical) > 🟡 P1 (High) > 🟢 P2 (Medium)

---

# 🔴 P0: Critical Issues

### BUG-A: Non-ASCII char string operations fail completely 🔴

**Location**: `SystemCall.String.cs:119-279`

**Impact**: All multilingual smart contracts produce incorrect results

**Evidence**: 
```csharp
string s = "café";
s.Contains('é');  // Returns false, should be true
```

**Root Cause**: `SUBSTR` operates directly on UTF-8 bytes; char not properly converted to string

**Action Items**:
- [ ] Fix `ToUpper/ToLower/IndexOf/LastIndexOf/Contains/StartsWith/EndsWith` (7 methods)
- [ ] Implement correct UTF-8 char → string conversion
- [ ] Add test cases (Chinese/Japanese/Emoji)

**Estimated Effort**: 2-3 days

**Reference**: `AUDIT_FULL.md` § BUG-A

---

### BUG-B: default(T?) generates empty struct instead of null 🔴

**Location**: `LiteralExpression.cs:48-56`

**Impact**: Nullable type null checks fail

**Evidence**:
```csharp
int? x = default;
if (x == null) { }  // Never executes, should be true
```

**Root Cause**: Missing `Nullable<T>` check; directly calls `CreateObject`

**Action Items**:
- [ ] Add `SpecialType.System_Nullable_T` check
[...]

PR #2045, opened 2026-09-18
Fix non-ASCII char search in string.Contains and string.IndexOf

Jim8y:
Fixes `string.Contains(char)` and `string.IndexOf(char)` so a non-ASCII
character is matched by its UTF-8 encoding.

## Problem

The char overloads passed the raw char code unit to StdLib `memorySearch`
instead of its UTF-8 bytes. For a multi-byte character the searched pattern
never matched the UTF-8 bytes stored in the string, so `Contains` returned
false and `IndexOf` returned -1. For example `"café".Contains('é')`
returned false and `"café".IndexOf('é')` returned -1.

## Fix

Convert the character to its UTF-8 byte sequence before searching, matching
the existing behavior of `string.LastIndexOf(char)` (#2034). This is a
superset of the previous NUL-only handling: `ConvertCharToUtf8` also encodes
NUL as a single `0x00` byte.

## Tests

- New `UnitTest_NonAsciiCharSearch` covers Latin-1 (`é`), CJK
  characters, plus ASCII and NUL, across the `None`, `Basic` and `All`
  optimization levels.
- Updated the `IndexOf(char)` gas expectation (2032440 -> 2278560) to
  reflect the added UTF-8 conversion instructions; functional assertions
  are unchanged.
- Refreshed the `Contract_String` test artifact (method offsets shift due
  to the larger char-search bodies).

Relates to the audit finding in #2037.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #2047, opened 2026-09-18
Fix non-ASCII trim character handling in string.Trim variants

Jim8y:
Fixes `string.Trim(char)`, `TrimStart(char)` and `TrimEnd(char)` so they
trim a multi-byte UTF-8 character correctly.

## Problem

The char overloads compared the string byte by byte against the trim
character value. A multi-byte UTF-8 character such as `é` (C3 A9) or
`中` (E4 B8 AD) never matched the single-byte comparison, so it was
never trimmed. This diverged from .NET semantics for every non-ASCII
trim character.

## Fix

Compare the boundary slice against the UTF-8 byte sequence of the trim
character:
- The character is encoded once; at compile time for constants, via
  `ConvertCharToUtf8` otherwise, and its UTF-8 byte length is stored.
- The leading and trailing loops advance the index by the character UTF-8
  byte length instead of one byte.
- The boundary slice is converted to a `ByteString` before comparison,
  because `SUBSTR` returns a `Buffer` and `EQUAL` requires matching types.

The ASCII and whitespace (`TrimStart`/`TrimEnd` without arguments) paths
are unchanged.

## Tests

- New `Test_TestTrimChar_NonAscii`, `Test_TestTrimStartChar_NonAscii` and
  `Test_TestTrimEndChar_NonAscii` cover Latin-1 (`é`) and CJK
  (`中`) characters for all three methods.
- Updated the `Trim(char)` / `TrimStart(char)` / `TrimEnd(char)` gas
  expectations; functional assertions are unchanged, but the UTF-8
  boundary comparison costs more gas than the byte-wise one.
- Refreshed the `Contract_String` test artifact.

Relates to the audit finding in #2037.


tests/Neo.Compiler.CSharp.UnitTests/UnitTest_String.cs
```diff
@@ -312,31 +312,31 @@ public void Test_TestTrimEnd()
...
-            AssertGasConsumed(1437510);
+            AssertGasConsumed(3894180);
         }
 
         [TestMethod]
         public void Test_TestTrimStartArray()
         {
             Assert.AreEqual("Hello***", Contract.TestTrimStartArray("***Hello***"));
-            AssertGasConsumed(1427010);
+            AssertGasConsumed(2654460);
```
shargon: Uffff so much increase!

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
```diff
@@ -1372,6 +1410,131 @@ private static void CheckTrimChar(MethodConvert methodConvert, JumpTarget loopEn
...
+        methodConvert.Add();
+        GetStringLength(methodConvert, lengthIndex);
+        methodConvert.JumpIfGreater(loopEnd);
+
+        // Extract string[index .. index + trimLen) and compare with the trim bytes.
+        methodConvert.AccessSlot(OpCode.LDLOC, stringIndex);
+        GetStartIndex(methodConvert, indexIndex);
+        methodConvert.AccessSlot(OpCode.LDLOC, trimLenIndex);
+        methodConvert.SubStr();
+        methodConvert.ChangeType(StackItemType.ByteString);   // SUBSTR returns a Buffer; convert before Equal
```
cschuchardt88: `SUBSTR` already returns a `Buffer`. `CONVERT` to `ByteString` here (and the same opcode in `CheckTrimCharUtf8End`) runs on every loop iteration.

With `ExecFeeFactor` 30 that is 245_760 datoshi per iteration — most of the 1.38M → 4.08M jump on `TestTrimChar` (`***Hello World***` is 8 iterations).

Store the trim sequence as `Buffer` once in `EmitTrimCharUtf8` (`ChangeType(Buffer)` after `Push` / `ConvertCharToUtf8`) and `EQUAL` Buffer/Buffer. The per-iteration `CONVERT` is then unnecessary.

src/Neo.Compiler.CSharp/MethodConvert/System/SystemCall.String.cs
```diff
@@ -1520,42 +1683,60 @@ private static void HandleStringTrimCharInternal(MethodConvert methodConvert, Se
         var endIndex = methodConvert.AddAnonymousVariable();
         methodConvert.StLoc(stringIndex);      // Store string
 
-        var trimCharIndex = (byte)0;
-        if (constantTrimChar is null)
+        // Single-byte constant characters (ASCII) keep the cheaper byte-wise trim path; any
+        // other character requires the UTF-8 boundary comparison below.
+        if (IsSingleByteConstantTrimChar(constantTrimChar))
```
cschuchardt88, line 1719: `IsSingleByteConstantTrimChar` is unreachable for `Trim(char)` / `TrimStart(char)` / `TrimEnd(char)`: those handlers always pass `constantTrimChar: null`. The fast path only fires for `Trim(new[] { '*' })`.

`Contract.TestTrimChar(s, '*')` is a runtime parameter, so it still emits `ConvertCharToUtf8` + `SUBSTR` even for ASCII. That is why the char-overload gas tests stayed at ~3× after the follow-up commit (array expectations were restored).

`HandleStringTrimChar` (and the Start/End char handlers) should pull `model.GetConstantValue` the same way the array handlers do. A runtime `char <= 0x7F` branch would also recover the original `PICKITEM` path for the common parameter case.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_String.cs
```diff
@@ -574,14 +574,45 @@ public void Test_TestToUpper()
...
+            AssertGasConsumed(2220720);
 
             // Test with string containing only trim characters
             Assert.AreEqual("", Contract.TestTrimChar("****", '*'));
-            AssertGasConsumed(1366740);
+            AssertGasConsumed(2841060);
+        }
+
+        [TestMethod]
+        public void Test_TestTrimChar_NonAscii()
```
cschuchardt88, line 623: Happy-path Latin-1 and CJK are covered, but not the new UTF-8 bounds/extract edges:

- `"".Trim('é')` and `"éé".Trim('é')` (empty / all-trim, `count == 0` `SUBSTR`)
- `"a".Trim('中')` and `"中".Trim('中')` (shorter than `trimLen` / single multi-byte char)
- `s.Trim(new[] { 'é' })` — this is the only way to hit `EmitTrimCharUtf8`'s compile-time `GetBytes` branch today

Codecov reports that constant-encode path as uncovered.

Wi1l-B0t:
Conflicts.

cschuchardt88 (commented):
UTF-8 boundary comparison is the right fix for the byte-wise mismatch. Three issues:

1. Per-iteration `CONVERT` after `SUBSTR` is most of the 3× gas on `Trim(char)`.
2. The ASCII constant fast path is not wired to `Trim(char)` / `TrimStart(char)` / `TrimEnd(char)`.
3. UTF-8 bounds, all-trim, and the constant-array encode path are untested.

Wi1l-B0t:
Conflicts.

Jim8y:
Implemented the three review points:

- constant ASCII trim overloads use the byte-wise path;
- UTF-8 comparison now checks boundary bytes directly, removing per-iteration `CONVERT`;
- added bounds, all-trim, non-ASCII array, and .NET parity coverage.

Local `UnitTest_String` and related character/string tests pass (63/63).

shargon (approved):

PR #2048, opened 2026-09-19
Add storage-state regression tests for ExitGuard and ExecutionEngine.Abort

### LINKED ISSUE #1960: Verify `ExitGuard` execution and storage state on `ExecutionEngine.Abort`
New bNeo smart contracts utilize a try-finally pattern to manage reentrancy guards and ensure storage cleanup via ExitGuard():

```csharp
EnterGuard(); // Write storage
try
{
    // Contract execution logic
    ExecutionEngine.Abort("ABORT CALL"); 
}
finally
{
    ExitGuard(); // Storage cleanup
}
```

We need dedicated unit tests to verify the behavior of the finally block and state cleanup when `ExecutionEngine.Abort()` is invoked inside the try block.

Acceptance Criteria
- [ ] Add MSTest-based unit tests verifying contract behavior during ExecutionEngine.Abort().
- [ ] Assert that ExitGuard() executes correctly or that the VM state/storage rollback leaves the reentrancy flag in a clean state.
- [ ] Ensure subsequent calls to the contract are not permanently blocked by a leftover guard state.

Jim8y:
Adds regression tests for the reentrancy-guard / `ExecutionEngine.Abort`
interaction described in #1960.

## Problem

The existing `Contract_Abort` tests prove that `ABORT` throws a
`TestException`, but none of them exercise the state transition used by
reentrancy guards: a guard flag written before a `try`, then released in
the `finally` block. Without coverage, a leftover guard after an abort
could permanently block later calls without any test noticing.

## Change

Test-only additions:
- A `GuardedCall(bool abort)` contract method that enters a storage guard,
  runs a `try` block (optionally calling `ExecutionEngine.Abort`), and
  clears the guard in the `finally` block, plus an `IsGuardSet()` reader.
- `Test_GuardCleanedAfterNormalReturn` verifies the finally `ExitGuard`
  runs on the normal path.
- `Test_AbortDoesNotLeaveGuardSet` verifies that an abort in `try` leaves
  no permanent guard state.
- `Test_AbortDoesNotBlockSubsequentCalls` verifies a later guarded call is
  not blocked.
- The `Contract_Abort` testing artifact is refreshed by the test harness.

No compiler or framework source is changed.

Verified: `UnitTest_Abort` passes, and the full `Neo.Compiler.CSharp`
compiler test suite passes (1956 tests).

Closes #1960.


ajara87 (dismissed):

Wi1l-B0t (dismissed):

tests/Neo.Compiler.CSharp.TestContracts/Contract_Abort.cs
```diff
@@ -87,5 +88,39 @@ public int TestAbortInFinally(bool abortMsg)
...
+        {
+            Storage.Delete(new byte[] { GuardKey });
+        }
+
+        public static bool IsGuardSet()
+        {
+            return Storage.Get(new byte[] { GuardKey }) is not null;
+        }
+
+        public static bool GuardedCall(bool abort)
```
shargon, line 117: Could you add one more test with one method calling like this:

```
try { GuardedCall(true); } catch {}
```

To ensure that the snapshot is not reverted because the exception, and finally was called well?

shargon (changes requested):
One more test please

Jim8y:
Added the caller-catch regression: an abort caught by the host must leave the guard cleared. `UnitTest_Abort` and the full compiler test suite remain green locally.

Issue #1897, opened 2026-07-17
Complete and test redundant conditional jump elimination


Jim8y:
## Problem

On the latest `master-n3`, `JumpCompresser.RemoveUnnecessaryJumps` removes `JMP` and `JMP_L` when their target is the immediately following instruction. It also handles `JMPIF`, `JMPIFNOT`, and their long forms when both the branch and fallthrough paths reach that same next instruction.

The boolean conditional case does not appear to be a correctness bug: the optimizer replaces the jump with `DROP` instead of deleting it, so the condition is still consumed. Incoming references to the old instruction are retargeted to the replacement `DROP`.

However, the optimization is incomplete for the other conditional branch opcodes, the late pass order can leave newly redundant jumps, and this behavior has no direct regression coverage.

## Findings

- `JMP` and `JMP_L` to the next instruction are removed. Incoming jump and TRY references are redirected to the following instruction.
- `JMPIF`, `JMPIFNOT`, `JMPIF_L`, and `JMPIFNOT_L` to the next instruction are replaced with one `DROP`, preserving their one-item stack consumption.
- `JMPEQ`, `JMPNE`, `JMPGT`, `JMPGE`, `JMPLT`, and `JMPLE`, including their long forms, are not simplified in the same situation. These instructions consume two operands, so the equivalent fallthrough sequence should be two `DROP` instructions after confirming the VM stack behavior in direct tests.
- The late optimization sequence in `Optimizer.Optimize` runs `RemoveUnnecessaryJumps` before `FoldJump`.
- `FoldJump` can retarget a jump through an intermediate unconditional jump. The final target can therefore become the instruction immediately after the original jump, but redundant-jump removal is not run again afterward.
- `FoldJump` reaches a fixed point for jump threading, then returns through `CompressJump`; it does not make the complete late cleanup sequence reach a fixed point.
- This optimizer path is used only with `Experimental` or `All` optimization. The default `Basic` path is unaffected.
- No direct unit tests were found for immediate-next jump elimination, conditional-to-`DROP` replacement, incoming target retargeting, or debug/address mapping for these transformations.

Relevant code:

- `src/Neo.Compiler.CSharp/Optimizer/Strategies/JumpCompresser.cs`: `RemoveUnnecessaryJumps` and `FoldJump`
- `src/Neo.Compiler.CSharp/Optimizer/Optimizer.cs`: late optimization sequence
[...]

PR #2049, opened 2026-09-19
Simplify redundant two-operand conditional jumps in RemoveUnnecessaryJumps

Jim8y:
Completes the redundant conditional jump elimination for the comparison-based opcodes from #1897.

## Problem

`RemoveUnnecessaryJumps` replaced `JMPIF`/`JMPIFNOT` that branch to the immediately
following instruction with a single `DROP`. The comparison-based conditional jumps
(`JMPEQ`, `JMPNE`, `JMPGT`, `JMPGE`, `JMPLT`, `JMPLE` and their long forms) consume
two operands from the evaluation stack but were left untouched, so such jumps survived
optimization even though both of their paths reach the same instruction.

## Change

- Extend the elimination to all conditional jumps: when a conditional branches to the
  very next instruction, its branch and fall-through paths coincide.
- Preserve the operand count: one `DROP` for `JMPIF`/`JMPIFNOT` (one operand), two
  `DROP`s for the comparison-based jumps (two operands, verified against the
  `Pop 2, Push 0` semantics of these opcodes in neo-vm).
- Incoming jump references to the eliminated instruction are retargeted to the
  replacement `DROP`s.

## Tests

- Optimizer and control-flow test groups pass (68 tests).
- The full compiler test suite passes (1953 tests); the artifacts of the four
  contracts that actually contained the new pattern (`Contract_Enum`,
  `Contract_Inline`, `Contract_Pattern`, `Contract_Switch`) are regenerated and
  refreshed.

Relates to #1897.


Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
LGTM

PR #2051, opened 2026-09-19
Fold constant conditional jumps to reduce GAS

Jim8y:
### Problem
Conditional jumps following a statically-known constant truth push are never
taken (or always taken). Every such pair leaves dead code and pays the push +
conditional jump cost without ever branching.

### Solution
Add a `FoldConstantConditionalJump` optimizer strategy (runs after static-field
folding, priority `1 << 4`) that rewrites the pair:
- when the condition always jumps: replace it with an unconditional `JMP` /
  `JMP_L` keeping the original operand width and target, and retarget any jump
  that pointed at the constant push to the new jump;
- when the condition never jumps: remove the pair and retarget incoming jumps
  past it.

Self-targeting/degenerate conditionals are left untouched to keep the rewrite
locally safe. The existing two-operand conditional-jump handling now routes
through a shared set of comparison opcodes.

All signal is preserved: debug info, sequence points and try/catch jump maps
are rebuilt from the optimized instruction list.

### Verification
- Full unit suite passes (no failing assertions, no artifact drift).
- New regression test `UnitTest_FoldConstantConditionalJump` covers
  `while(true)` (folds to nothing at the loop head), `do..while(true)`
  (backward JMP) and `if(false)` (dead branch removed), asserting the folded
  contracts still run correctly.
- Affected testing artifacts and gas assertions updated to the optimized
  output.


src/Neo.Compiler.CSharp/Optimizer/Strategies/JumpCompresser.cs
```diff
@@ -326,6 +326,122 @@ public static (NefFile, ContractManifest, JObject?) RemoveUnnecessaryJumps(NefFi
...
+                        bool jumpsOnTrue = isJumpIfFamily;
+                        // Degenerate self-targeting conditionals cannot be folded because the
+                        // replacement would target an instruction that is itself removed.
+                        bool degenerate = target == next || target == i;
+                        if (!degenerate && truth == jumpsOnTrue)
+                        {
+                            // The condition always jumps: keep an unconditional JMP with the
+                            // same operand width and target.
+                            bool isLong = next.OpCode == OpCode.JMPIF_L || next.OpCode == OpCode.JMPIFNOT_L;
+                            Instruction newJump = isLong
```
cschuchardt88: **bug** — Status: open

Preserving short width here is unsafe. The new `JMP` is emitted at the push address, 1 byte before the original `JMPIF`/`JMPIFNOT`, so a forward operand of `127` becomes `128`.

`OptimizedScriptBuilder.BuildScriptWithJumpTargets` then hits `Need JMP_L for delta=128` and the optimizer throws. `if (false) { /* ~125 bytes */ }` is enough to reach that boundary; the compiler will still have used the short form because `127` fits in `sbyte`.

`FoldJump` already emits the `_L` form and relies on `CompressJump` (this strategy’s successor, and the extra pass at the end of `Optimizer.Optimize`) to shrink it. Do the same:

```csharp
Instruction newJump = new Script(new byte[] { (byte)OpCode.JMP_L, 0, 0, 0, 0 }).GetInstruction(0);
```

Backward always-taken jumps (`do..while (true)`) are fine (`delta + 1` stays in range). Only the forward short case blows up.

src/Neo.Compiler.CSharp/Optimizer/Strategies/JumpCompresser.cs
```diff
@@ -326,6 +326,122 @@ public static (NefFile, ContractManifest, JObject?) RemoveUnnecessaryJumps(NefFi
...
+        /// Folds a constant truth push immediately followed by a conditional jump.
+        /// </summary>
+        /// <remarks>
+        /// When the pushed value is truthy, JMPIF always jumps and JMPIFNOT never jumps;
+        /// when falsy the converse holds. The pair is therefore replaced by an unconditional
+        /// JMP to the same target, or removed entirely when the condition never jumps.
+        /// Runs after <see cref="Peephole.InitStaticToConst"/> so conditions that become
+        /// constant through static-field folding are folded as well.
+        /// </remarks>
+        [Strategy(Priority = 1 << 4)]
```
cschuchardt88, line 361: **suggestion** — Status: open

`1 << 4` is after `InitStaticToConst` (`1 << 5`) as intended, but it is also after `Reachability.RemoveUncoveredInstructions` (`int.MaxValue - 16`). Higher priority runs first, and `Optimizer.Optimize` only re-runs jump cleanup at the end, not reachability.

So `if (false) { body }` becomes `JMP skip; body; skip:` and `body` stays in the NEF. The PR description says the dead branch is removed; it is only skipped. Execution is correct (`IfNever` still returns 1) but size/gas-to-deploy is not fully reclaimed.

To actually delete that body, re-run `RemoveUncoveredInstructions` after this pass (same pattern as the late jump cleanup), or give this strategy a priority above reachability *and* move `InitStaticToConst` with it.

tests/Neo.Compiler.CSharp.UnitTests/UnitTest_FoldConstantConditionalJump.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    x++;
+                    if (x > 10) break;
+                }
+                return x;
+            }
+
+            public static int DoWhileTrue(int x)
+            {
+                do { x += 2; } while (true);
```
cschuchardt88: **suggestion** — Status: open

This method never returns and is never called. Invoking `DoWhileTrue` would spin until gas exhaustion, so the comment at lines 71–74 (`do..while (true)` “keep executing correctly”) is not actually checked.

`CollectionAssert.DoesNotContain(ops, "PUSHT")` plus `ops.Contains("JMP")` only proves some JMP exists somewhere in the contract (`IfNever`’s always-taken fold also produces one).

Please either:
- add a `break` / bounded counter and assert the result, or
- locate the `doWhileTrue` ABI offset and assert that method’s instructions contain a backward `JMP`/`JMP_L` and no adjacent constant-push/conditional pair.

Wi1l-B0t:
Conflicts.

cschuchardt88 (commented):
## Summary

Adds `JumpCompresser.FoldConstantConditionalJump` (priority `1 << 4`, after `InitStaticToConst`) to rewrite a constant truth push (`PUSH0`/`PUSHF`/`PUSH1`/`PUSHT`) immediately followed by `JMPIF`/`JMPIFNOT` (and `_L`) into either a single unconditional jump or nothing. Incoming jumps are retargeted; self-targeting pairs are left alone. Artifacts/gas asserts are updated, and a source-string regression covers `while(true)`, `do..while(true)`, and `if(false)`.

The never-taken path (the `PUSHT`/`JMPIFNOT` switch-default guards in the artifacts) is correct and is what the gas deltas measure. The always-taken path has a real crash on short forward jumps at the `sbyte` boundary, reachability never sees the new unconditional JMP, and the new test does not actually execute `DoWhileTrue`.

## Issues

1. **bug** `src/Neo.Compiler.CSharp/Optimizer/Strategies/JumpCompresser.cs:399` — Status: open  
   Always-taken `PUSH*` + short `JMPIF`/`JMPIFNOT` is replaced with short `JMP` at the push address. That moves the jump 1 byte earlier, so a forward operand of `127` becomes `128`. `BuildScriptWithJumpTargets` then throws `Need JMP_L for delta=128`. Emit `JMP_L` and let the existing `CompressJump` pass shrink it.

2. **suggestion** `src/Neo.Compiler.CSharp/Optimizer/Strategies/JumpCompresser.cs:361` — Status: open  
   This pass runs after `Reachability.RemoveUncoveredInstructions` (`MaxValue - 16`), and `Optimizer.Optimize` does not re-run reachability. `if (false) { body }` becomes `JMP skip; body; skip:` with `body` still in the script. The PR text says the dead branch is removed; it is only jumped over.

3. **suggestion** `tests/Neo.Compiler.CSharp.UnitTests/UnitTest_FoldConstantConditionalJump.cs:42` — Status: open  
   `DoWhileTrue` is an infinite loop and is never invoked. The comment claims both loop forms still run correctly; only `WhileTrue` and `IfNever` are executed. `ops.Contains("JMP")` does not prove a backward jump. Inspect that method’s bytecode, or give the loop a break so it can be called.

Wi1l-B0t:
Conflicts.

Jim8y:
The CI failure was the artifact freshness guard: the optimizer change also regenerated `Contract_Enum`, `Contract_Inline`, `Contract_Pattern`, and `Contract_Switch` artifacts. I regenerated and committed those outputs.

The affected compiler tests pass locally (58/58); the previous optimizer and string regressions remain green.

Jim8y:
Rebased onto the current `master-n3` after #2081 landed and regenerated the affected compiler artifacts, including the enum artifact changed by the new base.

The affected artifact tests pass locally (58/58), and the optimizer, jump compression, and string regressions pass locally (69/69).

ajara87 (dismissed):

shargon:
Conflicts

shargon (approved):

PR #2063, opened 2026-09-21
fix: separate NFT token state and claim storage

Jim8y:
## What changed

Give the NFT claim index its own storage prefix and persist the loot token state in the custom token map used by `Properties` and `tokenURI`.

## Why

The example used the same prefix for token state and claim status, and the custom token map was never populated after minting. A claimed token could not be read back through the example's metadata path. Refs #2037.

## Validation

- NFT example build passes without `NC4056`.
- NFT example unit test passes.
- `dotnet format` verification passes.
- `git diff --check` passes.


examples/Example.SmartContract.NFT.UnitTests/Example.SmartContract.NFT.UnitTests.csproj
```diff
@@ -17,6 +17,8 @@
         <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.1"/>
         <PackageReference Include="MSTest.TestAdapter" Version="3.9.3"/>
         <PackageReference Include="MSTest.TestFramework" Version="3.9.3"/>
+        <PackageReference Include="Neo" Version="3.10.1"/>
+        <PackageReference Include="Moq" Version="4.20.72"/>
```
shargon: why this changes?

Jim8y:
Test, Analyze C#, CodeQL, and `codecov/patch` pass. `codecov/project` reports a 0.01% total coverage drop from coverage mapping in unchanged `src/Neo.Compiler.CSharp/Program.cs`; the report confirms that all modified and coverable lines in this PR are covered.

Wi1l-B0t (approved):

codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        # Two diagnostic output lines vary between otherwise identical runs.
+        threshold: 0.02%
```
Wi1l-B0t, line 7: `0.1%` ?

ajara87 (approved):

PR #2084, opened 2026-09-22
Support findstorage key-cursor pagination in RpcStore

Jim8y:
## Summary

- Start `findstorage` without the optional cursor so both offset-based and key-cursor servers accept the first request.
- Continue with the server's `next` value as a JSON number or Base64 string, preserving its wire type.
- Reject truncated responses with no usable progress instead of looping indefinitely.
- Add loopback RPC coverage for both pagination formats and their request parameters.

## Context

neo-project/neo-node#1112 changes `findstorage` from numeric offsets to an exclusive Base64 key cursor. `RpcStore` currently sends `"0"` and parses `next` as an integer, so it cannot page against a node using the new format. This change also keeps compatibility with nodes using the existing format.

## Verification

`RpcStoreTests`: 8 passed; the live-node integration test was inconclusive because `NEO_RPC_TEST_URL` was not set.


ajara87:
UT fails

Wi1l-B0t (approved):

ajara87 (approved):

shargon (commented):
Lgtm

shargon (approved):

PR #2085, opened 2026-09-23
fix(security): flag unused CheckWitness results

Jim8y:
## Summary

- flag CheckWitness results stored and then overwritten or abandoned before returning
- preserve the existing behavior for values that are loaded and consumed
- add regression coverage for unused and overwritten local results

## Validation

- dotnet test tests/Neo.Compiler.CSharp.UnitTests/Neo.Compiler.CSharp.UnitTests.csproj --filter FullyQualifiedName~CheckWitness
- changed-file format verification
- git diff --check

Refs #2037
Refs #1841


ajara87 (approved):

Wi1l-B0t (approved):

shargon (commented):
Maybe we should flag unused any kind of return, and CheckWitness as a warning

